PR #42, opened 2017-09-27
Improve FunctionLoader validations

nelak:
Addresses issue: https://github.com/Azure/azure-functions-cli/issues/254

Reproduce steps
Configure a function for which the entry point doesn't return a function
e.g.: mistype module.exports and you'll end up with a valid entry point referencing an empty object

Proposed Solution
Improve the check and add a different message for when the entryPoint has been resolved but it is not a function

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,6 @@
     "google-protobuf": "^3.0.0",
     "grpc": "^1.6.0",
     "minimist": "^1.2.0",
-    "mock-require": "^2.0.2",
     "protobufjs": "6.7.0"
   },
   "devDependencies": {
@@ -14,6 +13,7 @@
     "@types/sinon": "^2.3.3",
     "chai": "^4.1.1",
     "mocha": "^3.5.0",
+    "mock-require": "^2.0.2",
     "node-pre-gyp": "0.6.37",
     "node-abi": "2.1.1",
     "rimraf": "^2.6.1",
--- a/package.json
+++ b/package.json
@@ -5,6 +5,7 @@
     "google-protobuf": "^3.0.0",
     "grpc": "^1.6.0",
     "minimist": "^1.2.0",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -5,6 +5,7 @@
     "google-protobuf": "^3.0.0",
     "grpc": "^1.6.0",
     "minimist": "^1.2.0",
+    "mock-require": "^2.0.2",
```
mamaso: bump this to devDependencies

src/FunctionLoader.ts
```diff
@@ -19,6 +19,9 @@ export class FunctionLoader implements IFunctionLoader {
       let scriptFilePath = <string>metadata.scriptFile;
       let script = require(scriptFilePath);
       let userFunction = getEntryPoint(script, metadata.entryPoint);
+      if(!isFunction(userFunction)) {
+        throw "Unable to determine function. Make sure the function has been correctly exported";
```
mamaso: How about `"The resolved entry point is not a function and cannot be invoked by the functions runtime. Make sure the function has been correctly exported."`

nelak:
I figured getting the environment setup was really painful since I'm not building on windows, but mostly was on the CLI, for debugging I got it working by chance with a mix of VSCode and playing with the chrome debugger.
The worker code was easiest to work with although it had some hardcoded paths and references to powershell, etc. What I still haven't figured out is where the worker gets instantiated on how the grpc communication workflow works but I guess that must be buried in some library referenced by the SDK that I couldn't step into because I lacked the pdbs with the debugging info.

mamaso:
@nelak I had a chat with some team members, we're going to make a push to release the pdbs and get them source linked to github, would make it much easier to contribute / debug issues

Thanks again for the contribution!

nelak:
@mamaso Pretty cool! That sounds great! Thanks!

PR #48, opened 2017-10-24
Change to use TypeData for rawbody (to allow for undefined)

christopheranderson:
Fixes an issue with the rawBody support for GET cases where it should be undefined (based on previous v1 behavior)


src/Converters.ts
```diff
@@ -9,7 +9,7 @@ export function fromRpcHttp(rpcHttp: rpc.RpcHttp$Properties) {
     query: rpcHttp.query,
     params: rpcHttp.params,
     body: fromTypedData(rpcHttp.body),
-    rawBody: rpcHttp.rawBody,
+    rawBody: fromTypedData(rpcHttp.rawBody, false),
```
mamaso: as mentioned in the other PR, might want to make this `string` only and avoid the boolean method flag

Issue #52, opened 2018-02-12
HTTP headers and status sent with response body when context.res.body is undefined


christopheranderson:
From https://github.com/Azure/azure-functions-host/issues/2305

Need to not send back just a json response if the body is missing.

https://github.com/Azure/azure-functions-nodejs-worker/blob/dev/src/Converters.ts#L18-L20

PR #53, opened 2018-02-13
always return http as type for toRpcHttp

christopheranderson:
fixes https://github.com/Azure/azure-functions-nodejs-worker/issues/52

tests will be added in functions-host repo (since they are still there...)

### DIFF
--- a/src/Converters.ts
+++ b/src/Converters.ts
@@ -16,15 +16,11 @@ export function fromRpcHttp(rpcHttp: rpc.RpcHttp$Properties) {
 }
 
 export function toRpcHttp(inputMessage): rpc.TypedData$Properties {
-  if (inputMessage.body !== undefined) {
     let httpMessage: rpc.RpcHttp$Properties = inputMessage;
     let status = inputMessage.statusCode || inputMessage.status;
     httpMessage.statusCode = status && status.toString();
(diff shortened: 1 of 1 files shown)


src/Converters.ts
```diff
@@ -16,15 +16,11 @@ export function fromRpcHttp(rpcHttp: rpc.RpcHttp$Properties) {
 }
 
 export function toRpcHttp(inputMessage): rpc.TypedData$Properties {
-  if (inputMessage.body !== undefined) {
     let httpMessage: rpc.RpcHttp$Properties = inputMessage;
```
pragnagopa: I think we made an assumption here that if the body is present rest of the HttpMessage properties would be available.  It's possbile inputMessage.statusCode might be missing. Do we have tests for this?

christopheranderson: We do in the main repo. I'll have another PR on azure-functions-host which will add all the HTTP tests back in.

christopheranderson: I've already run them and they pass. :) Need to make it so we can run those tests on PRs, though.

christopheranderson: Also, it's ok for status to be undefined/null/etc. It doesn't break today if you don't have status set. It just gets the default from the WebHost.

PR #55, opened 2018-03-28
Opt-in to content negotiation instead of as default

mhoeger:
Changing the "isRaw" flag to "enableContentNegotiation". Requires corresponding change when pulled through in https://github.com/Azure/azure-functions-host.

Part 1 of addressing https://github.com/Azure/azure-functions-host/issues/2552

### Background ###
 - [Docs on content negotiation](https://docs.microsoft.com/en-us/aspnet/web-api/overview/formats-and-model-binding/content-negotiation) in ASP.NET web API (what we use in [azure-functions-host](https://www.github.com/azure/azure-functions-host))

### DIFF
--- a/protos/FunctionRpc.proto
+++ b/protos/FunctionRpc.proto
@@ -251,6 +251,6 @@ message RpcHttp {
   map<string,string> params = 10;
   string status_code = 12;
   map<string,string> query = 15;
-  bool is_raw = 16;
+  bool enableContentNegotiation = 16;
   TypedData rawBody = 17;
 }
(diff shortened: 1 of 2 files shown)


protos/FunctionRpc.proto
```diff
@@ -251,6 +251,6 @@ message RpcHttp {
   map<string,string> params = 10;
   string status_code = 12;
   map<string,string> query = 15;
-  bool is_raw = 16;
+  bool enableContentNegotiation = 16;
```
pragnagopa: Proto file is moved here https://github.com/Azure/azure-functions-language-worker-protobuf

fabiocav: Really need to remove all the duplicates

PR #58, opened 2018-04-10
Host crashes if >=1 function throws an exception on initialization

### LINKED ISSUE #56: Entire host crashes if >=1 function throws an exception on initialization
"result" of GRPC StatusResult message must be string (not exception). We should not be setting an exception object to be "result", which expects a string (here: https://github.com/Azure/azure-functions-nodejs-worker/blob/58ea3456af16e5ca87b2439e7a7d2172ff4bb868/src/WorkerChannel.ts#L82)

Similar issues may be occurring where grpc message components have mismatched types.

Related to: https://github.com/Azure/azure-functions-host/issues/2510

mhoeger:
Resolves #56

### DIFF
--- a/src/WorkerChannel.ts
+++ b/src/WorkerChannel.ts
@@ -63,7 +63,7 @@ export class WorkerChannel {
 
   public functionLoadRequest(requestId: string, msg: rpc.FunctionLoadRequest) {
     if (msg.functionId && msg.metadata) {
-      let result: rpc.StatusResult$Properties = {
+      let status: rpc.StatusResult$Properties = {
         status: Status.Success
       };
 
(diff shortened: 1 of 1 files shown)


src/WorkerChannel.ts
```diff
@@ -63,31 +63,28 @@ export class WorkerChannel {
 
   public functionLoadRequest(requestId: string, msg: rpc.FunctionLoadRequest) {
     if (msg.functionId && msg.metadata) {
-      let result: rpc.StatusResult$Properties = {
+      let status: rpc.StatusResult$Properties = {
```
pragnagopa: Nit: rename status to functionLoadStatus

Issue #70, opened 2018-05-15
update proto file


pragnagopa:
headers field in RpcHttp is updated to <string, object> pairs.  ToRpcHttp and FromRpcHttp methods in Converters.ts need to be  updated.

pragnagopa:
@mhoeger - functions host changes and nodejs worker changes need to be merged at the same time as we are updating existing field.

PR #71, opened 2018-05-15
Consume changes to proto file

mhoeger:
- Delete google's duration.proto file
- Change HTTP headers to be a mapping of string : TypedData instead of string : string

Resolves https://github.com/Azure/azure-functions-nodejs-worker/issues/70

Updated subtree from https://github.com/azure/azure-functions-language-worker-protobuf. Branch: dev. Commit: 197b86287d23d4366c9efbdd06a9fd36ab06cab5


pragnagopa:
@mhoeger  - When I built with latest proto changes, there were build errors coming from [Converters.ts](https://github.com/Azure/azure-functions-nodejs-worker/blob/dev/src/Converters.ts). Can you verify?

PR #73, opened 2018-05-17
Throw an informative if using unsupported version of node (not 8.x)

### LINKED ISSUE #54: When using non-supported version, give helpful error message
Currently it "literally just craps all over itself" (Brady Gaster)

Repro:
Use Node 9. try to debug in VS Code.

mhoeger:
Resolves #54

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -492,6 +492,15 @@
         "long": "3.2.0"
       }
     },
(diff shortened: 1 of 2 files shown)


pragnagopa:
We should be able to test the error message shows up in logs via private site extension

PR #79, opened 2018-05-31
Updating headers to be string type, conversion handled on worker

### LINKED ISSUE #78: Non-string header values on http response throws error
See details on the original issue reported here: https://github.com/Azure/azure-functions-host/issues/2768
Initial fix was to update RpcHttp proto file to <string, TypedData> dictionary. According to [HttpSpec](https://tools.ietf.org/html/rfc2616#section-4.2) header values are only valid ASCII characters. Need to fix NodeJS worker to convert to strings before sending message to functions host

mhoeger:
Resolves #78

### DIFF
--- a/src/Converters.ts
+++ b/src/Converters.ts
@@ -6,7 +6,7 @@ export function fromRpcHttp(rpcHttp: rpc.RpcHttp$Properties) {
     method: <string>rpcHttp.method,
     url: <string>rpcHttp.url,
     originalUrl: <string>rpcHttp.url,
-    headers: fromRpcHttpHeaders(rpcHttp.headers),
+    headers: rpcHttp.headers,
     query: rpcHttp.query,
     params: rpcHttp.params,
     body: fromTypedData(rpcHttp.body),
@@ -14,14 +14,6 @@ export function fromRpcHttp(rpcHttp: rpc.RpcHttp$Properties) {
   };
 
   return httpContext;
-}
-
-export function fromRpcHttpHeaders(inputRpcHttpHeaders) {
-  let httpHeaders: {[key: string]: any} = {};
-  for (let key in inputRpcHttpHeaders) {
-    httpHeaders[key] = fromTypedData(inputRpcHttpHeaders[key]);
-  }
-  return httpHeaders;
 }
 
 export function toRpcHttp(inputMessage): rpc.TypedData$Properties {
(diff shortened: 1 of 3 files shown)


src/Converters.ts
```diff
@@ -34,9 +26,9 @@ export function toRpcHttp(inputMessage): rpc.TypedData$Properties {
 }
 
 export function toRpcHttpHeaders(inputHeaders: rpc.TypedData$Properties) {
-  let rpcHttpHeaders: {[key: string]: rpc.TypedData$Properties} = {};
+  let rpcHttpHeaders: {[key: string]: string} = {};
   for (let key in inputHeaders) {
-    rpcHttpHeaders[key] = toTypedData(inputHeaders[key]);
+    rpcHttpHeaders[key] = inputHeaders[key].toString();
```
pragnagopa: Check for null or empty values inputHeaders[key]?

src/Converters.ts
```diff
@@ -34,9 +26,11 @@ export function toRpcHttp(inputMessage): rpc.TypedData$Properties {
 }
 
 export function toRpcHttpHeaders(inputHeaders: rpc.TypedData$Properties) {
-  let rpcHttpHeaders: {[key: string]: rpc.TypedData$Properties} = {};
+  let rpcHttpHeaders: {[key: string]: string} = {};
   for (let key in inputHeaders) {
-    rpcHttpHeaders[key] = toTypedData(inputHeaders[key]);
+    if (inputHeaders[key] != null) {
+      rpcHttpHeaders[key] = inputHeaders[key].toString();
```
pragnagopa: Add test setting the response headers of different types "stringHeader" , 123

src/Converters.ts
```diff
@@ -34,9 +26,11 @@ export function toRpcHttp(inputMessage): rpc.TypedData$Properties {
 }
 
 export function toRpcHttpHeaders(inputHeaders: rpc.TypedData$Properties) {
-  let rpcHttpHeaders: {[key: string]: rpc.TypedData$Properties} = {};
+  let rpcHttpHeaders: {[key: string]: string} = {};
   for (let key in inputHeaders) {
-    rpcHttpHeaders[key] = toTypedData(inputHeaders[key]);
+    if (inputHeaders[key] != null) {
```
pragnagopa: if(inputHeaders[key]) should take care of undefined , null

mhoeger: it evaluates to false with a value of 0 though :(

pragnagopa: Ahh. Thanks for checking. Add a test for 0 too.

PR #91, opened 2018-06-08
NodeJs v10 support

mhoeger:
Expands supported versions to be 8.x and 10.x

Note: test failures still need to be addressed when using v10.0.0 from this issue: https://github.com/nodejs/node/issues/20258

**Must consume latest proto file before pulling into functions-host: https://github.com/Azure/azure-functions-language-worker-protobuf/commit/a42a580e5211f89d5cc605a869a48425fccbd880**

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -4,7 +4,6 @@
         "target": "es6",
         "noImplicitAny": false,
         "strict": true,
-        "strictNullChecks": false,
         "outDir": "dist/src",
         "sourceMap": true
     },
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -4,14 +4,12 @@
         "target": "es6",
         "noImplicitAny": false,
         "strict": true,
+        "strictNullChecks": false,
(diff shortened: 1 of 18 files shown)


tsconfig.json
```diff
@@ -4,6 +4,7 @@
         "target": "es6",
         "noImplicitAny": false,
         "strict": true,
+        "strictNullChecks": false,
```
pragnagopa: should we keep the default as it might be safer?

paulbatum:
Just curious - why does the upgrade to node 10 result in a breaking change in the proto file?

mhoeger:
We need to upgrade protobufjs to be able to use node 10, and upgrading protobufjs resulted in a few breaking changes. One of them is that we used to call the package "FunctionRpc" and the service "FunctionRpc" in our proto file. This now doesn't compile from type errors due to the naming conflict

paulbatum:
got it, thanks!

PR #100, opened 2018-06-26
Test Multiple Node Versions

mhoeger:


mhoeger:
@paulbatum @ColbyTresness - thoughts? is this enough?

We could also add node v8 and v10 to our E2E functions-host testing, although it's less straight-forward because our tests already take a while to run and most tests there aren't node-specific. It would be fairly easy to add v8 and v10 to our nightly builds or even just double-up on the node tests we run, but that's not very clean.

PR #108, opened 2018-07-31
Update node version warning message

### LINKED ISSUE #107: Node version error is not accurate
_From @EricJizbaMSFT on July 30, 2018 21:24_

I see the following in my logs when debugging a JavaScript function:
> Your Function App is currently set to use Node.js version v8.11.3, but the runtime requires a 10.x version.
> For deployed code, please change WEBSITE_NODE_DEFAULT_VERSION in App Settings. On your local machine, you can change node version using 'nvm' (make sure to quit and restart your code editor to pick up the changes)

But that's inconsistent with the docs here https://docs.microsoft.com/en-us/azure/azure-functions/functions-reference-node#node-version-and-package-management, plus I'm just assuming LTS would be supported. Everything seems to work for me.

_Copied from original issue: Azure/azure-functions-core-tools#604_

mhoeger:
Previously, we were going to deprecate node v8. Now, we are supporting Active LTS and Current versions.

Updating messaging to reflect decision. Resolves #107

### DIFF
--- a/src/GrpcService.ts
+++ b/src/GrpcService.ts
@@ -2,7 +2,7 @@ import * as semver from 'semver';
 
 // If user is not using Node.js v8 or v10, fail
 if (!(semver.satisfies(process.version, '8.x') || semver.satisfies(process.version, '10.x'))) {
-    let errorMessage = `Your Function App is currently set to use Node.js version ${process.version}, but the runtime requires an 8.x or 10.x version (ex: 8.11.1 or 10.6.0).
+    let errorMessage = `Your Function App is currently set to use Node.js version ${process.version}, but the runtime requires an Active LTS or Current version (ex: 8.11.1 or 10.6.0).
     For deployed code, please change WEBSITE_NODE_DEFAULT_VERSION in App Settings. On your local machine, you can change node version using 'nvm' (make sure to quit and restart your code editor to pick up the changes).`;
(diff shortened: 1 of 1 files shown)


src/GrpcService.ts
```diff
@@ -1,14 +1,11 @@
 import * as semver from 'semver';
 
-// If user is not using Node.js v10, warn to upgrade to v10
-if (!semver.satisfies(process.version, '10.x')) {
-    let errorMessage = `Your Function App is currently set to use Node.js version ${process.version}, but the runtime requires a 10.x version.
+// If user is not using Node.js v8 or v10, fail
+if (!(semver.satisfies(process.version, '8.x') || semver.satisfies(process.version, '10.x'))) {
+    let errorMessage = `Your Function App is currently set to use Node.js version ${process.version}, but the runtime requires an Active LTS or Current version (ex: 8.11.1 or 10.6.0).
```
pragnagopa: Delete - or current version

mhoeger: @pragnagopa - "Current" is a node.js release term which captures 10.x versions, i think we should keep it..!
https://github.com/nodejs/Release

mhoeger:
Yes! Although updating here too because 8.x versions are throwing errors now and they shouldn't be.

PR #111, opened 2018-08-02
Better error message for wrong init node worker args

### LINKED ISSUE #110: 'Connection info missing' too generic
https://github.com/Azure/azure-functions-nodejs-worker/blob/59d9fde5e676bc0d43cfaab4fd0c1e89b11f14e7/src/Worker.ts#L12

the error message thrown is too generic, not super helpful

mhoeger:
Helpful especially if there are azure-functions-host changes that result in missing or null worker start args.

REsolves #110

### DIFF
--- a/src/Worker.ts
+++ b/src/Worker.ts
@@ -10,7 +10,15 @@ export function startNodeWorker(args) {
   let { host, port, workerId, requestId, grpcMaxMessageLength } = parseArgs(args.slice(2));
   if (!host || !port || !workerId || !requestId || !grpcMaxMessageLength) {
     console.log('usage --host hostName --port portNumber --workerId workerId --requestId requestId --grpcMaxMessageLength grpcMaxMessageLength');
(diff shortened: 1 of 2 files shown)


src/Worker.ts
```diff
@@ -10,7 +10,7 @@ export function startNodeWorker(args) {
   let { host, port, workerId, requestId, grpcMaxMessageLength } = parseArgs(args.slice(2));
   if (!host || !port || !workerId || !requestId || !grpcMaxMessageLength) {
     console.log('usage --host hostName --port portNumber --workerId workerId --requestId requestId --grpcMaxMessageLength grpcMaxMessageLength');
-    throw new Error('Connection info missing');
+    throw new Error('gRPC client connection info is missing or null. Check \'hostName\', \'portNumber\', \'workerId\', \'requestId\', and \'grpcMaxMessageLength\'');
```
brettsam: Could you split these out to be individual validations so it's clear which is missing or incorrect?

My issue was `grpcMaxMessageLength` was 0. It confused me b/c it's a blanket check so it wasn't specific (and technically it wasn't missing or null). So I tested it locally for a bit until I realized that it was the culprit.

mhoeger:
@brettsam - changes made! mind taking a look now?

PR #112, opened 2018-08-03
Add constant to filter internal logs

mhoeger:
LANGUAGE_WORKER_LOG is an arbitrary constant, could be changed.


src/utils/Logger.ts
```diff
@@ -0,0 +1,7 @@
+export function systemLog(message?: any, ...optionalParams: any[]) { 
+    console.log(`LANGUAGE_WORKER_LOG${message}`, optionalParams); 
+};
+
+export function systemError(message?: any, ...optionalParams: any[]) { 
+    console.error(`LANGUAGE_WORKER_LOG${message}`, optionalParams);
```
pragnagopa: Rename constant to **[LanguageWorkerConsoleLog]**

PR #116, opened 2018-08-10
Fail build if packaging step fails

### LINKED ISSUE #114: Package script fails should fail build
If generating the nuget package in AppVeyor fails, the entire build should fail.

mhoeger:
Resolves #114

PowerShell doesn't automatically fail if an .exe the PowerShell script calls fails ([some context](https://stackoverflow.com/questions/9948517/how-to-stop-a-powershell-script-on-the-first-error)). 

Solution is to error if any errors were found.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@
 
 ## Publishing
 
-`publish.ps1` creates the nuget package for the worker.
+`package.ps1` creates the nuget package for the worker.
 
 It builds and webpacks the generated node files into a bundle.
 We include several grpc native modules, for x86/x64 versions of windows, osx, linux
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -8,7 +8,7 @@ branches:
 
 environment:
   matrix:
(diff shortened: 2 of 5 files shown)


mhoeger:
@pragnagopa :)

mhoeger:
@pragnagopa - mind taking a look again? did some more searching and found a cleaner solution, since we don't really care how long it takes us to fail.

pragnagopa:
Looks cleaner! Can you give an example when does the script failure take longer?

mhoeger:
This guy: https://ci.appveyor.com/project/appsvc/azure-functions-nodejs-worker/build/1.0.10075-pvcsbmps

Before, we were failing as soon as the script encountered an error, now we're waiting until the end. But it isn't too bad since the artifacts don't publish and it doesn't take tooo long :)

mhoeger:
@pragnagopa - the `exit $errors.Count` solution only works locally, as there appear to be many issues on appveyor that are actually benign (on a build we think is green, the error count is 256).

PR #126, opened 2018-09-06
Do not JSON.parse() array types

### LINKED ISSUE #125: JSON array data in odd format
Json array data (for example, bindingData properties such as enqueuedTimeStampArray for event hub trigger), comes deserialized not as a true array as it is expected.

cc: @ConnorMcMahon

mhoeger:
Parsing JavaScript array types leaves us with JSON objects that look like:
{ "0": "item 1", "1": "item 2" }

Resolves #125


mhoeger:
@ConnorMcMahon - this is the bug that's causing the weird deserialization you were seeing. Once this is checked in, I'll pull it through to the functions-host as well to see those changes!

PR #127, opened 2018-09-10
Return a promise returned from a user function

irnc:
Without added return the rejected async user function would cause an unhandled promise warning and failure to enact implicit call to context.done, which results in function timing out after 5 minutes.

@mhoeger This issue was introduced at https://github.com/Azure/azure-functions-nodejs-worker/commit/44410b6ae2cc9c786e887b2ff127c541cfc26476 when implementing `this` support for user functions.


mhoeger:
@irnc - thanks so much! this update will be deployed this week.

PR #130, opened 2018-09-24
Check for valid version and provide good error if it is invalid

christopheranderson:
This adds a check for Node.js v10 before continuing on.

I kind of just slashed this together. Probably need a proper test to check that it fails nicely. You'll need to run the test against 0.10.4 for it to do any good on App Service.

We could move it to a helper method and require it in to help unit test it more effectively, but it needs to be resilient to 0.10.4/etc., so I kept it simple for now.

Feel free to close this in favor of your own implementation if you have a different idea on how to enforce this.


mhoeger:
Thanks @christopheranderson :) this is a good idea - will modify/test with 0.10.4.

PR #131, opened 2018-09-30
Node version warning compatible down to v0

### LINKED ISSUE #128: Misleading error message when WEBSITE_NODE_DEFAULT_VERSION is missing
When an app is missing the app setting WEBSITE_NODE_DEFAULT_VERSION, we get the following error message:

``Function (HttpTriggerJS1) Error: Failed to start language worker process for: node. node exited with code 1 console.error(`${logPrefix}Couldn't require bundle, falling back to Worker,SyntaxError: Unexpected token ILLEGAL.``

- Timestamp: 2018-09-13T22:11:16.704Z
- Function App name: jsv2tester
- Function name(s) (as appropriate): HttpTriggerJS1
- Session ID: 7aa342bd276544bda9515e5215e6c30d
- Region: West US

#### Repro steps

1. Create a function app (V2) with a node function. Delete the app setting WEBSITE_NODE_DEFAULT_VERSION.
2. Try running that function

#### Expected behavior

An error message telling me that I should set that app setting (ideally with some valid values to set it to)

#### Actual behavior

I get the error message: ``Function (HttpTriggerJS1) Error: Failed to start language worker process for: node. node exited with code 1 console.error(`${logPrefix}Couldn't require bundle, falling back to Worker,SyntaxError: Unexpected token ILLEGAL.``

#### Related information 

Setting the version to one with the correct major version, but not the correct overall version (i.e. 8.0.0) works fine.

Setting the version to one without the correct major version (i.e. 6.0.0) gives the error `Function (HttpTriggerJS1) Error: Failed to start language worker process for: node. node exited with code 1 .`

mhoeger:
Changing the node version warning to be very simple javascript in case very old version (v0.10.28) default is hit.

Resolves #128

### DIFF
--- a/src/nodejsWorker.ts
+++ b/src/nodejsWorker.ts
@@ -3,30 +3,38 @@ var ACTIVE_LTS_VERSION = "v8";
 var CURRENT_BRANCH_VERSION = "v10";
 var worker;
 
-// Node version validation
-try {
-    var versionSplit = process.version.split(".");
+// Try validating node version
+// NOTE: This method should be manually tested if changed as it is in a sensitive code path 
+//       and is JavaScript that runs on at least node version 0.10.28
+function validateNodeVersion(version) {
     var message;
-    if (versionSplit.length != 3) {
-        message = "Could not parse Node.js version";
+    try {
+        var versionSplit = version.split(".");
+        var major = versionSplit[0];
+        // process.version returns invalid output
+        if (versionSplit.length != 3){
+            message = "Could not parse Node.js version: '" + version + "'";
+        // Unsupported version
+        } else if (major != ACTIVE_LTS_VERSION && major != CURRENT_BRANCH_VERSION) {
+            message = "Node.js version is too low. The version you are using is "
+                    + version +
+                    ", but the runtime requires an Active LTS or Current version (ex: 8.11.1 or 10.6.0). "
+                    + "For deployed code, change WEBSITE_NODE_DEFAULT_VERSION in App Settings. Locally, upgrade the node version used by your machine (make sure to quit and restart your code editor to pick up the changes).";
+        }
+    // Unknown error
+    } catch(err) {
+        var unknownError = "Error in validating Node.js version. ";
+        console.error(logPrefix + unknownError + err);
+        throw unknownError + err;
     }
-    var major = versionSplit[0];
-    if (major != ACTIVE_LTS_VERSION && major != CURRENT_BRANCH_VERSION) {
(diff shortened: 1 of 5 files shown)


mhoeger:
@pragnagopa - ping! (Sorry I know you're busy with Java too!)

pragnagopa:
Thanks for pinging. Missed the github notifcation.

src/nodejsWorker.ts
```diff
@@ -1,9 +1,38 @@
 var logPrefix = "LanguageWorkerConsoleLog";
+var ACTIVE_LTS_VERSION = "v8";
+var CURRENT_BRANCH_VERSION = "v10";
 var worker;
+
+// Node version validation
+try {
```
pragnagopa: Move this to a method and tests with valid versions and invalid semver versions.

mhoeger:
@pragnagopa - I'm having a hard time coming up with a good way to test this. If I separate it out to a method, it's still really hard to test because the requiring "nodejsWorker" loads the file and runs the script (`worker = require("./Worker.js");`). I don't want to move the validation logic to a separate file that is imported and run because it would mess with how we webpack & deploy (we would need to have this validation separately dealt with and moved to the root as nodejsWorker.js is). It doesn't feel too unclean, but i'm worried about additional overhead. I think end to end testing in the future may be the best way to go

src/nodejsWorker.ts
```diff
@@ -1,9 +1,46 @@
 var logPrefix = "LanguageWorkerConsoleLog";
+var ACTIVE_LTS_VERSION = "v8";
+var CURRENT_BRANCH_VERSION = "v10";
 var worker;
+
+// Try validating node version
+// NOTE: This method should be manually tested if changed as it is in a sensitive code path 
+//       and is JavaScript that runs on at least node version 0.10.28
+function validateNodeVersion(version) {
+    var message;
```
pragnagopa: We do not need E2E test. A unit test would be good enough to test the logic for version verification. Yes, you are right, that we do need a E2E test that sets the node version to something that is not supported. You can add that later.

mhoeger: It’s really hard to unit test because nodejsWorker has javascript code that executes on being loaded with ‘require’ - so every mocha test that attempts to import nodejsWorker will either fail with missing reference ‘./Worker.js’ or will run worker.startWorker. I think the way to get around this is to separate the method into a separate file and add unit tests to that exported method, but that has the challenges I mentioned above with referencing and wepacking

Issue #27, opened 2017-06-16
Produce a typings file for TypeScript users


christopheranderson:
We should produce a typings file for our externally facing APIs

paulbatum:
Does not need to be tracked in regards to V2 GA

PR #136, opened 2018-10-19
Refactor to isolate public interfaces

mhoeger:
This is step 1 to publishing typings. Idea is that `src/public/Interfaces` is the only file that will have typings generated from it and published ~~(either to @ types or as an npm package)~~ to @ types

Linked to https://github.com/Azure/azure-functions-nodejs-worker/issues/27

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -50,4 +50,4 @@ test/end-to-end/Azure.Functions.NodejsWorker.E2E/Azure.Functions.NodejsWorker.E2
 
 # Keeps internal utility functions from getting exported
 types
-!types/public/Interfaces.d.ts
+!*/types/public/Interfaces.d.ts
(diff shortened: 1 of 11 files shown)


ColbyTresness:
Looks good to me!

fabiocav:
Had a quick chat with @mhoeger and the two main pieces of feedback I had were:
 - Have our types in a namespace
 - Use actual definition `.d.ts` files

mhoeger:
@fabiocav - Thanks for the review! I've included the actual typings file we will publish. 

Also, found this relevant snippet on namespacing from the [TypeScript handbook](https://www.typescriptlang.org/docs/handbook/namespaces-and-modules.html):

> The general idea of namespacing is to provide logical grouping of constructs and to prevent name collisions. Because the module file itself is already a logical grouping, and its top-level name is defined by the code that imports it, it’s unnecessary to use an additional module layer for exported objects.

fabiocav:
I guess that does make sense in the context of modules, just wanted to make sure we were considering whether that was needed and that we were consistent with the rest of the ecosystem.

PR #140, opened 2018-10-25
Expose list of output bindings on the context object

### LINKED ISSUE #132: Expose list of output bindings on the context object
As part of function load request, functions host sends a list of input and output bindings
[LanguageWorkerChannel.cs#L352](https://github.com/Azure/azure-functions-host/blob/8f712eba4a95d1e57b93c98532f1ce02def89d83/src/WebJobs.Script/Rpc/LanguageWorkerChannel.cs#L352)
And NodeJS worker  stores this information as part of [FunctionInfo.ts#L19](https://github.com/Azure/azure-functions-nodejs-worker/blob/dev/src/FunctionInfo.ts#L19)

Exposing this list on the context object will make it easier to set values to output bindings.

mhoeger:
Resolves #132

### DIFF
--- a/src/Converters.ts
+++ b/src/Converters.ts
@@ -1,4 +1,5 @@
 import { AzureFunctionsRpcMessages as rpc } from '../azure-functions-language-worker-protobuf/src/rpc';
+import { FunctionInfo } from './FunctionInfo';
 import { HttpRequest } from './http/Request';
 import { IDict } from './Context';
 export function fromRpcHttp(rpcHttp: rpc.IRpcHttp) {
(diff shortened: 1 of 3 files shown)


mhoeger:
@pragnagopa - would you mind reviewing my PR? :)

src/Converters.ts
```diff
@@ -82,6 +96,17 @@ export function getNormalizedBindingData(request: rpc.IInvocationRequest): IDict
   return bindingData;
 }
 
+function getEnumName(direction: rpc.BindingInfo.Direction|null|undefined) {
+  switch (direction){
+    case rpc.BindingInfo.Direction.in:
+      return "in";
```
pragnagopa: Any way  get string value from enum ?

test/ConvertersTests.ts
```diff
@@ -82,4 +83,48 @@ describe('Converters', () => {
...
+            res: {
+                type: "http",
+                direction: rpc.BindingInfo.Direction.out
+            },
+            firstQueueOutput: {
+                type: "queue",
+                direction: rpc.BindingInfo.Direction.out
+            },
+            secondQueueOutput: {
+                type: "queue",
```
pragnagopa: Add
- binding with empty direction
 - binding without direction

mhoeger: @pragnagopa - is that a possible case? I thought that there would always be a binding defined ([docs seem to say so](https://docs.microsoft.com/en-us/azure/azure-functions/functions-triggers-bindings#binding-direction), although i'm sure that could change!)

pragnagopa: For unit testing we need test all possibilities 😊

mhoeger: true true :) will add!

PR #150, opened 2018-11-27
More precise 'bindingDefinitions' interface

mhoeger:


types/public/Interfaces.d.ts
```diff
@@ -7,55 +10,51 @@ export interface IContext {
...
-export interface IResponse {
-    statusCode?: string | number;
-    headers: {
-        [key: string]: any;
-    };
-    body?: any;
-    get(field: string): any;
-    set(field: string, val: any): IResponse;
-    header(field: string, val: any): IResponse;
-    status(statusCode: string | number): IResponse;
```
mhoeger: Removing IResponse from public interface _for now_ as we have two different "flavors" of HTTP responses (express-like and web-like)

types/public/Interfaces.d.ts
```diff
@@ -1,3 +1,6 @@
+export interface IFunction {
+    (context: IContext, ...args: any[]): void;
+}
```
mhoeger: Added basic interface for shape of functions we expect

test/InterfacesTest.ts
```diff
@@ -0,0 +1,40 @@
...
+
+let runServiceBus: azf.IFunction = function (context: azf.IContext, myQueueItem: string) { 
+    context.log('Node.js ServiceBus queue trigger function processed message', myQueueItem); 
+    context.log.verbose('EnqueuedTimeUtc =', context.bindingData.enqueuedTimeUtc); 
+    context.log.verbose('DeliveryCount =', context.bindingData.deliveryCount); 
+    context.log.verbose('MessageId =', context.bindingData.messageId); 
+    context.done(); 
+};
+
+export { runHttp, runHttpExpressModel, runServiceBus };
```
mhoeger: Sample uses of typescript interfaces

src/public/Interfaces.ts
```diff
@@ -1,21 +1,46 @@
...
+    bindingDefinitions: IBindingDefinition[];
     log: ILogger;
-    req?: IRequest;
-    res?: IResponse;
     done: IDoneCallback;
-};
+}
+
+export interface IHttpContext extends IContext {
+    req: IRequest;
```
pragnagopa: Should we just have IRequest? and leave res in IContext? Is it possible to define IResponse later after we figure out the right format?

test/InterfacesTest.ts
pragnagopa: How do we run these tests?

types/public/Interfaces.d.ts
```diff
@@ -7,55 +10,51 @@ export interface IContext {
...
-    };
-    body?: any;
-    get(field: string): any;
-    set(field: string, val: any): IResponse;
-    header(field: string, val: any): IResponse;
-    status(statusCode: string | number): IResponse;
+export interface IExecutionContext {
+    invocationId: string;
+    functionName: string;
+    functionDirectory: string;
```
pragnagopa: is exposing functionDirectory useful?

test/InterfacesTest.ts
```diff
@@ -0,0 +1,40 @@
+// Test typescript interfaces for ts compliation errors
+import * as azf from "../types/public/Interfaces";
+
+let runHttp: azf.IFunction = async function (context: azf.IHttpContext, req: azf.IRequest) {
+    context.log('JavaScript HTTP trigger function processed a request.');
```
pragnagopa: Update test to ILogger interface

test/InterfacesTest.ts
mhoeger: it will fail with typescript compilation errors if there's a typescript error, like this :) https://ci.appveyor.com/project/appsvc/azure-functions-nodejs-worker/builds/20603417/job/1n4239t9amph7o9u

src/public/Interfaces.ts
```diff
@@ -1,54 +1,150 @@
...
+    /**
+     * The name of your binding, as defined in function.json.
+     */
+    name: string,
+    /**
+     * The type of your binding, as defined in function.json.
+     */
+    type: string, 
+    /**
+     * The direction of your binding ('in' or 'out'), as defined in function.json.
```
pragnagopa: We also have have a [inout ](https://docs.microsoft.com/en-us/azure/azure-functions/functions-triggers-bindings#binding-direction) :)

src/public/Interfaces.ts
```diff
@@ -1,54 +1,150 @@
+/**
+ * The Azure Function that is exported (via module.exports or exports). This is the function that will
```
pragnagopa: Can we use same description as [here](https://docs.microsoft.com/en-us/azure/azure-functions/functions-reference-node)?

src/public/Interfaces.ts
```diff
@@ -1,54 +1,150 @@
+/**
+ * The Azure Function that is exported (via module.exports or exports). This is the function that will 
+ * execute when triggered. It is recommended that you declare this as an async function that implicitly
+ * returns a Promise.
+ * @param context An IContext object passed to your function from the Azure Function runtime.
+ * @returns Any output bindings or nothing.
```
pragnagopa: Optional output bindings?

src/public/Interfaces.ts
```diff
@@ -1,54 +1,150 @@
...
+}
+
+/**
+ * The context object can be used for reading and setting data from bindings, writing logs, and using the 
+ * context.done callback when your exported function is declared synchronous. A context object is passed 
+ * to your function from the Azure Function runtime on function invocation.
+ */
 export interface IContext {
+    /**
+     * A unique GUID per function invocation and execution.
```
pragnagopa: Remove `and execution`

src/public/Interfaces.ts
```diff
@@ -1,54 +1,149 @@
+/**
+ * Interface for your Azure Function code. This function must be exported (via module.exports or exports) 
+ * and will execute when triggered. It is recommended that you declare this as an async function that 
+ * implicitly returns a Promise.
```
ColbyTresness: "must be exported" contradicts the behavior when async. Or does this not apply here? Based on your second sentence, I think you should reword the first.

Also, "that implicitly returns a Promise" indicates that work needs to be done by the user, which I don't believe is the case. So maybe change to "async function, which will implicitly return a Promise"

src/public/Interfaces.ts
```diff
@@ -1,54 +1,149 @@
...
+ * @param context An IContext object passed to your function from the Azure Function runtime.
+ * @returns Optional output bindings.
+ */
+export interface IFunction {
+    (context: IContext, ...args: any[]): Promise<any> | void;
+}
+
+/**
+ * The context object can be used for reading and setting data from bindings, writing logs, and using the
```
ColbyTresness: "can be used for reading data from bindings, setting outputs, writing logs, and using the context.done callback" maybe?

src/public/Interfaces.ts
```diff
@@ -1,54 +1,157 @@
...
+ * function in the same order that they are defined in function.json.
+ * @returns Output bindings (optional).
+ */
+export interface IFunction {
+    (context: IContext, ...args: any[]): Promise<any> | void;
+}
+
+/**
+ * The context object can be used for writing logs, reading data from bindings, setting outputs and using 
+ * the context.done callback when your exported function is declared synchronous. A context object is passed
```
ColbyTresness: Grammatically, I feel like this should be either "function is declared synchronously." or "function is declared as synchronous." Just a nit and I could be wrong

types/public/Interfaces.d.ts
```diff
@@ -1,62 +1,162 @@
...
+    /**
+     * The name of your binding, as defined in function.json.
+     */
+    name: string;
+    /**
+     * The type of your binding, as defined in function.json.
+     */
+    type: string;
+    /**
+     * The direction of your binding ('in', 'out', or 'inout'), as defined in function.json.
```
ColbyTresness: Do you think we should elaborate on this more? Slash do we do this in the documentation? This has always been fairly confusing.

src/public/Interfaces.ts
mhoeger: I think "an async function that implicitly returns" is correct

src/public/Interfaces.ts
mhoeger: Yeah, agreed :) Thoughts on "your exported function is synchronous"?

types/public/Interfaces.d.ts
```diff
@@ -1,62 +1,162 @@
+/**
+ * Interface for your Azure Function code. This function must be exported (via module.exports or exports)
+ * and will execute when triggered. It is recommended that you declare this function as async, which will
+ * implicitly returns a Promise.
```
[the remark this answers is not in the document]
mhoeger: EEP ty!

src/public/Interfaces.ts
mhoeger: ^i see it now gotchu thank you!

PR #153, opened 2018-12-05
Incompatible Node.js version error message fix.

kashimiz:
### Issue
When using a version of Node that is not an Active LTS or Current version (ex. 11.3.0), the runtime currently reports "Node.js version is too low." This can lead to logically inconsistent statements, ex.

`[error] Node.js version is too low. The version you are using is v11.3.0, but the runtime requires an Active LTS or Current version (ex: 8.11.1 or 10.60.)`

### Fix
Reworded error message to be more generic.

### DIFF
--- a/src/nodejsWorker.ts
+++ b/src/nodejsWorker.ts
@@ -19,7 +19,7 @@ function validateNodeVersion(version) {
         } else if (major != ACTIVE_LTS_VERSION && major != CURRENT_BRANCH_VERSION) {
             message = "Incompatible Node.js version. The version you are using is "
                     + version +
-                    ", but the runtime requires an Active LTS or Current version (ex: 8.11.1 or 10.6.0). "
+                    ", but the runtime requires an even-numbered Active LTS or Current version (ex: 8.11.1 or 10.6.0). "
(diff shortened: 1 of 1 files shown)


andig:
Seems like a harsh constraint to exclude more current versions instead of just warning that they might now have been tested or are not fully supported?

mhoeger:
@andig - from our perspective, we think that the costs outweigh the benefits. First, we don’t want to support a local execution path that may not work in Azure (we don’t want to deploy unnecessary versions of node on our Azure machines, especially non-LTS current versions). Allowing everyone to run in a way that is not completely supported across the development experience (including testing) is a recipe for bugs around consistency for many users. As an example of where we've been burned in the past: in C#, we emit a warning that says async functions that return `void` should be avoided as they will not scale properly. We now regret making that a warning instead of a blocking error because of the number of incidents we see from people ignoring the warning and not realizing the consequences. These kinds of issues manifest in strange ways and can be very difficult to debug.

mikeerickson:
@mhoeger The LTS version 12 is not supported, thus it must be 10.x

mhoeger:
@mikeerickson - i'm a little confused by your comment! Can you explain? Also note that the information in this PR is outdated - please read current warning statement [here](https://github.com/Azure/azure-functions-nodejs-worker/blob/dev/src/nodejsWorker.ts)

mikeerickson:
@mhoeger I have things working for me, I use nvm so I just set the default to Node 10.13 and all is good.

PR #156, opened 2018-12-06
Add check and log if context logging requested after execution completes

### LINKED ISSUE #154: Warnings on context.log after context.done is called
It's very easy to mess up async patterns when using callbacks. We should emit a warning if we see any activity via context.log **after** context.done() has been called.

mhoeger:
Resolves #154

### DIFF
--- a/src/Context.ts
+++ b/src/Context.ts
@@ -5,7 +5,7 @@ import { Request, RequestProperties } from './http/Request';
 import { Response } from './http/Response';
 import LogLevel = rpc.RpcLog.Level;
 import { Context, ExecutionContext, Logger, BindingDefinition, HttpRequest } from './public/Interfaces' 
-import { systemError } from './utils/Logger';
+import { systemWarn } from './utils/Logger';
 
 export function CreateContextAndInputs(info: FunctionInfo, request: rpc.IInvocationRequest, logCallback: LogCallback, callback: ResultCallback) {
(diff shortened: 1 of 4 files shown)


src/Context.ts
```diff
@@ -87,16 +97,14 @@ class Context implements IContext {
...
-        warn: <ILog>(...args: any[]) => log(LogLevel.Warning, ...args),
-        info: <ILog>(...args: any[]) => log(LogLevel.Information, ...args),
-        verbose: <ILog>(...args: any[]) => log(LogLevel.Trace, ...args)
-      }
-    );
+// Emit warning if trying to log after function execution is done.
+function logWithAsyncCheck(done: boolean, log: ILogCallback, level: LogLevel, ...args: any[]) {
+  if (done) {
+    let badAsyncMsg = "Error: Unexpected call to 'log' after function execution has completed. Please check for asynchronous calls that are not awaited or did not use the 'done' callback where expected.";
+    log(LogLevel.Error, badAsyncMsg);
```
brettsam: Should we make this a Warning instead of an Error? I worry about existing customers that are happily running with this problem today (on accident) suddenly getting a huge wave of errors when we roll this out. Warnings would be a little more acceptable, I think.

fabiocav: Agreed.

The only other comment I have is a small nit on the test function name for the promise+callback scenario, which is currently `asyncThrowsError`, something more descriptive would be helpful.

PR #160, opened 2018-12-12
Add an interface for the HTTP response

janaagaard75:


mhoeger:
@janaagaard75 - thank you again for this :) we decided to leave out HTTP response, especially for the first round of types, because our HTTP response object is actually an interesting blend of express-like and http module-like ways of crafting a `Response`.

For example, in a JavaScript function, you will notice that you can both set status as a value and set status as a function:
```javascript
res = {
  body: {
    "hello": "world"
  },
  status: 200,
  headers: {
    "content-type": "application/json"
  }
};
```
```javascript
res.status(200).json({ "hello": "world" });
```

Maybe the best answer is to support two types that let you choose which flavor of response you want, but essentially decided that we could release the rest without having to worry about what experience would be best there.

janaagaard75:
How about adding an interface for the HTTP response, and set `status` to `any`? I have seen [@christopheranderson's comment about avoiding any](https://github.com/Azure/azure-functions-nodejs-worker/pull/166#discussion_r243390800), but isn't it a step in the right directly to have this crude interface for the HTTP response, instead of the whole response simply being marked as `any`?

```typescript
interface HttpResponse {
     body?: any
     headers?: { [key: string]: any }
     isRaw?: boolean
     status?: number
 }

janaagaard75:
Any updates on this, @mhoeger? If this is something that is not going to happen, then just close this issue. 🙂

mhoeger:
Hey @janaagaard75, no updates yet (was out for a bit, still working on pushing the initial types file). I don't think we'll be able to take this PR in, and I'd like to spend some more time fleshing out a more intentional plan for the HTTP response (to avoid breaking changes). Thanks again for the input :) But yes, I'll be closing this PR.

PR #161, opened 2018-12-12
Consider dropping the 'i' prefixes

janaagaard75:
Consider dropping the 'i' prefixes of the interfaces, since that is usually how it is done in TypeScript.

### DIFF
--- a/types/public/Interfaces.d.ts
+++ b/types/public/Interfaces.d.ts
@@ -61,9 +61,7 @@ export interface IContext {
     /**
      * HTTP response object. Provided to your function when using HTTP Bindings.
      */
-    res?: {
-        [key: string]: any;
-    };
+    res?: IResponse;
 }
 /**
  * HTTP request object. Provided to your function when using HTTP Bindings.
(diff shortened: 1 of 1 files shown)


janaagaard75:
Would it be possible to get some feedback for this pull request? It is mentioned [here](https://github.com/Azure/Azure-Functions/issues/483#issuecomment-446336973), that these type definition will be available soon, so I think it would be nice to get this approved or rejected before this is done, since it is very much a breaking change.

rhummelmose:
Completely agree, this isn't C# :)

janaagaard75:
> Completely agree, this isn't C# :)

Also my thought. 🙂 

Now that you are here, @rhummelmose, do have have any comments about the names that I have chosen? Do you think it would be better to prefix more of them with `Azure` or `Function`, for example `AzureHttpRequest` or `FunctionHttpResponse`, or would that just be overkill?

`AzureFunction`
`BindingDefinition`
`Context`
`DoneCallback`
`ExecutionContext`
`Logger`
`HttpRequest`
`HttpResponse`

rhummelmose:
@janaagaard75 I like the level of naming, but that is my own opinion :)

In case some developers think that further qualified naming makes code easier to comprehend, they can further qualify on import. I.e.:
`import * as AzureFunctions from "@azure/functions";

const index: AzureFunctions.AzureFunction = async function (context: AzureFunctions.Context, req: AzureFunctions.HttpRequest) { ... }`

@mhoeger In case you haven't seen this..

Issue #164, opened 2018-12-18
Warning message on Node version is inaccurate regarding current versions


hannesne:
The warning about the node version when running locally is not accurate, similar to issue #107 At the moment the current node version is v11, but the code checks for v10.x. (in https://github.com/Azure/azure-functions-nodejs-worker/blob/dev/src/nodejsWorker.ts)  When run against node 11, it gives an error that the node version is not LTS or Current. In the code it seems to only check for v10. 
In Issue #144 it is stated that v11 is not planned to be supported. 

We should update the error message to reflect to the user that only V10 and LTS is supported, dropping references to Current.

PR #165, opened 2018-12-18
Clarified incompatible node version error message.

hannesne:
This is in response to issue #164

### DIFF
--- a/src/nodejsWorker.ts
+++ b/src/nodejsWorker.ts
@@ -1,7 +1,6 @@
 var logPrefix = "LanguageWorkerConsoleLog";
 var errorPrefix = logPrefix + "[error] ";
-var ACTIVE_LTS_VERSION = "v8";
-var SUPPORTED_ACTIVE_BRANCH = "v10";
+var supportedVersions:string[] = ["v8","v10"];
 var worker;
 
 // Try validating node version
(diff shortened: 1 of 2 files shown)


src/nodejsWorker.ts
```diff
@@ -15,12 +15,12 @@ function validateNodeVersion(version) {
...
-        // Unsupported version
-        } else if (major != ACTIVE_LTS_VERSION && major != CURRENT_BRANCH_VERSION) {
+        // Unsupported version note: Documentation about Node's stable versions here: https://github.com/nodejs/Release#release-plan and an explanation here: https://medium.com/swlh/understanding-how-node-releases-work-in-2018-6fd356816db4
+        } else if (major != ACTIVE_LTS_VERSION && major != SUPPORTED_ACTIVE_BRANCH) {
             message = "Incompatible Node.js version. The version you are using is "
-                    + version +
-                    ", but the runtime requires an even-numbered Active LTS or Current version (ex: 8.11.1 or 10.14.1). "
-                    + "For deployed code, change WEBSITE_NODE_DEFAULT_VERSION in App Settings. Locally, upgrade the node version used by your machine (make sure to quit and restart your code editor to pick up the changes).";
+                + version +
+                ", but the runtime requires a stable Active version, currently V8.x (LTS) and V10.x. Stable versions have an even major version number (8.x, 10.x etc) as per https://github.com/nodejs/Release#release-plan. "
```
mhoeger: Can we change this to: "but the runtime requires an LTS-covered major version (ex: 8.11.1 or 10.14.1). LTS-covered versions have an even major version number (8.x, 10.x, etc.) as per https://github.com/nodejs/Release#release-plan. "

src/nodejsWorker.ts
```diff
@@ -1,7 +1,7 @@
 var logPrefix = "LanguageWorkerConsoleLog";
 var errorPrefix = logPrefix + "[error] ";
 var ACTIVE_LTS_VERSION = "v8";
```
mhoeger: `ACTIVE_LTS_VERSION_0` ?

src/nodejsWorker.ts
```diff
@@ -1,7 +1,7 @@
 var logPrefix = "LanguageWorkerConsoleLog";
 var errorPrefix = logPrefix + "[error] ";
 var ACTIVE_LTS_VERSION = "v8";
-var CURRENT_BRANCH_VERSION = "v10";
+var SUPPORTED_ACTIVE_BRANCH = "v10";
```
mhoeger: `ACTIVE_LTS_VERSION_1` ?

src/nodejsWorker.ts
mhoeger: (Note: would prefer to keep specific example versions as it is helpful for people when choosing node version in Azure)

hannesne: Yeah that looks good, I'll make that change and commit. I agree the stable part is confusing. Had to read the medium post to understand that part about the even version numbers and stable etc... We had a long discussion in CSE about it, so I think we got most of the kinks sorted out in the message. I definitely prefer the LTS-covered nomenclature though.

src/nodejsWorker.ts
hannesne: Yep. As an aside, we should probably collapse this and current_branch_version into an array of supported versions, once v12 comes out.

hannesne: I've gone ahead and changed that to an array of supported versions. Looks a bit cleaner.

hannesne:
Thanks! 😀

Issue #162, opened 2018-12-12
The log methods should be identified as methods


janaagaard75:
With the current type definitions for `error`, `warn` etc. VSCode is identifying them as properties and not methods. It works out though, but I think it would be a bit nicer if they were identified as methods.

Changing `info: ILog;` to `info(...args: any[]);`

VSCode sees `info` as a property when it's written as `info: ILog;` in the type definitions file.
<img width="358" alt="property" src="https://user-images.githubusercontent.com/1769450/49857606-4c131780-fdf3-11e8-9ac9-11a0d80fce24.png">

VSCode sees `info` as a method when it's written as `info(...args: any[]);` in the type definitions file.
<img width="472" alt="method" src="https://user-images.githubusercontent.com/1769450/49857608-4e757180-fdf3-11e8-8e84-3875a6205a07.png">

I haven't made a pull request for this change, because I don't know how to generalize the writing `ILog` as a method.

ColbyTresness:
@mhoeger

PR #166, opened 2018-12-20
Typescript Definitions: Remove 'I' prefix and make methods as methods

mhoeger:
Addressing the great feedback from @janaagaard75 (thank you!!). Unfortunately, I don't think we can keep the PR'd changes as is since we want to keep Interfaces.d.ts strictly generated. We reference many of those interfaces internally as well to make sure what we are returning truly contains values as expected.

Includes changes from:
https://github.com/Azure/azure-functions-nodejs-worker/pull/161
https://github.com/Azure/azure-functions-nodejs-worker/pull/159

And resolves: https://github.com/Azure/azure-functions-nodejs-worker/issues/162

### DIFF
--- a/src/Context.ts
+++ b/src/Context.ts
@@ -114,4 +114,5 @@ export interface Dict<T> {
   [key: string]: T
 }
 
+// Allowed input types
 export type InputTypes = HttpRequest | string | Buffer | null | undefined;
--- a/src/Context.ts
+++ b/src/Context.ts
@@ -1,20 +1,20 @@
 import { FunctionInfo } from './FunctionInfo';
 import { fromRpcHttp, fromTypedData, getNormalizedBindingData, getBindingDefinitions } from './Converters';
 import { AzureFunctionsRpcMessages as rpc } from '../azure-functions-language-worker-protobuf/src/rpc';
-import { Request, RequestBase } from './http/Request';
+import { Request, RequestProperties } from './http/Request';
 import { Response } from './http/Response';
 import LogLevel = rpc.RpcLog.Level;
-import { Context, ExecutionContext, Logger, BindingDefinition } from './public/Interfaces' 
+import { Context, ExecutionContext, Logger, BindingDefinition, InputTypes } from './public/Interfaces' 
 
 export function CreateContextAndInputs(info: FunctionInfo, request: rpc.IInvocationRequest, logCallback: LogCallback, callback: ResultCallback) {
   let context = new InvocationContext(info, request, logCallback, callback);
 
   let bindings: Dict<any> = {};
-  let inputs: any[] = [];
-  let httpInput: RequestBase | undefined;
+  let inputs: InputTypes[] = [];
+  let httpInput: RequestProperties | undefined;
   for (let binding of <rpc.IParameterBinding[]>request.inputData) {
     if (binding.data && binding.name) {
-      let input: any;
+      let input: InputTypes;
       if (binding.data && binding.data.http) {
         input = httpInput = fromRpcHttp(binding.data.http);
       } else {
@@ -104,17 +104,11 @@ export interface InvocationResult {
   bindings: Dict<any>;
 }
 
-export interface DoneCallback {
-  (err?: any, result?: any): void;
-}
+export type DoneCallback = (err?: Error | string, result?: any) => void;
 
-export interface LogCallback {
-  (level: LogLevel, ...args: any[]): void;
-}
+export type LogCallback = (level: LogLevel, ...args: any[]) => void;
 
-export interface ResultCallback {
-  (err?: any, result?: InvocationResult): void;
-}
+export type ResultCallback = (err?: any, result?: InvocationResult) => void;
 
 export interface Dict<T> {
   [key: string]: T
--- a/src/Context.ts
+++ b/src/Context.ts
@@ -4,7 +4,7 @@ import { AzureFunctionsRpcMessages as rpc } from '../azure-functions-language-wo
 import { Request, RequestProperties } from './http/Request';
 import { Response } from './http/Response';
 import LogLevel = rpc.RpcLog.Level;
-import { Context, ExecutionContext, Logger, BindingDefinition, InputTypes } from './public/Interfaces' 
+import { Context, ExecutionContext, Logger, BindingDefinition, HttpRequest } from './public/Interfaces' 
 
 export function CreateContextAndInputs(info: FunctionInfo, request: rpc.IInvocationRequest, logCallback: LogCallback, callback: ResultCallback) {
   let context = new InvocationContext(info, request, logCallback, callback);
@@ -113,3 +113,5 @@ export type ResultCallback = (err?: any, result?: InvocationResult) => void;
 export interface Dict<T> {
   [key: string]: T
 }
+
+export type InputTypes = HttpRequest | string | Buffer | null | undefined;
--- a/src/Context.ts
+++ b/src/Context.ts
@@ -1,17 +1,17 @@
 import { FunctionInfo } from './FunctionInfo';
 import { fromRpcHttp, fromTypedData, getNormalizedBindingData, getBindingDefinitions } from './Converters';
 import { AzureFunctionsRpcMessages as rpc } from '../azure-functions-language-worker-protobuf/src/rpc';
-import { Request, HttpRequest } from './http/Request';
+import { Request, RequestBase } from './http/Request';
 import { Response } from './http/Response';
 import LogLevel = rpc.RpcLog.Level;
-import { IContext, IExecutionContext, ILogger, IDoneCallback, IBindingDefinition } from './public/Interfaces' 
+import { Context, ExecutionContext, Logger, DoneCallback, BindingDefinition } from './public/Interfaces' 
 
-export function CreateContextAndInputs(info: FunctionInfo, request: rpc.IInvocationRequest, logCallback: ILogCallback, callback: IResultCallback) {
-  let context = new Context(info, request, logCallback, callback);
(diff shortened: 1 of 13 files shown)


mhoeger:
Hey @christopheranderson! We got some good feedback on the types package we're going to release. If you haven't already passed the phase of pushing out bits somewhere, we could incorporate that feedback? Should have reached out earlier but your feedback would also be very much appreciated!

src/Context.ts
```diff
@@ -31,23 +31,23 @@ export function CreateContextAndInputs(info: FunctionInfo, request: rpc.IInvocat
...
-class Context implements IContext {
+class InvocationContext implements Context {
   invocationId: string;
-  executionContext: IExecutionContext;
-  bindings: IDict<any>;
-  bindingData: IDict<any>;
-  bindingDefinitions: IBindingDefinition[];
-  log: ILogger;
+  executionContext: ExecutionContext;
+  bindings: Dict<any>;
```
christopheranderson: I wouldn't recommend introducing a generic Dict interface for just Functions.

Either do an anonymous type or do a type specific to each property. 

Also, the bindings can't contain "any". There is definitely a subset of types they support. I don't recommend shipping "any" on public types.

src/Context.ts
```diff
@@ -87,31 +87,35 @@ class Context implements IContext {
...
 
-export interface IInvocationResult {
+export interface InvocationResult {
   return: any;
-  bindings: IDict<any>;
+  bindings: Dict<any>;
 }
 
-export interface ILogCallback {
+export interface DoneCallback {
```
christopheranderson: You can do this as a type - don't need an interface.

export type DoneCallback = (err?: any, result?: any): void;

Also, `any` is dangerous here. Would recommend thinking hard about weather `any` is really acceptable for both parameters.

src/Context.ts
```diff
@@ -87,31 +87,35 @@ class Context implements IContext {
...
+  (err?: any, result?: any): void;
+}
+
+export interface LogCallback {
```
christopheranderson: Same here - callback doesn't need an interface

src/Context.ts
```diff
@@ -87,31 +87,35 @@ class Context implements IContext {
...
   (level: LogLevel, ...args: any[]): void;
 }
 
-export interface IResultCallback {
-  (err?: any, result?: IInvocationResult): void;
+export interface ResultCallback {
```
christopheranderson: Same here - callback doesn't need an interface

src/Context.ts
```diff
@@ -87,31 +87,35 @@ class Context implements IContext {
...
+  (err?: any, result?: InvocationResult): void;
 }
 
-export interface IDict<T> {
+export interface Dict<T> {
   [key: string]: T
 }
```
christopheranderson: Do you have tslint turned on with default settings? Usually it requires a new line at the end of the file.

src/Context.ts
```diff
@@ -87,31 +87,35 @@ class Context implements IContext {
...
 
-export interface IResultCallback {
-  (err?: any, result?: IInvocationResult): void;
+export interface ResultCallback {
+  (err?: any, result?: InvocationResult): void;
 }
 
-export interface IDict<T> {
+export interface Dict<T> {
```
christopheranderson: See comment above for discussion on the Dict type. This is an overly broad type that will invite some potential issues in the future if you want to change it.

src/Converters.ts
```diff
@@ -83,13 +86,13 @@ export function getBindingDefinitions(info: FunctionInfo): IBindingDefinition[]
     .map(name => { return { 
         name: name,
         type: bindings[name].type || "", 
-        direction: getDirectionName(bindings[name].direction) || ""
+        direction: <BindingDirection> getDirectionName(bindings[name].direction)
```
christopheranderson: Type assertions are dangerous. Consider creating an "isBindingDirection" helper method that does type guarding for you, rather than asserting the type. https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-definied-type-guards

src/http/Request.ts
```diff
@@ -1,8 +1,8 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License. See License.txt in the project root for license information.
-import { IRequest } from '../public/Interfaces';
+import { HttpRequest } from '../public/Interfaces';
 
-export class HttpRequest {
+export class RequestBase implements HttpRequest {
```
christopheranderson: "Base" types are generally abstract

src/public/Interfaces.ts
```diff
@@ -2,29 +2,27 @@
...
- * @param context IContext object passed to your function from the Azure Functions runtime.
+ * @param context Context object passed to your function from the Azure Functions runtime.
  * @param {any[]} args Optional array of input and trigger binding data. These binding data are passed to the 
  * function in the same order that they are defined in function.json.
  * @returns Output bindings (optional).
  */
-export interface IFunction {
-    (context: IContext, ...args: any[]): Promise<any> | void;
-}
+export type AzureFunction = ((context: Context, ...args: any[]) => Promise<any> | void);
```
christopheranderson: anys are dangerous

src/public/Interfaces.ts
```diff
@@ -50,11 +49,11 @@ export interface IContext {
      * @param err A user-defined error to pass back to the runtime. If present, your function execution will fail.
      * @param result An object containing output binding data.
      */
-    done: IDoneCallback;
+    done(err?: any, result?: any): void;
```
christopheranderson: duplicate of DoneCallback type?

src/public/Interfaces.ts
```diff
@@ -121,37 +120,33 @@ export interface IBindingDefinition {
...
-  
 /**
  * Allows you to write streaming function logs. 
  */
-export interface ILogger extends ILog {
+export interface Logger {
+    /**
+     * Writes streaming function logs at the default trace level. 
+     */
+    (...args: any[]): void;
```
christopheranderson: Can this really take anything for the first argument? (like a structured log?) Or will the first argument be expected to be a string?

(Ditto for other methods)

test/InterfacesTest.ts
```diff
@@ -1,7 +1,7 @@
 // Test typescript interfaces for ts compliation errors
-import { IFunction, IContext, IRequest} from "../types/public/Interfaces";
+import * as azf from "../types/public/Interfaces";
```
christopheranderson: Don't use `import * as ` syntax unless necessary. Just load the types you need.

src/public/Interfaces.ts
mhoeger: For the logs ones it's truly any (it looks like [util.format](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/bb560329b92bbb6a2a085bb4244c937637595da2/types/node/v8/index.d.ts#L5667) used in WorkerChannel.ts takes types of "any").

src/public/Interfaces.ts
mhoeger: Duplicate, but because Interfaces.ts is the only part that we're going to publish (better control of what we expose)... I didn't want to keep it as a type "DoneCallback" on context because I wanted done to show up as a method and not a property in intellisense. Thoughts?

src/Context.ts
mhoeger: Agreed. Since we're not publishing the types from Context.ts though, I think I'll file an issue to remove `any` and save it for later!

src/Context.ts
mhoeger: Same as previous comment, will push Dict type changes to future change since it isn't public facing

src/Converters.ts
```diff
@@ -1,13 +1,13 @@
...
+import { RequestProperties } from './http/Request';
 import { Dict } from '../src/Context';
 import { BindingDefinition } from './public/Interfaces';
 
 type BindingDirection = 'in' | 'out' | 'inout';
 
-export function fromRpcHttp(rpcHttp: rpc.IRpcHttp): RequestBase {
-  let httpContext: RequestBase = {
+export function fromRpcHttp(rpcHttp: rpc.IRpcHttp): RequestProperties {
+  let httpContext: RequestProperties = {
```
christopheranderson: nit: let should be const

We should get you hooked up with a linting tool. That would catch things like this.

test/InterfacesTest.ts
```diff
@@ -1,7 +1,7 @@
 // Test typescript interfaces for ts compliation errors
-import * as azf from "../types/public/Interfaces";
+import { AzureFunction, Context, HttpRequest } from "../types/public/Interfaces";
 
-let runHttp: azf.AzureFunction = async function (context: azf.Context, req: azf.HttpRequest) {
+let runHttp: AzureFunction = async function (context: Context, req: HttpRequest) {
```
christopheranderson: nit: let -> const

test/InterfacesTest.ts
```diff
@@ -16,7 +16,7 @@ let runHttp: azf.AzureFunction = async function (context: azf.Context, req: azf.
     }
 }
 
-let runServiceBus: azf.AzureFunction = function (context: azf.Context, myQueueItem: string) { 
+let runServiceBus: AzureFunction = function (context: Context, myQueueItem: string) {
```
christopheranderson: nit: let -> const

test/InterfacesTest.ts
```diff
@@ -25,7 +25,7 @@ let runServiceBus: azf.AzureFunction = function (context: azf.Context, myQueueIt
 };
 
 // Assumes output binding is named '$return'
-let runHttpReturn: azf.AzureFunction = async function (context: azf.Context, req: azf.HttpRequest) {
+let runHttpReturn: AzureFunction = async function (context: Context, req: HttpRequest) {
```
christopheranderson: nit: let -> const

rhummelmose:
Awesome work guys 🥇

PR #175, opened 2019-02-03
Update package.json

christopheranderson:
Without the "types" property, TypeScript can't auto-discover the types properly.


hanssens:
Simple, but nice one. This is also the recommended approach at the [TypeScript Handbook](http://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html).

mhoeger:
@christopheranderson - sorry i totally missed this before! 

I've removed "main" from the package.json as we don't really have an entry point. Will merge when green, thanks again!

PR #182, opened 2019-02-28
Refactored end-to-end tests to use xUnit's collection fixtures.

kashimiz:
Per offline discussion, this refactoring:

- removes the need to write test output to .txt file
- isolates test environments, useful for tests that require specific settings

### DIFF
--- a/test/end-to-end/Azure.Functions.NodejsWorker.E2E/Azure.Functions.NodejsWorker.E2E/CosmosDBEndToEndTests.cs
+++ b/test/end-to-end/Azure.Functions.NodejsWorker.E2E/Azure.Functions.NodejsWorker.E2E/CosmosDBEndToEndTests.cs
@@ -7,8 +7,16 @@ using Xunit;
 
 namespace Azure.Functions.NodeJs.Tests.E2E
 {
+    [Collection(Constants.FunctionAppCollectionName)]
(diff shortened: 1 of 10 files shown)


pragnagopa:
Thanks for the PR.
It is by design, that I wanted to keep starting host out of the test code as the tests can be reused against any deployed service. Right now we only run tests using CLI. Next step is to deploy to Azure and run the 
same set of tests.

test/end-to-end/Azure.Functions.NodejsWorker.E2E/Azure.Functions.NodejsWorker.E2E/CosmosDBEndToEndTests.cs
```diff
@@ -7,8 +7,16 @@
...
 namespace Azure.Functions.NodeJs.Tests.E2E
 {
+    [Collection(Constants.FunctionAppCollectionName)]
     public class CosmosDBEndToEndTests 
     {
+        private readonly FunctionAppFixture _fixture;
+
+        public CosmosDBEndToEndTests(FunctionAppFixture fixture)
+        {
+            this._fixture = fixture;
```
pragnagopa: Only use this fixture if FunctionsHostUrl  contains localhost. Otherwise, no need to spin up the host.

PR #189, opened 2019-03-12
fix rawbody => rawBody

mhoeger:


altano:
@mhoeger thanks for fixing this. This bit me today. +1 if you can get it merged and released to npm :)

mhoeger:
@altano - the change should be in npm! You may need to delete your package-lock.json and try again. Note that the prerelease version is what was incremented, which may be why you're not seeing it right away!

altano:
@mhoeger yeah I'm using the released, stable version not the pre-release. I'll wait for it to make its way to the stable version. Thanks!

altano:
Oh, I misunderstood, `latest` points at `1.0.2-beta2`. I updated and now have the fix. Thanks!

PR #194, opened 2019-03-26
add `null` for the error in `context.done(err,result)`

michael-watson:
Based on JavaScript documentation, `null` is passed if there is no error. Although an empty string works, technicall null should be valid: https://docs.microsoft.com/en-us/azure/azure-functions/functions-reference-node#contextdone-method

Should the docs be changed to be providing an empty string or is `null` the recommended approach for TypeScript? Both options work in JavaScript

### DIFF
--- a/types/public/Interfaces.d.ts
+++ b/types/public/Interfaces.d.ts
@@ -55,7 +55,7 @@ export interface Context {
      * @param result An object containing output binding data. `result` will be passed to JSON.stringify unless it is
      *  a string, Buffer, ArrayBufferView, or number.
      */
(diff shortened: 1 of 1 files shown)


mhoeger:
@michael-watson - thanks so much for the contribution!! our type definitions are generated output, so the fix also requires changing "Interfaces.ts". I made a change here that includes updating that information in the "contributing" section of the readme. Is it ok if I close this PR in favor of this? https://github.com/Azure/azure-functions-nodejs-worker/pull/195

michael-watson:
Of course! I assumed that was the case

PR #195, opened 2019-03-26
context.done should accept null as first param

mhoeger:
Added contribution notes for TS too.

Expands on work done here (thanks @michael-watson !) https://github.com/Azure/azure-functions-nodejs-worker/pull/194


michael-watson:
Thank you @mhoeger!

PR #197, opened 2019-04-03
Implement worker messages

### LINKED ISSUE #184: Implement all the message types in StreamingMessage
Functions host will start sending various messages defined under StreamingMessage: https://github.com/Azure/azure-functions-language-worker-protobuf/blob/dev/src/proto/FunctionRpc.proto#L20

To avoid breaking language worker, please implement all of them otherwise grpc will start throwing errors.
Note: only dummy implementations needed for now to avoid breaking changes

mhoeger:
Resolves #184


mhoeger:
@pragnagopa - unit tests and e2e tests are passing: 
image
image

src/WorkerChannel.ts
```diff
@@ -154,4 +194,54 @@ export class WorkerChannel {
       resultCallback(err);
     }
   }
+
+  /**
+   * Worker sends the host information identifying itself
+   */ 
+  public startStream(requestId: string, msg: rpc.StartStream): void {
+    // Not yet implemented
```
pragnagopa: startStream is the first message client sends. This is handled in worker.ts

src/WorkerChannel.ts
```diff
@@ -8,7 +8,27 @@ import { IEventStream } from './GrpcService';
 import { toTypedData } from './Converters';
 import { systemError } from './utils/Logger';
 
-export class WorkerChannel {
+/**
+ * The worker channel should have a way to handle all incoming gRPC messages.
+ * This includes all incoming StreamingMessage types (exclude *Response types and RpcLog type)
+ */
+interface IWorkerChannel {
```
pragnagopa: Move this to its own file

src/WorkerChannel.ts
```diff
@@ -154,4 +194,54 @@ export class WorkerChannel {
...
+   * Worker self terminates ..no response on this
+   */ 
+  public workerTerminate(requestId: string, msg: rpc.WorkerTerminate): void {
+    // Not yet implemented
+  }
+
+  /**
+   * NOT USED
+   */ 
+  public workerStatusRequest(requestId: string, msg: rpc.WorkerStatusRequest): void {
```
pragnagopa: Nit: This file is getting long. Move handlers to their own files

src/WorkerChannel.ts
```diff
@@ -154,4 +194,54 @@ export class WorkerChannel {
...
+   * Host requests worker to cancel invocation
+   */ 
+  public invocationCancel(requestId: string, msg: rpc.InvocationCancel): void {
+    // Not yet implemented
+  }
+  
+  /**
+   * Environment variables from the current process
+   */ 
+  public functionEnvironmentReloadRequest(requestId: string, msg: rpc.IFunctionEnvironmentReloadRequest): void {
```
pragnagopa: Please implement functionEnvironmentReloadRequest - This method needs to reload environment variables on the node process

src/WorkerChannel.ts
mhoeger: The way the WorkerChannel is currently implemented, it looks for handlers as properties on its own object. TypeScript doesn't have a notion of partial classes, so we can't split these out

src/WorkerChannel.ts
mhoeger: It's a project-wide pattern to have the interface in the same file, so i'm hesitant to change that for just one class. I think we should either change the pattern throughout the project or keep the pattern as is for consistency?

PR #238, opened 2019-08-01
Account for removal of duplicate data from host

yojagad:
Part of fix for #4735

Corresponding host PR
https://github.com/Azure/azure-functions-host/pull/4748

This needs to be merged before host PR goes in.

### DIFF
--- a/src/WorkerChannel.ts
+++ b/src/WorkerChannel.ts
@@ -5,6 +5,7 @@ import { IFunctionLoader } from './FunctionLoader';
 import { CreateContextAndInputs, LogCallback, ResultCallback } from './Context';
 import { IEventStream } from './GrpcService';
 import { toTypedData } from './converters';
+import { augmentTriggerMetadata } from './augmenters';
 import { systemError, systemLog } from './utils/Logger';
 import { InternalException } from './utils/InternalException';
 
@@ -82,8 +83,10 @@ export class WorkerChannel implements IWorkerChannel {
    * @param msg gRPC message content
    */
   public workerInitRequest(requestId: string, msg: rpc.WorkerInitRequest) {
-    let capabilitiesDictionary = {};
-    capabilitiesDictionary['RpcHttpBodyOnly'] = "true";
+    const capabilitiesDictionary = {
+      RpcHttpTriggerMetadataRemoved: "true",
+      RpcHttpBodyOnly: "true"
+    };
     this._eventStream.write({
       requestId: requestId,
       workerInitResponse: {
@@ -126,6 +129,9 @@ export class WorkerChannel implements IWorkerChannel {
    * @param msg gRPC message content
    */
   public invocationRequest(requestId: string, msg: rpc.InvocationRequest) {
+    // Repopulate triggerMetaData if http.
+    augmentTriggerMetadata(msg);
+
     let info = this._functionLoader.getInfo(<string>msg.functionId);
     let logCallback: LogCallback = (level, ...args) => {
       this.log({
--- a/src/augmenters/InvocationRequestAugmenters.ts
+++ b/src/augmenters/InvocationRequestAugmenters.ts
@@ -1,0 +1,34 @@
+import { AzureFunctionsRpcMessages as rpc } from '../../azure-functions-language-worker-protobuf/src/rpc';
+
+/**
+ * Augments TriggerMetadata from invocation with $request.
+ * @param request IInvocationRequest object
+ */
+export function augmentTriggerMetadata(request: rpc.IInvocationRequest) {
+    var key: any, value: any;
+    if (request.inputData) {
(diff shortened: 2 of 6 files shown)


src/converters/BindingConverters.ts
```diff
@@ -22,16 +22,70 @@ export function getBindingDefinitions(info: FunctionInfo): BindingDefinition[] {
 }
 
 export function getNormalizedBindingData(request: rpc.IInvocationRequest): Dict<any> {
+
+  // Repopulate triggerMetaData body and rawBody if they're not already populated.
+  var [inputDataBody, inputDataRawBody] = getBodyAndRawBody(request.inputData);
+  populateTriggerMetadataIfNotPopulated(request.triggerMetadata, inputDataBody, inputDataRawBody);
```
mhoeger: Is there a reason you're only removing the body/rawBody instead of the whole http request? the whole thing will be duplicated.

yojagad: That was my initial thought too. But on second thoughts I thought it might be better/safer to remove (and subsequently reconstruct on the worker side)  the bare minimum while trying to achieve huge perf gains. Besides, excluding body and rawbody, the rest of the elements are magnitudes of times smaller anyway.

In other words, I thought this was a lot less riskier than removing everything and trying to reconstruct it all on the worker side.

Please let me know if you have different thoughts around this and/or if I'm missing something

pragnagopa: Host should just stop sending DefaultHttpRequest as part of trigger metadata. Lets put this behind capabilities flag. That will allow you to merge this change without host changes.

yojagad: synced up offline about this. Changed this to the initial thought process of just reconstructing everything instead of just the body and the rawBody. Added this under capability flag and added unit test too.

src/WorkerChannel.ts
```diff
@@ -83,6 +83,7 @@ export class WorkerChannel implements IWorkerChannel {
    */
   public workerInitRequest(requestId: string, msg: rpc.WorkerInitRequest) {
     let capabilitiesDictionary = {};
+    capabilitiesDictionary['RpcTriggerMetadataPopulated'] = "true";
     capabilitiesDictionary['RpcHttpBodyOnly'] = "true";
```
mhoeger: You can also define it as:
```javascript
const capabilitiesDictionary = {
     RpcTriggerMetadataPopulated: "true"
     RpcHttpBodyOnly: "true"
};
```

Slight preference for this, as the capabilities dictionary should be a constant

mhoeger: Also name is a little confusing (commented on other PR too!) can it be something like `RemoveRpcHttpTriggerMetadata`

src/converters/BindingConverters.ts
```diff
@@ -22,16 +22,56 @@ export function getBindingDefinitions(info: FunctionInfo): BindingDefinition[] {
 }
 
 export function getNormalizedBindingData(request: rpc.IInvocationRequest): Dict<any> {
+
+  // Repopulate triggerMetaData if http.
+  repopulateHttp(request);
```
mhoeger: I'm a little torn on this, I know request originally had triggerMetadata with http, but I'm weary of this method just modifying request. And if that's what it's intended to do, I would move it to be called closer to where request is first received.

mhoeger: Actually I'm a little torn. I think it's cleaner to keep everything here since it's everything that creates "bindingData" from "triggerMetadata". Still a little weary of changing the object directly though at an arbitrary point

yojagad: I think I'm inclined towards what you initially suggested. Taking a second look, it is a little weird to modify the input parameter not just here but with any function invocation in general when the function's intention is just to create "bindingData" from whatever is provided as input. Let's chat tomorrow about where best to place it within the WorkerChannel so that it's reflected throughout.

src/augmenters/InvocationRequestAugmenters.ts
```diff
@@ -0,0 +1,36 @@
+import { 
+    AzureFunctionsRpcMessages as rpc
+}from '../../azure-functions-language-worker-protobuf/src/rpc';
```
mhoeger: nit: a little strange formatting? can we make this the same line?

test/WorkerChannelTests.ts
```diff
@@ -179,6 +212,9 @@ describe('WorkerChannel', () => {
         outputData: []
       }
     });
+
+    expect(JSON.stringify(actualInvocationRequest.triggerMetadata!.$request)).to.equal(JSON.stringify(inputDataValue.data));
+    expect(JSON.stringify(actualInvocationRequest.triggerMetadata!.req)).to.equal(JSON.stringify(inputDataValue.data));
```
mhoeger: Add comment?

PR #239, opened 2019-08-09
System logging pipeline

yojagad:
Part of work for Azure/azure-functions-host#3117


src/Context.ts
```diff
@@ -101,13 +102,13 @@ class InvocationContext implements Context {
...
 // Emit warning if trying to log after function execution is done.
-function logWithAsyncCheck(done: boolean, log: LogCallback, level: LogLevel, executionContext: ExecutionContext, ...args: any[]) {
+function logWithAsyncCheck(done: boolean, log: LogCallback, systemLog: SystemLogCallback, level: LogLevel, executionContext: ExecutionContext, ...args: any[]) {
   if (done) {
     let badAsyncMsg = "Warning: Unexpected call to 'log' on the context object after function execution has completed. Please check for asynchronous calls that are not awaited or calls to 'done' made before function execution completes. ";
     badAsyncMsg += `Function name: ${executionContext.functionName}. Invocation Id: ${executionContext.invocationId}. `;
     badAsyncMsg += `Learn more: https://go.microsoft.com/fwlink/?linkid=2097909 `;
     log(LogLevel.Warning, badAsyncMsg);
-    systemWarn(badAsyncMsg);
+    systemLog(SystemLogLevel.Warning, executionContext.functionName, executionContext.invocationId, badAsyncMsg);
```
pragnagopa: With in the execution context, we should not be logging system logs

src/Worker.ts
```diff
@@ -51,10 +52,25 @@ export function startNodeWorker(args) {
       errorMessage = `Worker ${workerId} uncaught exception (learn more: https://go.microsoft.com/fwlink/?linkid=2097909 ): ${err.stack || err}`;
     }
 
-    systemError(errorMessage);
+    eventStream.write({
```
mhoeger: This should be logged as a system error

src/Worker.ts
```diff
@@ -51,10 +52,25 @@ export function startNodeWorker(args) {
...
+          stacktrace:  err.stack
+        },
+        Level: LogLevel.Error
+      }
+    });
     process.exit(1);
   });
   process.on('exit', code => {
-    systemLog(`Worker ${workerId} exited with code ${code}`);
+    eventStream.write({
```
mhoeger: this should be logged as a system error

src/WorkerChannel.ts
```diff
@@ -46,11 +47,21 @@ export class WorkerChannel implements IWorkerChannel {
...
+        this.log({
+          message: `Worker ${workerId} had no handler for message '${event}'`,
+          level: LogLevel.Error,
+          logCategory: LogCategory.System
+        });
       }
     });
     eventStream.on('error', function (err) {
-      systemError(`Worker ${workerId} encountered event stream error: `, err);
+      eventStream.write({
```
mhoeger: I think this should be a system log as well because the event stream itself is in error. it's unlikely that the event stream recovered

PR #241, opened 2019-08-23
Log exception stack for uncaught exceptions

mhoeger:
Resolvse #240

### DIFF
--- a/src/utils/Logger.ts
+++ b/src/utils/Logger.ts
@@ -1,13 +1,20 @@
 const logPrefix = "LanguageWorkerConsoleLog";
 
 export function systemLog(message?: any, ...optionalParams: any[]) { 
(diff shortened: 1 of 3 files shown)


src/utils/Logger.ts
```diff
@@ -1,13 +1,20 @@
 const logPrefix = "LanguageWorkerConsoleLog";
 
 export function systemLog(message?: any, ...optionalParams: any[]) { 
-    console.log(logPrefix + message, ...optionalParams); 
+    console.log(logPrefix + removeNewLines(message), ...optionalParams);
```
mhoeger: We want to keep messages together whenever possible. a message sent to standard out / standard err with newlines will be treated as multiple, separate messages

PR #244, opened 2019-09-10
Accept TraceContext from host

yojagad:
Part of fix for Azure/azure-functions-host#3747
Host PR: https://github.com/Azure/azure-functions-host/pull/4917
LanguageWorker PR: https://github.com/Azure/azure-functions-language-worker-protobuf/pull/40

### DIFF
--- a/src/Context.ts
+++ b/src/Context.ts
@@ -1,119 +1,121 @@
 import { FunctionInfo } from './FunctionInfo';
-import { fromRpcHttp, fromTypedData, getNormalizedBindingData, getBindingDefinitions } from './converters';
+import { fromRpcHttp, fromTypedData, getNormalizedBindingData, getBindingDefinitions, fromRpcTraceContext } from './converters';
 import { AzureFunctionsRpcMessages as rpc } from '../azure-functions-language-worker-protobuf/src/rpc';
 import { Request, RequestProperties } from './http/Request';
 import { Response } from './http/Response';
 import LogLevel = rpc.RpcLog.Level;
 import LogCategory = rpc.RpcLog.RpcLogCategory;
-import { Context, ExecutionContext, Logger, BindingDefinition, HttpRequest } from './public/Interfaces' 
+import { Context, ExecutionContext, Logger, BindingDefinition, HttpRequest, TraceContext } from './public/Interfaces' 
 
 export function CreateContextAndInputs(info: FunctionInfo, request: rpc.IInvocationRequest, logCallback: LogCallback, callback: ResultCallback) {
-  let context = new InvocationContext(info, request, logCallback, callback);
+    let context = new InvocationContext(info, request, logCallback, callback);
 
-  let bindings: Dict<any> = {};
-  let inputs: InputTypes[] = [];
-  let httpInput: RequestProperties | undefined;
-  for (let binding of <rpc.IParameterBinding[]>request.inputData) {
-    if (binding.data && binding.name) {
-      let input: InputTypes;
-      if (binding.data && binding.data.http) {
-        input = httpInput = fromRpcHttp(binding.data.http);
-      } else {
-        input = fromTypedData(binding.data);
-      }
-      bindings[binding.name] = input;
-      inputs.push(input);
+    let bindings: Dict<any> = {};
+    let inputs: InputTypes[] = [];
+    let httpInput: RequestProperties | undefined;
+    for (let binding of <rpc.IParameterBinding[]>request.inputData) {
+        if (binding.data && binding.name) {
+            let input: InputTypes;
+            if (binding.data && binding.data.http) {
+                input = httpInput = fromRpcHttp(binding.data.http);
+            } else {
+                input = fromTypedData(binding.data);
+            }
+            bindings[binding.name] = input;
+            inputs.push(input);
+        }
     }
-  }
-  
-  context.bindings = bindings;
-  if (httpInput) {
-    context.req = new Request(httpInput);
-    context.res = new Response(context.done);
-  }
-  return {
-    context: <Context>context,
-    inputs: inputs
-  }
+
+    context.bindings = bindings;
+    if (httpInput) {
+        context.req = new Request(httpInput);
+        context.res = new Response(context.done);
+    }
+    return {
+        context: <Context>context,
+        inputs: inputs
+    }
 }
 
 class InvocationContext implements Context {
-  invocationId: string;
-  executionContext: ExecutionContext;
-  bindings: Dict<any>;
-  bindingData: Dict<any>;
-  bindingDefinitions: BindingDefinition[];
-  log: Logger;
-  req?: Request;
-  res?: Response;
-  done: DoneCallback;
+    invocationId: string;
+    executionContext: ExecutionContext;
+    bindings: Dict<any>;
+    bindingData: Dict<any>;
+    traceContext: TraceContext;
+    bindingDefinitions: BindingDefinition[];
+    log: Logger;
+    req?: Request;
+    res?: Response;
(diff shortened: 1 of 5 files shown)


src/ExecutionTraceContext.ts
```diff
@@ -0,0 +1,25 @@
...
+class ExecutionTraceContext implements TraceContext
+{
+  Traceparent: string | null | undefined;
+  Tracestate: string | null | undefined;
+  Attributes: {
+    [k: string]: string
+  } | null | undefined;
+
+  constructor(requestTraceContext: rpc.IRpcTraceContext | null | undefined) {
+    if(requestTraceContext!==null && requestTraceContext!==undefined)
```
mhoeger: `if (requestTraceContext)` is good too. Means "is not falsy", or is not:
- false
- 0 (zero)
- “” (empty string)
- null
- undefined
- NaN (Not A Number)

https://codeburst.io/javascript-null-vs-undefined-20f955215a2

src/public/Interfaces.ts
```diff
@@ -155,6 +155,19 @@ export interface ExecutionContext {
     functionDirectory: string;
```
mhoeger: In `Context`, there should be an additional property:

```
/**
 * Metadata used for distributed tracing. Follows proposed W3C recommendations.
 */
traceContext?: TraceContext 
```

src/public/Interfaces.ts
```diff
@@ -155,6 +155,19 @@ export interface ExecutionContext {
...
 }
 
+export interface TraceContext {
+    /** Populated during function invocation and corresponds to Activity.Current.Id during the invocation. */
```
mhoeger: Can we make this a more user-facing comment? Ex: 
`An ID that describes where in a tracing system the incoming request is.`

src/public/Interfaces.ts
```diff
@@ -155,6 +155,19 @@ export interface ExecutionContext {
...
+    Traceparent: string | null | undefined;
+
+    /** Populated during function invocation and corresponds to Activity.Current.TraceStateString during the invocation. */
+    Tracestate: string | null | undefined;
```
mhoeger: `Distributed tracing state`, or something more descriptive

src/public/Interfaces.ts
```diff
@@ -155,6 +155,19 @@ export interface ExecutionContext {
...
+  
+    /** Populated during function invocation and corresponds to Activity.Current.Tags during the invocation. */
+    Attributes: {
```
mhoeger: I actually don't know what this is? `Distributed tracing attributes`?

src/public/Interfaces.ts
```diff
@@ -155,6 +155,19 @@ export interface ExecutionContext {
     functionDirectory: string;
 }
 
+export interface TraceContext {
+    /** Populated during function invocation and corresponds to Activity.Current.Id during the invocation. */
+    Traceparent: string | null | undefined;
```
mhoeger: `traceparent: string`

src/public/Interfaces.ts
mhoeger: `tracestate: string`

src/public/Interfaces.ts
mhoeger: (i'm assuming attributes and tracestate may not exist while traceparent always does, which is why those are optional with the question mark. is that a fair assumption?)
```javascript
attributes?: { 
    [k: string]: string
};
```

src/Context.ts
```diff
@@ -50,6 +52,7 @@ class InvocationContext implements Context {
 
   constructor(info: FunctionInfo, request: rpc.IInvocationRequest, logCallback: LogCallback, callback: ResultCallback) {
     this.invocationId = <string>request.invocationId;
+    this.traceContext = CreateTraceContext(request.traceContext);
```
mhoeger: Could we do something like we do for `getNormalizedBindingData` and `getBindingDefinitions` instead of creating a new class ExecutionTraceContext class?

yojagad: Changed it to a Dict<any> as discussed. I'm resolving the other comments since they are probably not applicable anymore. Please let me know if I missed anything.

src/Context.ts
```diff
@@ -3,6 +3,7 @@ import { fromRpcHttp, fromTypedData, getNormalizedBindingData, getBindingDefinit
 import { AzureFunctionsRpcMessages as rpc } from '../azure-functions-language-worker-protobuf/src/rpc';
```
mhoeger: nit: you can import in the above `from './converters'`, because converters/index.js exports everything in `RpcConverters`

src/converters/RpcConverters.ts
```diff
@@ -30,6 +30,24 @@ export function fromTypedData(typedData?: rpc.ITypedData, convertStringToJson: b
     }
 }
 
+/**
+ * Converts 'IRpcTraceContext' input from RPC layer to dictionary of key value pairs.
+ * @param traceContext IRpcTraceContext object containing the activityId, tracestate and attributes.
+ */
+export function fromRpcTraceContext(traceContext: rpc.IRpcTraceContext | null | undefined)
+{
+    if(traceContext)
```
mhoeger: nit: `if (traceContext)`

src/public/Interfaces.ts
```diff
@@ -34,6 +34,10 @@ export interface Context {
      * Trigger metadata and function invocation data.
      */
     bindingData: { [key: string]: any };
+    /**
+     * TraceContext which is obtained from the host. Properties on this object are generated from the RpcRequest.traceContext
+     */
+    traceContext: { [key: string]: any };
```
mhoeger: I think you should still include the interface for the traceContext object, since that will always be the same. In RpcConverters, we can just say that we expect something that conforms to this interface

src/converters/RpcConverters.ts
```diff
@@ -30,6 +30,24 @@ export function fromTypedData(typedData?: rpc.ITypedData, convertStringToJson: b
     }
 }
 
+/**
+ * Converts 'IRpcTraceContext' input from RPC layer to dictionary of key value pairs.
+ * @param traceContext IRpcTraceContext object containing the activityId, tracestate and attributes.
+ */
+export function fromRpcTraceContext(traceContext: rpc.IRpcTraceContext | null | undefined)
```
mhoeger: `export function fromRpcTraceContext(traceContext: rpc.IRpcTraceContext | null | undefined): TraceContext`

src/converters/RpcConverters.ts
```diff
@@ -6,6 +6,8 @@ import {
     INullableTimestamp
 } from '../../azure-functions-language-worker-protobuf/src/rpc';
 import { InternalException } from "../utils/InternalException";
+import { TraceContext } from '../public/Interfaces';
+import { stringify } from 'querystring';
```
mhoeger: what's this import?

src/public/Interfaces.ts
```diff
@@ -34,6 +34,10 @@ export interface Context {
      * Trigger metadata and function invocation data.
      */
     bindingData: { [key: string]: any };
+    /**
+     * TraceContext which is obtained from the host. Properties on this object are generated from the RpcRequest.traceContext
+     */
+    traceContext: TraceContext;
```
mhoeger: Nit: can the comment up here also be
```
/**
 * TraceContext information to enable distributed tracing scenarios.
 */
```

src/converters/RpcConverters.ts
yojagad: Weird, I can't remember why/how that's there.. Anyway, removed it.

PR #255, opened 2019-11-13
Unblock Node.js v12 with appropriate warning

mhoeger:
@ColbyTresness and @fabiocav - could you please review?

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,5 +1,5 @@
 variables: {
-  WORKER_VERSION: '1.1.1',
+  WORKER_VERSION: '1.2.0',
   NODE_8: '8.x',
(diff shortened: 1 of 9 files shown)


azure-pipelines.yml
```diff
@@ -1,7 +1,8 @@
 variables: {
   WORKER_VERSION: '1.1.1',
   NODE_LOWER_LTS: '8.x',
-  NODE_HIGHER_LTS: '10.x'
+  NODE_HIGHER_LTS: '10.x',
+  NODE_PREVIEW_LTS: '12.x'
```
ColbyTresness: Node 12 is "Active LTS", not preview. I suppose you're saying this is preview for us though, right? That might be fine

Issue #204, opened 2019-04-30
Remove duplication / improve organization on "context" properties


mhoeger:
Originally posted here: https://github.com/Azure/azure-functions-host/issues/4335
In the above issue, it was expected that `context.bindings` should include output binding names as well as input/trigger binding data. Although we could add output binding names to `context.bindings`, this issue is to track improvements around the organization and duplication of input and output binding data.

This issue tracks future breaking changes to the context object's properties

@JamieMagee as FYI. If you could also let us know what your expectations around output bindings were, and what confused you the most, it would be helpful!

JamieMagee:
Thanks for following up with this. This doesn't necessarily need to be a breaking change right now. Updating the [documentation you linked](https://docs.microsoft.com/en-us/azure/azure-functions/functions-reference-node#contextbindings-property) to change:

> Returns a named object that contains all your input and output data

to

>Returns a named object that contains all your input and trigger data

Would have saved me some time :smile:

mhoeger:
@JamieMagee - omg... haha thanks, fixing now :)

mhoeger:
Much duplication was removed on the data passing layer but not on the "context" object itself - worth considering for V4, but for V3 we tended towards limiting breaking changes.

ejizba:
This is likely something we can address in https://github.com/Azure/azure-functions-nodejs-worker/issues/480

ejizba:
More feedback on why we need to simplify `context` in relation to unit testing:

> Azure function handler expects context object passed as the first parameter. It encapsulates request and response objects as well as information about function bindings. Normally it's prepared by the Azure runtime but in tests, we need to craft it ourselves. There are a lot of nested objects and duplicated bits of data in the context object so assembling it manually can be tedious. Luckily there's a carefully made [stub-azure-function-context](https://github.com/willmorgan/stub-azure-function-context) module which helps with stubbing the context.

https://www.maxivanov.io/unit-testing-azure-function-with-jest-typescript/

ejizba:
Minor details might still change, but overall this was addressed in the following commits as a part of the new programming model:

https://github.com/Azure/azure-functions-nodejs-library/commit/0aac975e880b8090c7519212df0942dd538a7228
https://github.com/Azure/azure-functions-nodejs-library/commit/15e0d384ec17e4b4ec4203356cc826c82e7df229
[...]

Issue #188, opened 2019-03-12
camelCase Timer Trigger object


mhoeger:
In V1, the timer trigger object looked something like this:

```javascript
{
    isPastDue: bool,
    next: string, // time
    last: string // time
}
```

In V2, the timer trigger object looks something like this:
```javascript
{
    IsPastDue: bool,
    Schedule: { ... }
    ScheduleStatus: {
        "Last": string, // time
        "LastUpdated": string, //time
        "Next": string // time
    }
}
```

In 3.x, we should be sure to camelCase the properties and hopefully address this todo: https://github.com/Azure/azure-functions-host/blob/v1.x/src/WebJobs.Script/Description/Node/NodeFunctionInvoker.cs#L408

> TODO: Need to generalize this model rather than hardcode so other extensions can also express their Node.js object model

PR #259, opened 2019-11-19
Test v2.x branch

### LINKED ISSUE #228: HTTP output through context.done(null, res) is not handled correctly
Setting HTTP output through context.done(null, res) and http output is not handled correctly. 

Example:
```javascript
let res = {
    status: 200,
    body: ["my", "array", "of", "results"],
    cookies: [ { name: "CookieName", value: "CookieValue" } ]
}
context.done(null, res);
```
Will return ```System.Collections.Generic.List`1[System.Object]``` as the body and not properly set the HTTP cookie.

We can't change this yet because the bugfix is a breaking change. Currently, this behavior is supported (which would break with the bugfix):
```javascript
let results = ["my", "array", "of", "results"];
context.done(null, results);
```
Today, this returns a 200 with the HTTP body ```[
  "my",
  "array",
  "of",
  "results"
]``` when http output binding is named `$return`

**The code fix:** In WorkerChannel.ts => invocationRequest => resultCallback => if (result.return), we need to check if it is an HTTP type and not just a generic "TypedData"

**The mitigation:** do not use context.done(err?, results?) to output results for HTTP output binding.

mhoeger:
This change adds v2,x as a branch to test. v2.x branch currently (see changes [here](https://github.com/Azure/azure-functions-nodejs-worker/compare/v2.x?expand=1)):
- removes Node.js v8
- changes worker version to 2.0.0
- introduces "v2Compatible" flag which reverts everything [breaking]
- [breaking] does not duplicate http payload in `context.bindingData`
- [breaking] fixes #228 (handles http output correctly on context.(null, { body: "hello" });

TODO:
- remove Node.js v10 support?
   - potentially leverage Node.js v12 features
- [breaking] fix camelCase of timer trigger object: #188 
   - potentially add to .proto
- [breaking] consider removing context.req and context.res objects: #204
- [breaking] in host and worker, make sure http payload "as raw" as possible
- explore options for bugs with protobuf.js and windows (Issue 1303, Issue 1275)
    - protoc or just default Linux and accept Windows bugs? Probably document windows bugs

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -8,6 +8,7 @@ name: $(WORKER_VERSION)-$(Date:yyyyMMdd)$(Rev:.r)
 trigger:
 - master
 - dev
+- v2.x
 
 jobs:
 - job: UnitTests


mhoeger:
@ColbyTresness and @pragnagopa and @fabiocav

ColbyTresness:
We decided with @fabiocav not to remove Node 10 here. Other than that looks good

PR #261, opened 2019-11-22
Camel case timer trigger

### LINKED ISSUE #188: camelCase Timer Trigger object

mhoeger:
Resolves #188

This PR also makes a change to re-add Node.js 8 binaries and move warning/throw for an unsupported version of Node.js 12 (for V2) and Node.js 8 (for V3). The check is needed in worker init, because this is where we will know which compatability mode the worker is running in. The check at the very start is needed in case an extremely old node version (ex: 0.28) is used. In these cases, which is a somewhat common scenario for Windows sites, the Node.js version is so old we won't get to the point of workerInit and the exception thrown will be very difficult to debug.

Updates flag name too to be more precise (v2Compatible => v1WorkerBehavior). 

When merged, must also make this change: https://github.com/Azure/azure-functions-templates/issues/909


src/WorkerChannel.ts
```diff
@@ -92,13 +92,31 @@ export class WorkerChannel implements IWorkerChannel {
...
-    if (msg.hostVersion) {
-      this._v2Compatible = true;
+    if (msg.capabilities["V2Compatable"]) {
+      this._v1WorkerBehavior = true;
     }
+
+    // Validate version
+    let version = process.version;
+    if (this._v1WorkerBehavior) {
+      if (version.startsWith("v12."))
```
yojagad: >v12. [](start = 30, length = 4)

Can this be V12. or something?

src/WorkerChannel.ts
```diff
@@ -92,13 +92,31 @@ export class WorkerChannel implements IWorkerChannel {
    */
   public workerInitRequest(requestId: string, msg: rpc.WorkerInitRequest) {
     // TODO: add capability from host to go to "non-breaking" mode
-    if (msg.hostVersion) {
-      this._v2Compatible = true;
+    if (msg.capabilities["V2Compatable"]) {
```
yojagad: >V2Compatable [](start = 26, length = 12)

nit: V2Compatible? Is this an existing capability?

src/WorkerChannel.ts
mhoeger: is always lowercase!

PR #263, opened 2019-12-01
Add support for TypedDataCollections 

### LINKED ISSUE #262: Event hub, Binary datatype and Cardinality many
An array of binary inputs will not be correctly serialized as such. 

https://github.com/Azure/azure-functions-java-worker/issues/297

mhoeger:
Resolves #262

### DIFF
--- a/src/converters/RpcConverters.ts
+++ b/src/converters/RpcConverters.ts
@@ -7,6 +7,7 @@ import {
 } from '../../azure-functions-language-worker-protobuf/src/rpc';
 import { InternalException } from "../utils/InternalException";
 import { TraceContext } from '../public/Interfaces';
+import { isLong } from 'long'
 
 /**
  * Converts 'ITypedData' input from the RPC layer to JavaScript types.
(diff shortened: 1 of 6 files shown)


src/converters/RpcConverters.ts
```diff
@@ -28,6 +29,16 @@ export function fromTypedData(typedData?: rpc.ITypedData, convertStringToJson: b
...
+    } else if (typedData.collectionBytes && typedData.collectionBytes.bytes) {
+        let byteCollection = <Uint8Array[]>typedData.collectionBytes.bytes;
+        return byteCollection.map(element => Buffer.from(<Buffer>element));
+    } else if (typedData.collectionString && typedData.collectionString.string) {
+        return <string[]>typedData.collectionString.string;
+    } else if (typedData.collectionDouble && typedData.collectionDouble.double) {
+        return <number[]>typedData.collectionDouble.double;
+    } else if (typedData.collectionSint64 && typedData.collectionSint64.sint64) {
+        let longCollection = <(Long|number)[]>typedData.collectionSint64.sint64;
+        return longCollection.map(element => isLong(element) ? element.toString() : <number>element);
```
amamounelsayed: What will happen in case the type is not defined in these if, specially there is no else default case? This case should not ever happen.

mhoeger: @amamounelsayed - agreed... good catch :) I'm going to track this as an issue to fix separately, since it's an improvement on our current behavior and we do want to get this in for the release today: https://github.com/Azure/azure-functions-nodejs-worker/issues/267

PR #264, opened 2019-12-02
Fix return output binding logic

### LINKED ISSUE #232: Node: Blob output binding should support passing a Buffer
#### What problem would the feature you're requesting solve? Please describe.

I'm running a Node.js Azure Function with an output Blob binding. I'm trying to save a binary file from a `Buffer`.

My binding:

``` json
{
  "name":"copy",
  "type":"blob",
  "direction":"out",
  "path":"files/copy-{name}",
  "connection":"StorageConnectionStringKey",
  "dataType":"binary"
}
```

Note that I tried `dataType=binary` but it doesn't seem to make any difference.

and in the function body:

``` js
return { copy: new Buffer("Hello") };
```

The blob contents is written as

```
{"type":"Buffer","data":[72,101,108,108,111]}
```

#### Describe the solution you'd like

It should save the data from the `Buffer` as-is. `Hello` in this case.

#### Related Issues

- [Set a Content-Type](https://github.com/Azure/azure-functions-host/issues/1151)
- [Blob bindings can't set ContentType and other properties](https://github.com/Azure/azure-functions-host/issues/364)

mhoeger:
Resolves #232


mhoeger:
https://github.com/Azure/azure-functions-nodejs-worker/issues/265

Issue #272, opened 2019-12-04
[Worker V2] Output bindings should be able to be assigned in a combination of ways


mhoeger:
Per docs, anything in context.done/return *overrides* anything set by context.bindings. However, context.done/return should not invalidate anything set in context.bindings that has not been set in context.done/return
https://docs.microsoft.com/en-us/azure/azure-functions/functions-reference-node#contextdone-method

Ex:
```
    context.bindings.outputEventHubMessage = Buffer.from("hellooo");

    return { 
        res: { body: "hello" }
    };
```
should be a valid way to set both output bindings

PR #273, opened 2019-12-04
Don't override all output binding data when return result present

mhoeger:
Resolves https://github.com/Azure/azure-functions-nodejs-worker/issues/272

### DIFF
--- a/src/WorkerChannel.ts
+++ b/src/WorkerChannel.ts
@@ -186,9 +186,12 @@ export class WorkerChannel implements IWorkerChannel {
         invocationId: msg.invocationId,
         result: this.getStatus(err)
       }
+      // explicitly set outputData to empty array to concat later
+      response.outputData = [];
 
       try {
         if (result) {
+          // Set results from return / context.done
           if (result.return) {
             if (this._v1WorkerBehavior) {
               response.returnValue = toTypedData(result.return);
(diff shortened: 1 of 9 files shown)


test/WorkerChannelTests.ts
```diff
@@ -503,12 +570,7 @@ describe('WorkerChannel', () => {
         result:  {
           status: rpc.StatusResult.Status.Success
         },
-        outputData: [{
-          data: {
-            http: httpResponse
-          },
-          name: "res"
-        }],
```
mhoeger: With this change, we will no longer see outputData being populated with "res", which is defined automatically, if there is a return value. this is ok because in cases where an http output is defined through return/context.done, it always takes priority

src/WorkerChannel.ts
```diff
@@ -208,14 +211,15 @@ export class WorkerChannel implements IWorkerChannel {
               }
             }
           }
-          // Data from return supersedes data from context.bindings
-          if (result.bindings && !response.outputData) {
-            response.outputData = Object.keys(info.outputBindings)
-              .filter(key => result.bindings[key] !== undefined)
+          // Set results from context.bindings
+          if (result.bindings) {
+            response.outputData = response.outputData.concat(Object.keys(info.outputBindings)
```
yojagad: >response.outputData.concat [](start = 34, length = 26)

From looking up, I think concat flattens things, right? 

for instance,
context.bindings.outputEventMessage = [1,2]
return {
   res: {body: [3,4]}
}

do you want [1,2,3,4] to be returned or does it make sense to return it as [[1,2],[3,4]]? I don't know if this example clearly indicates what I'm trying to say / if this is even a valid example...

mhoeger: @yojagad - One thing to note is that every element in outputData is mapped to an object with name and data properties, so:
outputEventMessage maps to:
```
{
  name: outputEventMessage,
  data: [1, 2]
}
```
and res maps to (roughly... data is actually converted to a different type):
```
{
  name: res,
  data: { body: [3, 4] }
}
```
So the concat is between 
```
[{
  name: outputEventMessage,
  data: [1, 2]
}]
```
and
```
[{
  name: res,
  data: { body: [3, 4] }
}]
```
which I want to become
```
[{
  name: outputEventMessage,
  data: [1, 2]
},
{
  name: res,
  data: { body: [3, 4] }
}]
```

PR #277, opened 2020-01-09
Fix durable-js in Azure Functions V3

mhoeger:
An orchestration trigger binding doesn't have a corresponding, explicit output, but does expect to be returned. This change returns that information in JSON form.

### DIFF
--- a/src/FunctionInfo.ts
+++ b/src/FunctionInfo.ts
@@ -13,7 +13,7 @@ export class FunctionInfo {
     [key: string]: rpc.IBindingInfo & { converter: (any) => rpc.ITypedData }
   };
   public httpOutputName: string;
-  public isHttp: boolean;
+  public hasHttpTrigger: boolean;
 
   constructor(metadata: rpc.IRpcFunctionMetadata) {
     this.name = <string>metadata.name;
@@ -21,7 +21,7 @@ export class FunctionInfo {
     this.bindings = {};
     this.outputBindings = {};
     this.httpOutputName = "";
-    this.isHttp = false;
+    this.hasHttpTrigger = false;
 
     if (metadata.bindings) {
       let bindings = this.bindings = metadata.bindings;
@@ -39,10 +39,10 @@ export class FunctionInfo {
           }
         });
       
-      this.isHttp = Object.keys(bindings)
+      this.hasHttpTrigger = Object.keys(bindings)
         .filter(name => {
           let type = bindings[name].type;
-          return type && type.toLowerCase() == 'httptrigger';
+          return type && type.toLowerCase() === 'httptrigger';
         })
         .length > 0;
     }
--- a/src/FunctionInfo.ts
+++ b/src/FunctionInfo.ts
@@ -42,7 +42,7 @@ export class FunctionInfo {
       this.isHttp = Object.keys(bindings)
         .filter(name => {
           let type = bindings[name].type;
-          return type && type.toLowerCase() == 'httpTrigger';
+          return type && type.toLowerCase() == 'httptrigger';
         })
         .length > 0;
     }
--- a/src/FunctionInfo.ts
+++ b/src/FunctionInfo.ts
@@ -13,6 +13,7 @@ export class FunctionInfo {
     [key: string]: rpc.IBindingInfo & { converter: (any) => rpc.ITypedData }
   };
   public httpOutputName: string;
+  public isHttp: boolean;
 
   constructor(metadata: rpc.IRpcFunctionMetadata) {
     this.name = <string>metadata.name;
@@ -20,6 +21,7 @@ export class FunctionInfo {
     this.bindings = {};
     this.outputBindings = {};
     this.httpOutputName = "";
+    this.isHttp = false;
 
     if (metadata.bindings) {
       let bindings = this.bindings = metadata.bindings;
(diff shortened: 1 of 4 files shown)


src/WorkerChannel.ts
```diff
@@ -209,6 +209,11 @@ export class WorkerChannel implements IWorkerChannel {
                     data: info.outputBindings[key].converter(result.return[key])
                   });
               }
+              // returned value does not match any output bindings - pass it along
+              // this specifically is to allow the durable binding to work
+              if (!response.returnValue && response.outputData.length == 0) {
+                response.returnValue = toTypedData(result.return);
+              }
```
cgillum: This is not specific to the change you're making, but I'm trying to understand the logic of this larger code block. I see that we check on line 195 whether `return.return` is truthy. I also see on line 204 that we make the same check again. Won't that second check always return true based on what was established at line 195? In other words, could we just need to just use `else` instead of `else if (return.return)` on line 204?

More closely related to your change, I see that we set the return value only if `response.returnValue` and `response.outputData` are null/empty. What would happen if the function has a regular output binding but also returns a value? Would that cause line 215 to be skipped? I see that this code is only about return values, but is there any other situation where `response.outputData.length` might be non-zero?

mhoeger: @cgillum -
> could we just need to just use else instead of else if (return.return) on line 204?

yes!! changing, thanks!

> What would happen if the function has a regular output binding but also returns a value?

There's [three documented ways](https://docs.microsoft.com/en-us/azure/azure-functions/functions-reference-node#outputs) of using output bindings for the node worker. The existing code handles for these three situations. We decided to treat all returned output as belonging to an output binding named $return, if it is present in the host I believe in v1: [dupe logic here](https://github.com/Azure/azure-functions-host/blob/a2a11fb1adc87d6068019b7c8b6a288c37274201/src/WebJobs.Script/Description/Workers/WorkerFunctionInvoker.cs#L42-L49). For a function app that also defines output bindings using `context.bindings`, the rest of the code (starting with `result.bindings`) should handle that.

The new code only applies to output that is returned but does not have a corresponding "out" binding defined in function.json. As far as I know, durable is the only extension that uses returned output without having a corresponding "out" binding in function.json. @fabiocav - do you know if that's true?

mhoeger: Update on this -
I think [this code in the durable extension](https://github.com/Azure/azure-functions-durable-extension/blob/dev/src/WebJobs.Extensions.DurableTask/Bindings/OrchestrationTriggerAttributeBindingProvider.cs#L95-L99) registers an implicit "$return". 

I've looked in these repos:
https://github.com/Azure/azure-webjobs-sdk-extensions
https://github.com/Azure/azure-functions-durable-extension
https://github.com/Azure/azure-functions-eventgrid-extension
https://github.com/Azure/azure-functions-iothub-extension
https://github.com/Azure/azure-webjobs-sdk

It does look like [HTTP triggers](https://github.com/Azure/azure-webjobs-sdk-extensions/blob/dev/src/WebJobs.Extensions.Http/HttpTriggerAttributeBindingProvider.cs#L248) are also capable of an implicit "$return" value. However, from testing, as long as an output binding is explicitly defined, I don't see the implicit "$return" and returned value changing behavior, even with this PR.

However, one behavior change introduced with this PR is that when an HTTP output binding **is not** defined, we see the returned value is made into the http response. Before, it would return a "204". Now, it will return a "best effort" http object that is serialized in this buggy way: https://github.com/Azure/azure-functions-nodejs-worker/issues/228

I think we can either keep this behavior in this PR or hard-code behavior around functions with "orchestrationTrigger" bindings. Hard-coding is the safest option that guarantees no affect on HTTP :\
@fabiocav - could you weigh in?

src/FunctionInfo.ts
```diff
@@ -13,13 +13,15 @@ export class FunctionInfo {
...
   public httpOutputName: string;
+  public isHttp: boolean;
 
   constructor(metadata: rpc.IRpcFunctionMetadata) {
     this.name = <string>metadata.name;
     this.directory = <string>metadata.directory;
     this.bindings = {};
     this.outputBindings = {};
     this.httpOutputName = "";
+    this.isHttp = false;
```
ConnorMcMahon: nit: consider a name with trigger in the name, like `hasHttpTrigger`. `isHttp` is a little vague. Given the context of `FunctionInfo`, `isHttp` is probably clear enough to mean it uses an HTTP trigger, so if you prefer the name, keep it.

src/FunctionInfo.ts
```diff
@@ -28,13 +30,21 @@ export class FunctionInfo {
...
             this.outputBindings[name] = Object.assign(bindings[name], { converter: toRpcHttp });
           } else {
             this.outputBindings[name] = Object.assign(bindings[name], { converter: toTypedData });
           }
         });
+      
+      this.isHttp = Object.keys(bindings)
+        .filter(name => {
+          let type = bindings[name].type;
+          return type && type.toLowerCase() == 'httptrigger';
```
ConnorMcMahon: nit: should probably use `===`.

src/FunctionInfo.ts
```diff
@@ -28,13 +30,21 @@ export class FunctionInfo {
...
+        })
+        .length > 0;
```
ConnorMcMahon: Is it ever valid to have 2+ bindings with trigger? I'm assuming somewhere in the pipeline we are already handling this invalid case.

mhoeger: I don't think it's valid to have 2+ binding with trigger - assuming that is handled on the host end! I think it's ok to say > 0, although potentially less precise.

PR #278, opened 2020-01-13
test against host v3

mhoeger:
Cosmos DB client wasn't correctly reading document. Root cause: https://github.com/MicrosoftDocs/azure-docs/issues/23792

Changing test to be against v3 host and allowing adding test result publish action

### DIFF
--- a/setup-e2e-tests.ps1
+++ b/setup-e2e-tests.ps1
@@ -25,11 +25,11 @@ Remove-Item -Recurse -Force ./Azure.Functions.Cli -ErrorAction Ignore
 
 if (-not (Test-Path env:CORE_TOOLS_URL)) 
 { 
-  $env:CORE_TOOLS_URL = "https://functionsclibuilds.blob.core.windows.net/builds/2/latest/Azure.Functions.Cli.win-x86.zip"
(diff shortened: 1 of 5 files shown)


setup-e2e-tests.ps1
```diff
@@ -25,11 +25,11 @@ Remove-Item -Recurse -Force ./Azure.Functions.Cli -ErrorAction Ignore
 
 if (-not (Test-Path env:CORE_TOOLS_URL)) 
 { 
-  $env:CORE_TOOLS_URL = "https://functionsclibuilds.blob.core.windows.net/builds/2/latest/Azure.Functions.Cli.win-x86.zip"
+  $env:CORE_TOOLS_URL = "https://functionsclibuilds.blob.core.windows.net/builds/3/latest/Azure.Functions.Cli.win-x86.zip"
```
pragnagopa: Might be worth it to keep runs on V2 as well. Can we have test runts for V1 and V2 for now?

mhoeger: This is just for the "v2.x" branch of the Node.js worker (which only runs with V3 bits) - so running against V2 was not a valid scenario!

pragnagopa: gotcha!

PR #286, opened 2020-03-07
Add new SameSite property and tests

mhoeger:
Addresses https://github.com/Azure/azure-functions-host/issues/4890
Resolves https://github.com/Azure/azure-functions-host/issues/5384

Most of the changes in this PR is related to fixing the end to end tests to make more sense. RpcHttpConverters has the behavior-changing changes.

### DIFF
--- a/src/converters/RpcHttpConverters.ts
+++ b/src/converters/RpcHttpConverters.ts
@@ -96,10 +96,13 @@ function toRpcHttpCookie(inputCookie: Cookie): rpc.IRpcHttpCookie {
     // Resolve SameSite enum, a one-off
     let rpcSameSite: rpc.RpcHttpCookie.SameSite = rpc.RpcHttpCookie.SameSite.None;
     if (inputCookie && inputCookie.sameSite) {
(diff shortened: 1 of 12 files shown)


test/end-to-end/Azure.Functions.NodejsWorker.E2E/Azure.Functions.NodejsWorker.E2E/Utilities.cs
```diff
@@ -4,12 +4,6 @@
 using System;
```
mhoeger: Moving Http scenario helper methods to "HttpHelpers" not utilities

src/converters/RpcHttpConverters.ts
```diff
@@ -96,10 +96,13 @@ function toRpcHttpCookie(inputCookie: Cookie): rpc.IRpcHttpCookie {
     // Resolve SameSite enum, a one-off
```
mhoeger: This is the bulk of the change

PR #316, opened 2020-07-05
Fix to load prototype 'run' or 'index' as entry point if exported obj…

### LINKED ISSUE #315: Error if exported object has single non-function property, though prototype object has function 'run()'

- Invocation ID: e1bd289e-b080-4fbd-8d41-a91468f299ea
- Region: Japan East

#### Repro steps

Following code **without** setting "entryPoint" in function.json

```javascript
class A {
    constructor() {
        this.prop = true;
    }

    async run(context) {
        context.res = {
            body: "hello"
        };
    }
}

module.exports = new A();
```

#### Expected behavior

Prototype function  "run" is used as entry point.

#### Actual behavior

An error is occured:
`Error: The resolved entry point is not a function and cannot be invoked by the functions runtime. Make sure the function has been correctly exported.`

#### Known workarounds

To add another property, causes no errors. Function "run" is uesd as entry point.

```javascript
class B {
    constructor() {
        this.prop = true;
        this.prop2 = true;  // add this property
    }

    async run(context) {  // called as entry point
        context.res = {
            body: "hello"
        };
    }
}

module.exports = new B();
```

#### Related information 

At function `getEntryPoint()` in `FunctionLoader.ts`,  
the condition to judge what property use as entry point is insufficient.

```javascript

function getEntryPoint(f: any, entryPoint?: string): Function {
    if (isObject(f)) {
        var obj = f;
        if (entryPoint) {
            // the module exports multiple functions
            // and an explicit entry point was named
            f = f[entryPoint];
        }
        else if (Object.keys(f).length === 1) {    // <-- match even if Object.keys(f)[0] is not a function!!
            // a single named function was exported
            var name = Object.keys(f)[0];
            f = f[name];
        }
        else {
            // finally, see if there is an exported function named
            // 'run' or 'index' by convention
            f = f.run || f.index;
        }
        
    // ...
```

wkwkhautbois:
resolve #315 

Fix condition to consider whether the single property is function.

### DIFF
--- a/src/FunctionLoader.ts
+++ b/src/FunctionLoader.ts
@@ -55,15 +55,15 @@ export class FunctionLoader implements IFunctionLoader {
 function getEntryPoint(f: any, entryPoint?: string): Function {
     if (isObject(f)) {
         var obj = f;
+        let keys = Object.keys(f);
         if (entryPoint) {
             // the module exports multiple functions
             // and an explicit entry point was named
             f = f[entryPoint];
         }
-        else if (Object.keys(f).length === 1) {
+        else if (keys.length === 1 && isFunction(f[keys[0]])) {
             // a single named function was exported
-            var name = Object.keys(f)[0];
-            f = f[name];
+            f = f[keys[0]];
         }
         else {
             // finally, see if there is an exported function named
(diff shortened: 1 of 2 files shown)


mhoeger (approved):
✅  This looks great - thank you! And thank you for adding tests :)!!

Issue #327, opened 2020-09-02
Unblock Node.js 14 with warning (development only)


mhoeger:
DEVELOPMENT ONLY

PR #329, opened 2020-09-11
Node.js 14 preview

mhoeger:
Resolves https://github.com/Azure/azure-functions-nodejs-worker/issues/327

(Note: node.js 14 support not added for functions v2, only v3)

### DIFF
--- a/src/WorkerChannel.ts
+++ b/src/WorkerChannel.ts
@@ -101,7 +101,7 @@ export class WorkerChannel implements IWorkerChannel {
     if (this._v1WorkerBehavior) {
       if (version.startsWith("v12.")) {
         systemWarn("The Node.js version you are using (" + version + ") is not fully supported with Azure Functions V2. We recommend using one the following major versions: 8, 10.");
-      } else if (version.startsWith("14.")) {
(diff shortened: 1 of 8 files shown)


mhoeger:
Hey @anthonychu and @pragnagopa - mind reviewing? thanks!

pragnagopa (approved):
LGTM!

src/WorkerChannel.ts
```diff
@@ -101,6 +101,13 @@ export class WorkerChannel implements IWorkerChannel {
     if (this._v1WorkerBehavior) {
       if (version.startsWith("v12.")) {
         systemWarn("The Node.js version you are using (" + version + ") is not fully supported with Azure Functions V2. We recommend using one the following major versions: 8, 10.");
+      } else if (version.startsWith("14.")) {
```
anthonychu:
```suggestion
      } else if (version.startsWith("v14.")) {

```

image

anthonychu (changes requested):
I think the `v14.` comment is the only blocker. Not blocking but curious about only checking for specific LTS versions to reject but we don't seem to do anything about non-LTS versions like 11 or 13.

PR #336, opened 2020-09-23
[WIP] Avoid parsing numeric string to number

### LINKED ISSUE #215: Number written as string treat as number when call callActivity, so large number will be rounded
**Describe the bug**

The size of numbers that can be handled by the JavaScript number type is limited.
We sometimes treat large-number as BigInt or string for avoiding that limitation.

That large number was treated as number when I used stringified number as `callActivity`' s parameter.

**Investigative information**

- Durable Functions extension version: 
- durable-functions npm module version: 1.4.3
- Language (JavaScript/TypeScript) and version: JavaScript
- Node.js version: 12

***If deployed to Azure App Service***

- Timeframe issue observed:
- Function App name: 
- Function name(s):
- Region: japan-east
- Orchestration instance ID(s): 

**To Reproduce**

See repro repo
https://github.com/yamachu/repro-durable-functions-js-stringify-large-number

1. request [DurableFunctionsHttpStart](https://github.com/yamachu/repro-durable-functions-js-stringify-large-number/tree/master/DurableFunctionsHttpStart) with big_id parameter
2. see logs

https://github.com/yamachu/repro-durable-functions-js-stringify-large-number/blob/master/logs/view-azure-portal-logstream-curl-big_id-12345678901234567890.txt

This log was logged at Azure Portal LogStream.

**Expected behavior**

If `callActivity("activityName", "12345678901234567890")`, 
we got input "12345678901234567890" as string in "activityName" activity.

**Actual behavior**

If `callActivity("activityName", "12345678901234567890")`, 
we got input 12345678901234567000 as number in "activityName" activity.

**Known workarounds**

Add some characters "@", "n" or .... to stringified number, and remove these characters in activity.

**Additional context**

- Development environment (ex. Visual Studio): Visual Studio Code Insiders
- Links to source: https://github.com/yamachu/repro-durable-functions-js-stringify-large-number
- Additional bindings used
- Function invocation IDs

yamachu:
fix: https://github.com/Azure/azure-functions-durable-js/issues/215

As described in Issue above, if you pass a string of numbers as the binding value, they will be treated as numbers in the function.
That process is taking place ...
https://github.com/Azure/azure-functions-nodejs-worker/blob/ac130db668086b6ef75a2765ae3a69c4dcda4081/src/converters/RpcConverters.ts#L24

If this process fails, the `str` passed in is used as the return value of this function, or the `typedData` passed in the caller is treated as the return value.

### problem

[`JSON.parse` ](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse ) can parse Objects, Arrays and Numbers.
ref: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON

In Node.js environment we cannot treat large number(such as 12345678901234567890) correctly with `number` type, so we sometimes uses it stringified-number (ex: `'12345678901234567890'`).

In this product, `JSON.parse` processes the passed str without exception, so any stringified-numbers that we want to treat as strings will also be treated as numbers.

### solution

Test that the `str` passed in the regular expression is in a valid format as JSON. (does not support JSON completely...)

### does this change has breaking-channge?

Yes

And now, this change does not fully-supported JSON syntax...
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON#Full_JSON_syntax:~:text=Full%20JSON%20syntax,-The

Numbers, null, boolean will be affected.

### DIFF
--- a/src/converters/RpcConverters.ts
+++ b/src/converters/RpcConverters.ts
@@ -20,7 +20,7 @@ export function fromTypedData(typedData?: rpc.ITypedData, convertStringToJson: b
     if (str !== undefined) {
         if (convertStringToJson) {
             try {
-                if (str != null) {
+                if (str != null && isJsonFormedStr(str)) {
                     str = JSON.parse(str);
                 }
             } catch (err) { }
@@ -181,4 +181,9 @@ export function toNullableTimestamp(dateTime: Date | number | undefined, propert
         }
     }
     return undefined;
+}
+
+function isJsonFormedStr(str: string): boolean {
+    const maybeJsonFormedStrRegex = new RegExp(/^\{.*\}$|^\[.*\]$/sm);
+    return maybeJsonFormedStrRegex.test(str);
 }
--- a/test/RpcConvertersTests.ts
+++ b/test/RpcConvertersTests.ts
@@ -1,4 +1,4 @@
-import { toNullableBool, toNullableString, toNullableDouble, toNullableTimestamp, fromRpcTraceContext } from '../src/converters';
+import { toNullableBool, toNullableString, toNullableDouble, toNullableTimestamp, fromRpcTraceContext, fromTypedData } from '../src/converters';
 import { expect } from 'chai';
 import { AzureFunctionsRpcMessages as rpc } from '../azure-functions-language-worker-protobuf/src/rpc';
 import 'mocha';
@@ -161,5 +161,10 @@ describe('Rpc Converters', () => {
   it('does not convert undefined to NullableTimestamp', () => {
     let nullable = toNullableTimestamp(undefined, "test");
     expect(nullable && nullable.value).to.be.undefined;
+  });
+
+  it('numeric-string keeps string type', () => {
+    let numericString = fromTypedData({ string: '12345678901234567890' });
+    expect(numericString).to.equal('12345678901234567890');
   });
 })


davidmrdavid:
Hi @yamachu ,

Thank you so much for this contribution. Do you have a reproducer you could attach that works using (non-Durable) Azure Functions? For the Durable case, there's a fix we could apply as shown here: https://github.com/Azure/azure-functions-durable-js/pull/217

But if this also occurs at the Azure Functions level, then maybe it makes more sense to apply the change here as well. 

Looping @amdeel and @mhoeger, both of whom were part of an investigation around a similar, potentially related, JSON-parsing issue here: https://github.com/Azure/azure-functions-durable-extension/issues/1476

mhoeger:
Hi @yamachu , thanks so much for the contribution :) 

I think I echo @davidmrdavid 's suggestion that a fix in the azure-functions-durable-js would be a better fix. The biggest issue is that we cannot make the breaking change you pointed out except on a major version change as it would be very impactful to all customers.

In this area, I feel that JSON.parse has sufficient and expected behavior. Since the issue you faced will be fixed by David, I'm inclined to close this PR? But please let us know if you think otherwise!

One thing I noticed when experimenting with the repro you provided was that there is a workaround that could be applied without needing to append characters.

Here is an example from our "Hello World" template functions.

Orchestrator:
```typescript
    outputs.push(yield context.df.callActivity("Hello", "12345678901234567890"));
    outputs.push(yield context.df.callActivity("Hello", { number: "12345678901234567890" }));
```

Activity Function:
```typescript
    return `Hello ${JSON.stringify(context.bindings.name)}!`;
```

Output:
```
"output":["Hello 12345678901234567000!","Hello {\"number\":\"12345678901234567890\"}!"]
```

yamachu:
Hi @davidmrdavid @mhoeger 

Thank you for quick-reply!
I didn't reproduce in non-Durable Azure Functions...
I'm still trying to understand converters and other things related to `rpc.ITypedData`, so I'm going to use Azure Functions and if I have any concerns with Azure Functions, I'll do Issue and PR again. :)

I realized that the destructive changes were too big after submitted this PR.
I'll close this PR and wait for https://github.com/Azure/azure-functions-durable-js/pull/217 and other solutions.

When I wrapped it in Object as you mentioned, it worked!
I'm wondering what happens in the case of Durable Entity, so I'll give it a try.

Thanks

PR #338, opened 2020-10-01
feature: add before/after InvocationRequest hooks

markwolff:
Adds two `InvocationRequest` hooks to `WorkerChannel` to enable context correlation and node.js worker APM instrumentation. Sample use case for Azure Monitor node.js SDK.

- `invocationRequestBefore`: `(context, userFunction) => contextWrappedUserFunction`
- `invocationRequestAfter`: `() => void`

> note: `workerChannel` would need to be passed to app insights in some way, i.e. in `nodejsWorker.ts`:
```ts
if (shouldSetupAppInsightsFunctionsIntegration) {
  require("/path/to/applicationinsights/agent/functions")
    .setupFunctionsInvocationHooks(workerChannel);
}
```

In App Insights SDK, `setupFunctionsInvocationHooks` would be defined:
```ts
workerChannel.invocationRequestBefore = (context, userFunction) => {
    // Map Functions context to AppInsights context
    const correlationContext = appInsights.startOperation(context, "Functions.JavaScript.Trigger");

    // Wrap the Function runtime with correlationContext
    return appInsights.wrapWithCorrelationContext(userFunction, correlationContext);
}

workerChannel.invocationRequestAfter = () => {
    // Do any extra cleanup or flushing here
    appInsights.defaultClient.flush(); // can be async or sync
}
```

### DIFF
--- a/src/WorkerChannel.ts
+++ b/src/WorkerChannel.ts
@@ -11,6 +11,9 @@ import { Context } from './public/Interfaces';
 import LogCategory = rpc.RpcLog.RpcLogCategory;
 import LogLevel = rpc.RpcLog.Level;
 
+type InvocationRequestBefore = (context: Context, userFn: Function) => Function;
+type InvocationRequestAfter = (context: Context) => void;
+
 /**
  * The worker channel should have a way to handle all incoming gRPC messages.
  * This includes all incoming StreamingMessage types (exclude *Response types and RpcLog type)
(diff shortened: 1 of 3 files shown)


src/WorkerChannel.ts
```diff
@@ -25,12 +26,17 @@ interface IWorkerChannel {
...
+  invocationRequestBefore?: (context: Context, userFn: Function) => Function;
+  invocationRequestAfter?: Function;
 }
 
 /**
  * Initializes handlers for incoming gRPC messages on the client
  */
 export class WorkerChannel implements IWorkerChannel {
+  public invocationRequestBefore?: (context: Context, userFn: Function) => Function;
+  public invocationRequestAfter?: Function;
```
lzchen: Curious as to why `context` isn't exposed in `After` as well.

src/WorkerChannel.ts
```diff
@@ -183,19 +193,34 @@ export class WorkerChannel implements IWorkerChannel {
...
     // catch user errors from the same async context in the event loop and correlate with invocation
     // throws from asynchronous work (setTimeout, etc) are caught by 'unhandledException' and cannot be correlated with invocation
     try {
       let result = userFunction(context, ...inputs);
 
       if (result && isFunction(result.then)) {
-        result.then(result => (<any>context.done)(null, result, true))
-          .catch(err => (<any>context.done)(err, null, true));
+        result.then(result => {
+          if (this.invocationRequestAfter) {
```
lzchen: Are we able to call `this.invocationRequestAfter();` at the end of this whole function? Instead of putting multiple separate calls for possible outcomes of the result (regular, catching errors, catch outstanding errors).

lzchen:
> note: workerChannel would need to be passed to app insights in some way, i.e. in nodejsWorker.ts:

Instead of needing to pass in the workerChannel, why not expose some api to add a callback which would BE the actual before/after functions?

src/WorkerChannel.ts
markwolff: These are all happening in separate async paths which never converge on completion. I could see adding some `this.finish()` to each path to converge them. And within that we could call `invocationRequestAfter()`, but this would look pretty similar to what is already there. wdyt?

src/WorkerChannel.ts
markwolff: Agree. It seems like it would be required to correlate internal logging with the trace context

src/WorkerChannel.ts
lzchen: `this.finish()` would be more elegant but not really a blocking issue :)

PR #339, opened 2020-10-19
feature: add before/after InvocationRequest hooks

markwolff:
Copied from #338 

Adds two `InvocationRequest` hooks to `WorkerChannel` to enable context correlation and node.js worker APM instrumentation. Sample use case for Azure Monitor node.js SDK.

- `invocationRequestBefore`: `(context, userFunction) => contextWrappedUserFunction`
- `invocationRequestAfter`: `() => void`

> note: `workerChannel` would need to be passed to app insights in some way, i.e. in `nodejsWorker.ts`:
```ts
if (shouldSetupAppInsightsFunctionsIntegration) {
  require("/path/to/applicationinsights/agent/functions")
    .setupFunctionsInvocationHooks(workerChannel);
}
```

In App Insights SDK, `setupFunctionsInvocationHooks` would be defined:
```ts
workerChannel.invocationRequestBefore = (context, userFunction) => {
    // Map Functions context to AppInsights context
    const correlationContext = appInsights.startOperation(context, "Functions.JavaScript.Trigger");

    // Wrap the Function runtime with correlationContext
    return appInsights.wrapWithCorrelationContext(userFunction, correlationContext);
}

workerChannel.invocationRequestAfter = () => {
    // Do any extra cleanup or flushing here
    appInsights.defaultClient.flush(); // can be async or sync
}
```

### DIFF
--- a/src/WorkerChannel.ts
+++ b/src/WorkerChannel.ts
@@ -11,6 +11,9 @@ import { Context } from './public/Interfaces';
 import LogCategory = rpc.RpcLog.RpcLogCategory;
 import LogLevel = rpc.RpcLog.Level;
 
+type InvocationRequestBefore = (context: Context, userFn: Function) => Function;
+type InvocationRequestAfter = (context: Context) => void;
+
 /**
  * The worker channel should have a way to handle all incoming gRPC messages.
  * This includes all incoming StreamingMessage types (exclude *Response types and RpcLog type)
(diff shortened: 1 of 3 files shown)


src/WorkerChannel.ts
```diff
@@ -160,6 +178,8 @@ export class WorkerChannel implements IWorkerChannel {
         invocationId: msg.invocationId,
         result: this.getStatus(err)
       }
+      
+      this.runInvocationRequestAfter(context);
```
mhoeger: Remove this?

test/WorkerChannelTests.ts
```diff
@@ -226,67 +283,114 @@ describe('WorkerChannel', () => {
...
+  });
+
+  it ('invokes function', () => {
+    loader.getFunc.returns((context) => context.done());
+    loader.getInfo.returns({
+      name: 'test',
+      outputBindings: {}
+    })
+
+    assertInvokedFunction();
```
mhoeger: thanks for tests :)

src/WorkerChannel.ts
```diff
@@ -304,4 +333,18 @@ export class WorkerChannel implements IWorkerChannel {
 
     return status;
   }
+
+  private runInvocationRequestBefore(context: Context, userFunction: Function): Function {
+    let wrappedFunction = userFunction;
+    for (let before of this._invocationRequestBefore) {
+      wrappedFunction = before(context, wrappedFunction);
```
mhoeger, line 344: Just double checking that you don't want to make anything async? Could do a Promise.all

src/WorkerChannel.ts
markwolff: I can move this to be the `else` case for the `.then` Promise checks below if you think that makes sense. Is there a better place for this to live to handle UserFunctions that have callbacks?

src/WorkerChannel.ts
markwolff: I could see these being useful as Promises. Currently each `before` updates `wrappedFunction` on completion. Is there a nice way to accomplish this with Promise.all?

src/WorkerChannel.ts
mhoeger: Ahh sorry - pointed to the wrong parts to remove!! Can you remove the calls in line 221 and 225 to "runInvocationRequestAfter"? In lines 222 and 226, "context.done" is called, which is the same as this "resultCallback". So in cases of async / promise-returning user functions, "runInvocationRequestAfter" would be called twice

markwolff: Good catch! Refactored the tests a bit to support tests for promise user functions

src/WorkerChannel.ts
mhoeger: Thoughts on await on `before` and `after` calls? Otherwise this PR looks good!

mhoeger (approved):
✅ LGTM - on registering async functions, since it's not like a one line change, let's just cross that bridge when we get there.

Thanks!!

Issue #344, opened 2020-10-28
Add support for ES Modules


anthonychu:
Allow Node 14 functions to use ESM by supporting `.mjs` files.

A function would look something like this: 

```js
// index.mjs
export default async function (context, req) {
    context.res = {
        body: process.version
    };
};
```

fabiocav:
Work in progress. Moving to sprint 90

PR #350, opened 2020-11-17
Adding ability to use .mjs files for Node.js 14 (preview)

mhoeger:
Note that this only enables ES Modules through the `.mjs` file extension, which is 1 of 3 ways to enable ES Modules currently: https://nodejs.org/api/esm.html#esm_enabling (excludes the package.json "type" field and the --input-type flag)

If .msj is used with node.js version < 14.x, then the result will be a "500" not a "404" (same behavior if there are other issues loading a script file).

**TODO**: In host, need to update this line of code to also look for "index.mjs"
https://github.com/Azure/azure-functions-host/blob/dd062a5cc9e1f8e8bb875d3e6068b1df3865257a/src/WebJobs.Script/Host/FunctionMetadataProvider.cs#L245

**TODO:** Verify no performance regressions

### DIFF
--- a/src/FunctionLoader.ts
+++ b/src/FunctionLoader.ts
@@ -28,7 +28,7 @@ export class FunctionLoader implements IFunctionLoader {
           // use eval so it doesn't get compiled into a require()
           script = await eval("import(scriptFilePath)");
         } else {
-            throw new InternalException(`Please use a Node.js version >= v14 to use ES Modules for '${scriptFilePath}'`);
+          throw new InternalException(`Please use a Node.js version >= v14 to use ES Modules for '${scriptFilePath}'`);
         }
       } else {
         script = require(scriptFilePath);
--- a/src/FunctionLoader.ts
+++ b/src/FunctionLoader.ts
@@ -5,7 +5,7 @@ import { FunctionInfo } from './FunctionInfo'
 import { InternalException } from "./utils/InternalException";
 
 export interface IFunctionLoader {
-  load(functionId: string, metadata: rpc.IRpcFunctionMetadata): void;
+  load(functionId: string, metadata: rpc.IRpcFunctionMetadata): Promise<void>;
   getInfo(functionId: string): FunctionInfo;
   getFunc(functionId: string): Function;
 }
--- a/src/FunctionLoader.ts
+++ b/src/FunctionLoader.ts
@@ -15,13 +15,24 @@ export class FunctionLoader implements IFunctionLoader {
         info: FunctionInfo,
         func: Function
     }} = {};
+    private allowESModules = process.version.startsWith("v14");
 
-    load(functionId: string, metadata: rpc.IRpcFunctionMetadata): void {
+    async load(functionId: string, metadata: rpc.IRpcFunctionMetadata): Promise<void> {
       if (metadata.isProxy === true) {
           return;
       }
       let scriptFilePath = <string>(metadata && metadata.scriptFile);
-      let script = require(scriptFilePath);
+      let script: any;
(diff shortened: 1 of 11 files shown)


test/end-to-end/testFunctionApp/HttpTriggerESModules/function.json
```diff
@@ -0,0 +1,21 @@
...
+        "post"
+      ]
+    },
+    {
+      "type": "http",
+      "direction": "out",
+      "name": "res"
+    }
+  ],
+  "scriptFile": "./index.mjs"
```
mhoeger, line 20: This is required for now - need to make [change in host](https://github.com/Azure/azure-functions-host/blob/dd062a5cc9e1f8e8bb875d3e6068b1df3865257a/src/WebJobs.Script/Host/FunctionMetadataProvider.cs#L245)

test/WorkerChannelTests.ts
```diff
@@ -211,14 +211,16 @@ describe('WorkerChannel', () => {
...
+  it('responds to function load', async () => {
     stream.addTestMessage({
       requestId: 'id',
       functionLoadRequest: {  
         functionId: 'funcId',
         metadata: { }
       }
     });
+    // Set slight delay 
+    await new Promise(resolve => setTimeout(resolve, 100));
```
mhoeger, line 223: I'd love to find a better way to do this suggestions very welcome :|

src/FunctionLoader.ts
```diff
@@ -15,13 +16,31 @@ export class FunctionLoader implements IFunctionLoader {
...
           return;
       }
       let scriptFilePath = <string>(metadata && metadata.scriptFile);
-      let script = require(scriptFilePath);
+      let script: any;
+      if (scriptFilePath.endsWith(".mjs")) {
+        if (this.allowESModules) {
+          // IMPORTANT: pathToFileURL is only supported in Node.js version >= v10.12.0
+          // @ts-ignore
+          let scriptFileUrl = url.pathToFileURL(scriptFilePath);
```
mhoeger, line 31: "import" takes a file URL and not a path... this behavior only shows up on windows though. It's recommended to use pathToFileURL per Node.js devs

src/FunctionLoader.ts
```diff
@@ -15,13 +16,31 @@ export class FunctionLoader implements IFunctionLoader {
...
+          if (scriptFileUrl.href) {
+            // use eval so it doesn't get compiled into a require()
+            script = await eval("import(scriptFileUrl.href)");
+          } else {
+            throw new InternalException(`'${scriptFilePath}' could not be converted to file URL (${scriptFileUrl.href})`);
```
mhoeger, line 36: I don't think this will happen but ehh!

anthonychu (approved):
Tested it out with a few things with a .mjs function locally and it's working as expected:

* import npm module
* import .js file
* import .mjs file

.js functions still working as well.

anthonychu:
Wonder if there should be a warning/info message printed if there’s at least one mjs file to say the feature is in preview.

pragnagopa:
@mhoeger -  following is legacy from V1 - never got a chance a delete it. Worker should be handling default primary file if script file property is not set in function.json , Please do open a separate issue for this and do let me know if have any ideas on how to handle this in a non-breaking way.

> TODO: In host, need to update this line of code to also look for "index.mjs"
https://github.com/Azure/azure-functions-host/blob/dd062a5cc9e1f8e8bb875d3e6068b1df3865257a/src/WebJobs.Script/Host/FunctionMetadataProvider.cs#L245

mhoeger:
Resolves #344

PR #359, opened 2020-12-21
fix: use interfaces to allow declaration merging

G-Rath:
This is sort of the equivalent to [this PR](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/49973) that I made for `@types/aws-lambda`.

By using interfaces we can [declaration merging](https://www.typescriptlang.org/docs/handbook/declaration-merging.html) to type expected input values for our functions even further.

For example, I'm currently creating a GitHub webhook on Azure Functions, which has some headers.
With this change, I'll be able to do this:

```
declare module '@azure/functions' {
  export interface HttpRequestHeaders {
    'X-GitHub-Event': GithubEventName;
    'X-GitHub-Delivery': string;
    'X-Hub-Signature': string;
    'X-Hub-Signature-256': string;
  }
}
```

### DIFF
--- a/types/public/Interfaces.d.ts
+++ b/types/public/Interfaces.d.ts
@@ -10,6 +10,12 @@
  * result will be passed to JSON.stringify unless it is a string, Buffer, ArrayBufferView, or number.
  */
 export declare type AzureFunction = ((context: Context, ...args: any[]) => Promise<any> | void);
+export interface ContextBindings {
+    [name: string]: any;
+}
+export interface ContextBindingData {
+    [name: string]: any;
+}
 /**
  * The context object can be used for writing logs, reading data from bindings, setting outputs and using
  * the context.done callback when your exported function is synchronous. A context object is passed
@@ -28,15 +34,11 @@ export interface Context {
      * Input and trigger binding data, as defined in function.json. Properties on this object are dynamically
      * generated and named based off of the "name" property in function.json.
      */
-    bindings: {
-        [key: string]: any;
-    };
+    bindings: ContextBindings;
     /**
      * Trigger metadata and function invocation data.
      */
-    bindingData: {
-        [key: string]: any;
-    };
+    bindingData: ContextBindingData;
     /**
      * TraceContext information to enable distributed tracing scenarios.
      */
@@ -71,6 +73,15 @@ export interface Context {
         [key: string]: any;
     };
 }
+export interface HttpRequestHeaders {
+    [name: string]: string;
+}
+export interface HttpRequestQuery {
+    [name: string]: string;
+}
+export interface HttpRequestParams {
+    [name: string]: string;
+}
 /**
  * HTTP request object. Provided to your function when using HTTP Bindings.
  */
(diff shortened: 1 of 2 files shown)


G-Rath:
I think the type should also be `string | undefined` instead of just `string`, but will save that for a follow up PR (or this one after it's been reviewed).

types/public/Interfaces.d.ts
```diff
@@ -10,6 +10,12 @@
  * result will be passed to JSON.stringify unless it is a string, Buffer, ArrayBufferView, or number.
  */
 export declare type AzureFunction = ((context: Context, ...args: any[]) => Promise<any> | void);
+export interface ContextBindings {
+    [name: string]: any;
+}
+export interface ContextBindingsData {
```
mhoeger: Nit: Maybe drop the "s" for `ContextBindingData`? 
Consistent with "bindings" => ContextBindings and "bindingData" => ContextBindingData

mhoeger (commented):
Hi @G-Rath - Very sorry about the delay and thanks for the contribution! I have one nit comment but otherwise looks good

types/public/Interfaces.d.ts
G-Rath: Opps, good spotting! 😅

mhoeger (approved):
✅

G-Rath:
@mhoeger thanks!

How would you feel about changing the type to be `any | undefined`?

mhoeger:
Hmm! @G-Rath I don't think there's a case where it would ever be `undefined`. `bindingData` always has at least [these properties](https://docs.microsoft.com/en-us/azure/azure-functions/functions-reference-node?tabs=v2#contextbindingdata-property) and `bindings` would always contain event data from the trigger (and there's no function without a trigger event). Any limitations you have with just `any` or was that more a suggestion for trying to give more information about the object than "any"?

G-Rath:
Actually, you'd want to change it to `unknown` 😅 

My thinking is that with `any`, you get no type checking - so while we know that any binding being referenced *should* exist since that's the whole reason we're referencing them, until now we've not been able to tell TypeScript anything about those bindings.

Now however we can, using declaration merging - so by changing the type to `unknown`, you'd encourage people to define their bindings for TypeScript and that'd make your code safer.

i.e

```
interface MyBoundContext {
  value: string;
}

declare module '@azure/functions' {
  export interface ContextBindings {
    myBoundContext: MyBoundContext 
  }
}

const fn = (bindings: ContextBindings) => {
  console.log(bindings.myBoundContext.value.toUpperCase());
  // ts: error!
  console.log(bindings.undefinedContext.value.toUpperCase());
}
```

(I've not actually done a lot of work with Azure functions yet, so don't have a good example of the type for a binding)

mhoeger:
That's a good point, thanks! Today we are on TypeScript 2.0 though, so I think this counts as future work when we make an upgrade to TypeScript 3.0 (as @davidmrdavid is also [suggesting](https://github.com/Azure/azure-functions-nodejs-worker/issues/365))

Around the time we first published our types, we got some advice that staying on 2.0 would actually be a good choice, since it could be more broadly adopted by existing projects. We haven't revisited this in a while though and I think it's a good call to!

PR #366, opened 2021-01-25
fix: rebuild types

G-Rath:
@mhoeger sorry, I realised I didn't run `npm run build` for #359 😞

### DIFF
--- a/types/public/Interfaces.d.ts
+++ b/types/public/Interfaces.d.ts
@@ -74,13 +74,13 @@ export interface Context {
     };
 }
 export interface HttpRequestHeaders {
-    [name: string]: string;
+    [name: string]: string | undefined;
 }
 export interface HttpRequestQuery {
-    [name: string]: string;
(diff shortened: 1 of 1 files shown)


mhoeger (approved):
That's my bad too - thanks for catching :)

G-Rath:
If you like, I could setup a basic github action to check if PRs need to run this :)

mhoeger:
Love it! Yeah, I would be very grateful :) We generally use Azure DevOps over GitHub Actions across our repos, so a change to [this yaml](https://github.com/Azure/azure-functions-nodejs-worker/blob/v2.x/azure-pipelines.yml) is more consistent with what we've done, but anything you add I would add to the DevOps pipeline anyways. Thanks!!

G-Rath:
@mhoeger interesting - based off that config, you should already have hecks running on PRs that'd catch this; so I'm guessing the issue is actually the CI isn't configured properly to come through on GitHub?

PR #372, opened 2021-01-29
feat: support optional generic for return type of `AzureFunction`

G-Rath:
This lets us do:

```
export const handler: AzureFunction<FnResponse> = async (
  context,
  request: HttpRequest
) => {
```

This can be improved further by making `args` a generic too:

```
export declare type AzureFunction<TReturn = any, TArgs extends any[] = any[]> = ((context: Context, ...args: TArgs) => Promise<TReturn> | void);
```

which lets us do

```
export const handler: AzureFunction<FnResponse, [HttpRequest]> = async (
  context,
  request // has type HttpRequest
) => {
```

However, this requires higher than TypeScript 2 (maybe lower than 3.3, but the playground doesn't go lower)

### DIFF
--- a/src/public/Interfaces.ts
+++ b/src/public/Interfaces.ts
@@ -9,7 +9,7 @@
  * @returns Output bindings (optional). If you are returning a result from a Promise (or an async function), this 
  * result will be passed to JSON.stringify unless it is a string, Buffer, ArrayBufferView, or number.
  */
-export type AzureFunction = ((context: Context, ...args: any[]) => Promise<any> | void);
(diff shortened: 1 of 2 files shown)


ejizba:
Thank you for contributing! This was adopted in https://github.com/Azure/azure-functions-nodejs-worker/pull/374#discussion_r573406165, but never released. I have removed the change in https://github.com/Azure/azure-functions-nodejs-worker/pull/438 with some of my reasoning mentioned in that PR. If you are still interested in this change, please file a new issue and we can have a full discussion on the advantages and disadvantages before doing a PR

PR #374, opened 2021-02-10
Upgrading to TS 4.1.3

### LINKED ISSUE #365: Any thoughts on upgrade our TypeScript version?
As per our [package.json](https://github.com/Azure/azure-functions-nodejs-worker/blob/v2.x/package.json#L29), we are currently pinned to TypeScript version  `2.x`. This prevents us from having neat tools like a [linter rule](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/strict-boolean-expressions.md) that would warn us of _suspicious_ if-statements on `falsy` values. Something like this would have helped us prevent [this](https://github.com/Azure/azure-functions-durable-js/issues/50) error which is caused by [this](https://github.com/Azure/azure-functions-nodejs-worker/blob/4247c6b39d7b48720b5f36dbcb4ce42461360b4a/src/WorkerChannel.ts#L233) line 😄 .

Wondering if there's a known reason why we haven't made the upgrade or if its just a matter or upgrading and fixing the expected type mismatches that would now show up? If its the latter, I'd be happy to help make the upgrade as it would probably be a positive change across the board. Please let me know your thoughts!

_Thanks!_

cc/ @mhoeger

alrod:
resolves #365

### DIFF
--- a/src/public/Interfaces.ts
+++ b/src/public/Interfaces.ts
@@ -1,17 +1,4 @@
 /**
- * Interface for your Azure Function code. This function must be exported (via module.exports or exports) 
- * and will execute when triggered. It is recommended that you declare this function as async, which 
- * implicitly returns a Promise.
- * @param context Context object passed to your function from the Azure Functions runtime.
- * @param {any[]} args Optional array of input and trigger binding data. These binding data are passed to the 
- * function in the same order that they are defined in function.json. Valid input types are string, HttpRequest, 
- * and Buffer.
- * @returns Output bindings (optional). If you are returning a result from a Promise (or an async function), this 
- * result will be passed to JSON.stringify unless it is a string, Buffer, ArrayBufferView, or number.
- */
-export declare type AzureFunction<TReturn = any, TArgs extends any[] = any[]> = ((context: Context, ...args: TArgs) => Promise<TReturn> | void);
-
-/**
  * Context bindings object. Provided to your function binding data, as defined in function.json.
  */
 export interface ContextBindings {
@@ -22,6 +9,7 @@ export interface ContextBindings {
  * Context binding data. Provided to your function trigger metadata and function invocation data.
 */
 export interface ContextBindingData {
+  invocationId?: string | null;
   [name: string]: any;
 }
 
@@ -250,3 +238,4 @@ export interface Logger {
      */
     verbose(...args: any[]): void;
 }
+
--- a/src/public/Interfaces.ts
+++ b/src/public/Interfaces.ts
@@ -9,14 +9,20 @@
  * @returns Output bindings (optional). If you are returning a result from a Promise (or an async function), this 
  * result will be passed to JSON.stringify unless it is a string, Buffer, ArrayBufferView, or number.
  */
-export type AzureFunction = ((context: Context, ...args: any[]) => Promise<any> | void);
+export type AzureFunction<TReturn = unknown> = ((context: Context, ...args: any[]) => Promise<TReturn> | void);
 
+/**
+ * Context bindings object. Provided to your function binding data, as defined in function.json.
+ */
 export interface ContextBindings {
-  [name: string]: any;
+  [name: string]: unknown;
 }
 
+/** 
+ * Context binding data. Provided to your function trigger metadata and function invocation data.
+*/
 export interface ContextBindingData {
-  [name: string]: any;
+  [name: string]: unknown;
 }
 
 /**
@@ -75,16 +81,25 @@ export interface Context {
     res?: { [key: string]: any };
 }
 
-export interface HttpRequestHeaders {
-  [name: string]: string | undefined;
+/**
+ * HTTP request headers.
+ */
+export interface HttpHeaders {
+  [name: string]: string | unknown;
 }
 
+/**
+ * Query string parameter keys and values from the URL.
+ */
 export interface HttpRequestQuery {
-  [name: string]: string | undefined;
+  [name: string]: string | unknown;
 }
 
+/**
+ * Route parameter keys and values.
+ */
 export interface HttpRequestParams {
-  [name: string]: string | undefined;
+  [name: string]: string | unknown;
 }
 
 /**
@@ -102,7 +117,7 @@ export interface HttpRequest {
     /**
      * HTTP request headers.
      */
-    headers: HttpRequestHeaders;
+    headers: HttpHeaders;
     /**
      * Query string parameter keys and values from the URL.
      */
--- a/src/public/Interfaces.ts
+++ b/src/public/Interfaces.ts
@@ -15,14 +15,14 @@ export declare type AzureFunction<TReturn = any, TArgs extends any[] = any[]> = 
  * Context bindings object. Provided to your function binding data, as defined in function.json.
  */
 export interface ContextBindings {
-  [name: string]: unknown;
+  [name: string]: any;
 }
 
 /** 
  * Context binding data. Provided to your function trigger metadata and function invocation data.
 */
 export interface ContextBindingData {
-  [name: string]: unknown;
+  [name: string]: any;
 }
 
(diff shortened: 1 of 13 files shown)


src/public/Interfaces.ts
```diff
@@ -9,14 +9,20 @@
...
  */
-export type AzureFunction = ((context: Context, ...args: any[]) => Promise<any> | void);
+export type AzureFunction<TReturn = unknown> = ((context: Context, ...args: any[]) => Promise<TReturn> | void);
 
+/**
+ * Context bindings object. Provided to your function binding data, as defined in function.json.
+ */
 export interface ContextBindings {
-  [name: string]: any;
+  [name: string]: unknown;
```
alrod: changing from any -> unknown is a breaking change.
This will encourage customers to extend public interfaces to include properties they are using in the code.

src/public/Interfaces.ts
```diff
@@ -9,14 +9,20 @@
  * @returns Output bindings (optional). If you are returning a result from a Promise (or an async function), this 
  * result will be passed to JSON.stringify unless it is a string, Buffer, ArrayBufferView, or number.
  */
-export type AzureFunction = ((context: Context, ...args: any[]) => Promise<any> | void);
+export type AzureFunction<TReturn = unknown> = ((context: Context, ...args: any[]) => Promise<TReturn> | void);
```
mhoeger: @alrod - can we do what @G-Rath suggested [here](https://github.com/Azure/azure-functions-nodejs-worker/pull/372#issue-563688769) now that we're bumping typescript version? 
```
export declare type AzureFunction<TReturn = any, TArgs extends any[] = any[]> = ((context: Context, ...args: TArgs) => Promise<TReturn> | void);
```

src/public/Interfaces.ts
```diff
@@ -75,16 +81,25 @@ export interface Context {
     res?: { [key: string]: any };
 }
 
-export interface HttpRequestHeaders {
-  [name: string]: string | undefined;
+/**
+ * HTTP request headers.
+ */
+export interface HttpHeaders {
+  [name: string]: string | unknown;
```
mhoeger: I think this is actually `[name: string]: string`

Headers, query, and params will always be a string (even if that string is `""`) per this change: https://github.com/Azure/azure-functions-nodejs-worker/commit/544f7e028409d8be54de85d81ba69d4dbeaf5138

src/public/Interfaces.ts
```diff
@@ -75,16 +81,25 @@ export interface Context {
...
 }
 
+/**
+ * Query string parameter keys and values from the URL.
+ */
 export interface HttpRequestQuery {
-  [name: string]: string | undefined;
+  [name: string]: string | unknown;
```
mhoeger: Same: `[name: string]: string`

src/public/Interfaces.ts
```diff
@@ -75,16 +81,25 @@ export interface Context {
...
 }
 
+/**
+ * Route parameter keys and values.
+ */
 export interface HttpRequestParams {
-  [name: string]: string | undefined;
+  [name: string]: string | unknown;
```
mhoeger: ~Same: `[name: string]: string`~ Should remain `[name: string]: string | undefined`

src/public/Interfaces.ts
mhoeger: Any reason in particular to change this from `HttpRequestHeaders` => `HttpHeaders`?

mhoeger (commented):
Thanks for the extra interface comments and the tests!! :) 

One suggestion on the `AzureFunction` type and some requested changes to HTTP header/query/params

src/public/Interfaces.ts
G-Rath: So the reason I made these `| undefined` is because otherwise you can access "any" property without confirming it actually exists - this is meant to gel with being able to use declaration merging to define properties that you know will exist.

i.e

```
import { HttpRequest } from '@azure/functions';

declare module '@azure/functions' {
  export interface HttpRequestHeaders {
    'x-github-event': GithubEventName;
    'x-github-delivery': string;
    'x-hub-signature': string;
    'x-hub-signature-256': string;
  }
}

export const getEvent = (request: HttpRequest): GithubEvent => {
  const rawBody = request.rawBody as string;
  const signature: string = request.headers['x-hub-signature-256'];
  const { GH_WEBHOOK_SECRET } = process.env;

  if (verify(GH_WEBHOOK_SECRET, rawBody, signature)) {
    // ensure we're referencing a header that is defined, outside of the cast
    const name = request.headers['x-github-event'];
    const payload = JSON.parse(rawBody) as GithubEvent['payload'];

    return { name, payload } as GithubEvent;
  }

  throw new Error('event did not come from github');
};
```

In the above, I don't have to do any checking on if the headers are `undefined` or not before using them as I've explicitly typed them as being just of type `string` (effectively I've told TS that they'll always be there).

Removing the `| undefined` from this signature will mean I could mistakenly typo a header and not be told, i.e if I did `X-GitHub-Event` (which is how the docs refer to the header) TS will give me a error alerting me to the typo (which is accurate as Azure downcases the headers).

This actually happened when I was originally building my azure function, as the above code was from an AWS lambda which preserves the headers original casing.

mhoeger: Ohhh sure - great point!! 

@alrod - could you change back to `[name: string]: string | undefined`? In these cases it's always known to only be a string value or undefined (nothing more, nothing less)

src/public/Interfaces.ts
mhoeger: EDIT: Should remain `[name: string]: string | undefined`, I'm wondering if it should remain HttpRequestHeaders too!

types/public/Interfaces.d.ts
```diff
@@ -9,12 +9,18 @@
  * @returns Output bindings (optional). If you are returning a result from a Promise (or an async function), this
  * result will be passed to JSON.stringify unless it is a string, Buffer, ArrayBufferView, or number.
  */
-export declare type AzureFunction = ((context: Context, ...args: any[]) => Promise<any> | void);
+export declare type AzureFunction<TReturn = any, TArgs extends any[] = any[]> = ((context: Context, ...args: TArgs) => Promise<TReturn> | void);
```
G-Rath, line 12:
```suggestion
export declare type AzureFunction<TReturn = unknown, TArgs extends unknown[] = unknown[]> = ((context: Context, ...args: TArgs) => Promise<TReturn> | void);
```

These can be `unknown` instead of `any` :)

src/public/Interfaces.ts
```diff
@@ -9,14 +9,20 @@
  * @returns Output bindings (optional). If you are returning a result from a Promise (or an async function), this 
  * result will be passed to JSON.stringify unless it is a string, Buffer, ArrayBufferView, or number.
  */
-export type AzureFunction = ((context: Context, ...args: any[]) => Promise<any> | void);
+export declare type AzureFunction<TReturn = any, TArgs extends any[] = any[]> = ((context: Context, ...args: TArgs) => Promise<TReturn> | void);
```
G-Rath:
```suggestion
export declare type AzureFunction<TReturn = unknown, TArgs extends unknown[] = unknown[]> = ((context: Context, ...args: TArgs) => Promise<TReturn> | void);
```

These can be `unknown` instead of `any` :)

src/public/Interfaces.ts
G-Rath: Do you want to change any other `any`s? as theres a couple of others in this file :)

alrod: After chatting with @mhoeger offline we decided to leave it as `any`. Changing to `unknown `will be quite hard and we do not want to introduce breaking changes for the customers code.

mhoeger (approved):
✅

PR #375, opened 2021-02-11
Add missing originalUrl to HttpRequest interface

dhensby:
The `HttpRequest` interface is missing the `originalUrl` property as per the docs and runtime.

See https://docs.microsoft.com/en-us/azure/azure-functions/functions-reference-node?tabs=v2#request-object

| Property | Description |
| -------- | ----------- |
| *body*   | An object that contains the body of the request. |
| *headers* | An object that contains the request headers. |
| *method* | The HTTP method of the request. |
| *originalUrl* | The URL of the request. |
| *params* | An object that contains the routing parameters of the request. |
| *query* | An object that contains the query parameters. |
| *rawBody* | The body of the message as a string. |

### DIFF
--- a/types/public/Interfaces.d.ts
+++ b/types/public/Interfaces.d.ts
@@ -95,6 +95,10 @@ export interface HttpRequest {
      */
     url: string;
     /**
+     * Raw request url
+     */
+    originalUrl: string,
+    /**
      * HTTP request headers.
      */
     headers: HttpRequestHeaders;


ejizba:
Closing this for a few reasons:
1. It's out of date with the v2.x branch after I did a few refactors
2. The `originalUrl` property is the exact same as the `url` property (see [here](https://github.com/Azure/azure-functions-nodejs-worker/blob/736a9cbd0ab42139085e0fbc320e533a94d75108/src/converters/RpcHttpConverters.ts#L25)) and I don't think we need both in the types. If you still want the property for some reason, please file a new issue and we can discuss

Issue #169, opened 2019-01-10
Add an interface for the HTTP response


mhoeger:
We should add a TypeScript interface for an HTTP response object. This gets tricky because we support both express-like and http-module-like ways of crafting a Response.

See discussion here:
https://github.com/Azure/azure-functions-nodejs-worker/pull/160

IGx89:
Would love this! Also, FYI, currently the response object is defined as optional which seems to be incorrect according to discussion in #160. It misleading suggests that `if (!context.res) { ... }` is a valid way to check if a response has been set, which isn't.

mhoeger:
@IGx89 - the response object is optional because in non-HTTP scenarios, there is no context.res (ex: functions that are queue-triggered, cosmos DB-triggered, event grid-triggered, etc.). But you are right that the res object is always there for HTTP scenarios (which is most common for JS developers), and that function code should never be checking `if (context.res)`

IGx89:
Ha, good point! Completely forgot about the non-HTTP scenarios. If you don't mind me asking here, what is the proper way to check if an HTTP response (`context.res = {}`) has already been set or not?

ShawnTalbert:
> @IGx89 - the response object is optional because in non-HTTP scenarios, there is no context.res (ex: functions that are queue-triggered, cosmos DB-triggered, event grid-triggered, etc.). But you are right that the res object is always there for HTTP scenarios (which is most common for JS developers), and that function code should never be checking `if (context.res)`

Is there a reason we're not considering interface inheritance for this? 

As an aside, perhaps this is a bit of a design smell around the abstraction function bindings are attempting to provide.

fabiocav:
@mhoeger do you have an update on this item?

G-Rath:
According to the [docs](https://docs.microsoft.com/en-us/azure/azure-functions/functions-reference-node?tabs=v2#response-object) the `res` interface should have `body`, `headers`, `isRaw`, `status`, and `cookies`.

Would be great to get an interface for this landed to compliment `HttpRequest` - even if it needs interaction, having an interface would serve as a starting point.

(also, in my testing I've found that `statusCode` works too)

hossam-nasr:
Partially addressed in #529. Closing this issue in favor of #551

PR #376, opened 2021-02-11
Add HttpResponse interface

dhensby:
The `Context.res?` property is represented as an unbound map of values (`[key: string]: any`) when the shape of a the response is more specific than that.

In addition to this I've added some undocumented (but present) functions:

When logging the shape of a standard `Response` from a function execution (and when looking at the source - https://github.com/Azure/azure-functions-nodejs-worker/blob/v2.x/src/http/Response.ts) we can see that these functions do form part of the internal interface

`Response {
  headers: {},
  cookies: [],
  send: [Function: end],
  header: [Function: setHeader],
  set: [Function: setHeader],
  get: [Function: getHeader],
  _done: [Function] }`

---

see https://docs.microsoft.com/en-us/azure/azure-functions/functions-reference-node?tabs=v2#response-object

| Property | Description |
| -------- | ----------- |
| *body*   | An object that contains the body of the response. |
| *headers* | An object that contains the response headers. |
| *isRaw* | Indicates that formatting is skipped for the response. |
| *status* | The HTTP status code of the response. |
| *cookies* | An array of HTTP cookie objects that are set in the response. An HTTP cookie object has a `name`, `value`, and other cookie properties, such as `maxAge` or `sameSite`. |

### DIFF
--- a/types/public/Interfaces.d.ts
+++ b/types/public/Interfaces.d.ts
@@ -96,6 +96,22 @@ export interface HttpResponse {
      * An HTTP cookie object has a name, value, and other cookie properties, such as maxAge or sameSite.
      */
     cookies?: Cookie[],
+    /**
+     * A helper to access headers on the response
+     */
+    get(header: string): string | undefined,
+    /**
+     * A helper to set headers on the response
+     */
+    set(header: string, value?: string): void,
+    /**
+     * A helper to set headers on the response
+     */
+    header(header: string, value?: string): void,
+    /**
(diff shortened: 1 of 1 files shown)


ejizba:
Hello, thanks for contributing! I've been working on the types package recently and this PR is a bit out of date so I'm going to close it. That being said, we've had plenty of people ask for types on the response object and so I hope to get this fixed soon. Tracking with https://github.com/Azure/azure-functions-nodejs-worker/issues/169

PR #383, opened 2021-03-26
Serialize falsy values for Durable Functions

davidmrdavid:
Replaces: https://github.com/Azure/azure-functions-nodejs-worker/pull/364

Due to legacy behavior, falsy values ("", false, 0) in JS currently do not get serialized. We cannot patch this behavior without breaking some JS customers, but also leaving this behavior unpatched breaks Durable Functions. In this PR, I try to find a middle-ground by ensuring that falsy values get serialized only when returning to the Durable Functions output binding.

### DIFF
--- a/src/WorkerChannel.ts
+++ b/src/WorkerChannel.ts
@@ -228,8 +228,10 @@ export class WorkerChannel implements IWorkerChannel {
       response.outputData = [];
 
       // As legacy behavior, falsy values get serialized to `null` in AzFunctions.
-      // This breaks Durable Functions expectations so we check if we're serializing
-      // for durable and, if so, ensure falsy values get serialized.
+      // This breaks Durable Functions expectations, where customers expect any
+      // JSON-serializable values to be preserved by the framework,
+      // so we check if we're serializing for durable and, if so, ensure falsy
+      // values get serialized.
       let isDurableBinding = info?.bindings?.name?.type == 'activityTrigger';
 
       try {
--- a/src/WorkerChannel.ts
+++ b/src/WorkerChannel.ts
@@ -226,11 +226,12 @@ export class WorkerChannel implements IWorkerChannel {
       // explicitly set outputData to empty array to concat later
       response.outputData = [];
 
+      let returnBinding = info.getReturnBinding();
+      let isDurableBinding = returnBinding === 'orchestrationTrigger';
       try {
-        if (result) {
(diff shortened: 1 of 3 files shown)


src/WorkerChannel.ts
```diff
@@ -226,11 +226,16 @@ export class WorkerChannel implements IWorkerChannel {
       // explicitly set outputData to empty array to concat later
       response.outputData = [];
 
+      // As legacy behavior, falsy values get serialized to `null` in AzFunctions.
+      // This breaks Durable Functions expectations so we check if we're serializing
+      // for durable and, if so, ensure falsy values get serialized.
+      let isDurableBinding = info?.bindings?.name?.type == 'activityTrigger';
```
davidmrdavid, line 235: I still need to test if sub-orchestrations also need this patch

src/WorkerChannel.ts
```diff
@@ -227,11 +227,16 @@ export class WorkerChannel implements IWorkerChannel {
...
       response.outputData = [];
 
+      // As legacy behavior, falsy values get serialized to `null` in AzFunctions.
+      // This breaks Durable Functions expectations so we check if we're serializing
+      // for durable and, if so, ensure falsy values get serialized.
+      let isDurableBinding = info?.bindings?.name?.type == 'activityTrigger';
+
       try {
-        if (result) {
+        if (result || (isDurableBinding && result != null)) {
```
alrod, line 238: if `isDurableBinding == true` the `result` can be (0, false, undef, '') ? Can you please add a comment describing why it happens for durable but not for regular executions.

davidmrdavid: That's correct, in durable bindings such as `activityTrigger`s, the result can be 0, false, and "". This is because Durable Function activities are meant to be abstractions similar to regular function invocations, so the programming model should, in theory, allow for any output value that would be valid to return in regular JavaScript _and_ is JSON serializable. This includes the values: 0, false, and `""`. This was also a behavior we used to have in the past, **but it stopped working recently**, as described here: https://github.com/Azure/azure-functions-durable-js/issues/50

I've also expanded my comments in the PR to make this clearer :) Please let me know if there's anything else I can clarify!

As for serializing `undefined`, I think that gets serialized as `null` anyways. I can certainly write code to explicitly avoid serializing `undefined` though. Please let me know if you would prefer I added that.

alrod (approved):
please squash to one commit

PR #414, opened 2021-09-17
Fix binding data serialization

alrod:
V4 breaking change proposal:
https://github.com/Azure/Azure-Functions/issues/2007

### DIFF
--- a/src/converters/BindingConverters.ts
+++ b/src/converters/BindingConverters.ts
@@ -46,8 +46,11 @@ function isBindingDirection(input: string | undefined): boolean {
 export function convertKeysToCamelCase(obj: any) {
   var output = {};
   for (var key in obj) {
+      // Only "undefined" will be replaced with original object property. For example:
+      //{ string : "0" } -> 0
+      //{ string : "false" } -> false
+      //"test" -> "test" (undefined returned from fromTypedData)
       let valueFromDataType = fromTypedData(obj[key]);
-      // Only "undefined" will be replaced with origin object property.
       let value = valueFromDataType === undefined ? obj[key] : valueFromDataType;
       let camelCasedKey = key.charAt(0).toLocaleLowerCase() + key.slice(1);
       // If the value is a JSON object (and not array and not http, which is already cased), convert keys to camel case
--- a/src/converters/BindingConverters.ts
+++ b/src/converters/BindingConverters.ts
@@ -46,9 +46,9 @@ function isBindingDirection(input: string | undefined): boolean {
 export function convertKeysToCamelCase(obj: any) {
   var output = {};
   for (var key in obj) {
-      let value = fromTypedData(obj[key]);
+      let valueFromDataType = fromTypedData(obj[key]);
       // Only "undefined" will be replaced with origin object property.
-      value = value === undefined ? obj[key] : value;
+      let value = valueFromDataType === undefined ? obj[key] : valueFromDataType;
       let camelCasedKey = key.charAt(0).toLocaleLowerCase() + key.slice(1);
       // If the value is a JSON object (and not array and not http, which is already cased), convert keys to camel case
       if (!Array.isArray(value) && typeof value === 'object' && value && value.http == undefined) {
--- a/src/converters/BindingConverters.ts
+++ b/src/converters/BindingConverters.ts
@@ -46,7 +46,9 @@ function isBindingDirection(input: string | undefined): boolean {
 export function convertKeysToCamelCase(obj: any) {
   var output = {};
   for (var key in obj) {
-      let value = fromTypedData(obj[key]) || obj[key];
+      let value = fromTypedData(obj[key]);
+      // Only "undefined" will be replaced with origin object property.
+      value = value === undefined ? obj[key] : value;
       let camelCasedKey = key.charAt(0).toLocaleLowerCase() + key.slice(1);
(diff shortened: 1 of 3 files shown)


src/converters/BindingConverters.ts
```diff
@@ -46,7 +46,9 @@ function isBindingDirection(input: string | undefined): boolean {
 export function convertKeysToCamelCase(obj: any) {
   var output = {};
   for (var key in obj) {
-      let value = fromTypedData(obj[key]) || obj[key];
+      let valueFromDataType = fromTypedData(obj[key]);
+      // Only "undefined" will be replaced with origin object property.
```
davidmrdavid: by "origin", did you mean "original"? Just making sure :)

src/converters/BindingConverters.ts
```diff
@@ -46,7 +46,9 @@ function isBindingDirection(input: string | undefined): boolean {
...
+      let value = valueFromDataType === undefined ? obj[key] : valueFromDataType;
```
davidmrdavid, line 54: Can you elaborate, with an example, on why **only** undefined gets the original value?

test/BindingConvertersTests.ts
```diff
@@ -19,6 +19,21 @@ describe('Binding Converters', () => {
...
+           string : "false",
+        },
+        "falsyNull": {
+            string : "null"
+        },
+        "falsyEmptyString": {
+           string : "",
+        },
+        "falsyUndefined": {
+          string : undefined
```
davidmrdavid, line 36: Can we do these tests for the durable bindings as well?

davidmrdavid (commented):
Left some questions, thanks!

src/converters/BindingConverters.ts
alrod: fixed

src/converters/BindingConverters.ts
alrod: added examples

test/BindingConvertersTests.ts
alrod: Let's do it in a separate PR as it not related to the change. 

Do we want to release the change for both v3 and v4 or only v4?
https://github.com/Azure/azure-functions-nodejs-worker/commit/896e88c44a71a07572fa6196ec1a09245f3e9262
It looks like it was not released for v3 yet

src/converters/BindingConverters.ts
davidmrdavid: Thanks! One final question: does this need to handle the case where the input is `{ string : "{"foo": "bar"}" }`?

In other words, for expression bindings, don't we need to handle the case where the data is a JSON object? And, if so, do we need to recursively de-serialize it?

test/BindingConvertersTests.ts
davidmrdavid: Regarding adding tests in a new PR: That's fair, let's just make sure to create a tracking ticket for it then :) .

As for `896e88c`, is there a reason it hasn't been released yet for V3?

alrod: no, I think it safe to release it to v3. I just saw that it was not released for v3 yet.

davidmrdavid: In that case, let's release it to V3 as well :)

src/converters/BindingConverters.ts
alrod: No, we do not need this. there is a special type for "json", fromTypedData receives the interface:
```
    /** Properties of a TypedData. */
    interface ITypedData {

        /** TypedData string */
        string?: (string|null);

        /** TypedData json */
        json?: (string|null);

        /** TypedData bytes */
        bytes?: (Uint8Array|null);

        /** TypedData stream */
        stream?: (Uint8Array|null);

        /** TypedData http */
        http?: (AzureFunctionsRpcMessages.IRpcHttp|null);

        /** TypedData int */
        int?: (number|Long|null);

        /** TypedData double */
        double?: (number|null);

        /** TypedData collectionBytes */
        collectionBytes?: (AzureFunctionsRpcMessages.ICollectionBytes|null);

        /** TypedData collectionString */
        collectionString?: (AzureFunctionsRpcMessages.ICollectionString|null);

        /** TypedData collectionDouble */
        collectionDouble?: (AzureFunctionsRpcMessages.ICollectionDouble|null);

        /** TypedData collectionSint64 */
        collectionSint64?: (AzureFunctionsRpcMessages.ICollectionSInt64|null);
    }
}
```

davidmrdavid: In that interface above, it seems to me that `json` and `string` resolve to the same declaration, so I'm unsure of how this interface detects the JSON case:

```JavaScript
        /** TypedData string */
        string?: (string|null);

        /** TypedData json */
        json?: (string|null);
```

But if you're confident that the two cases don't overlap in practice, then I would be ready to approve this PR. Just let me know if this has been tested / we have confidence in this and, if so, I'll happily approve :) Thanks!

alrod: yes, it actually works in the same way for string and json.
I validated that for the such case ``({ string : '{"foo": "bar"}' })`` the logic returns the same result for old and new code so I am confident:
```
id: {
    foo: "bar"
}
```

davidmrdavid (approved):
LGTM!

PR #416, opened 2021-09-21
V4 only supports Node14

alrod:
1. we only want to support Node14 for V4: 
https://github.com/Azure/Azure-Functions/issues/1999

2. we want to remove V2 capability mode for V4:
https://docs.microsoft.com/en-us/azure/azure-functions/functions-app-settings#functions_v2_compatibility_mode


Francisco-Gamino (approved):
LGTM

Francisco-Gamino:
@alrod -- The E2E tests are failing. You probably also need to update this https://github.com/Azure/azure-functions-nodejs-worker/blob/v2.x/setup-e2e-tests.ps1#L24 to install the Core Tools for V4 for the test run.

pragnagopa (approved):

PR #417, opened 2021-09-24
Node16 support

### LINKED ISSUE #312: Upgrade to use grpc-js
https://grpc.io/blog/grpc-js-1.0/

Should I use @grpc/grpc-js or grpc?
> The original Node gRPC library (grpc) will no longer receive feature updates and we plan to deprecate it in a year, so we recommend that you use gRPC-JS, @grpc/grpc-js.
> However, some advanced features haven’t been ported to gRPC-JS yet, such as full compression support or support for other load balancing policies. If you need one of these features, you should use the grpc library, but open a feature request over gRPC-JS to let us know which features you are missing the most.

alrod:
Main challenge was  that "grpc" library we are using was deprecated as of April 2021 and last supported version is Node14. Authors of the library recommend to use pure javascript "grpc-js":
https://www.npmjs.com/package/grpc

Comparison "grpc" and "grpc-js"
https://github.com/grpc/grpc-node/blob/master/PACKAGE-COMPARISON.md

There can be a potential perf degradation as "grpc" is using some C++ native code and "grpc-js" is pure javascript but we really do not  have a choose here. Need to test the perf in prod.

Related to https://github.com/Azure/azure-functions-nodejs-worker/issues/455


pragnagopa (commented):
cc @Francisco-Gamino 

@alrod  - Will the E2E integration tests kick off for this branch as well?

alrod:
> cc @Francisco-Gamino
> 
> @alrod - Will the E2E integration tests kick off for this branch as well?

Chatted with @Francisco-Gamino offline,  v4.x E2E integration pipeline is coming next week.

pragnagopa (approved):

PR #422, opened 2021-10-04
Update VS Code config to prevent startup prompts

ejizba:
I fixed the tasks which are very out of date and suppressed a prompt from the Functions extension which is detecting the project in the `test/end-to-end/testFunctionApp` folder


alrod (approved):
can we also port this to v3.x

PR #424, opened 2021-10-04
Update information on branches

ejizba:
And remove references to dev/master because they are not used.

I think the default branch should always have a table like this at the top of the README since it's the home-page of the repo, but I figure non-default branches can just include the info for their version version instead of the whole table (see https://github.com/Azure/azure-functions-nodejs-worker/pull/425 for v1.x example).


jamiehaywood (approved):

alrod (approved):
can we also port this to v3.x

PR #434, opened 2021-10-07
Update package-lock to have reproducible builds

ejizba:
We should be using [`npm ci`](https://docs.npmjs.com/cli/v7/commands/npm-ci) in Azure Dev Ops otherwise it may install different dependencies than what's listed in package-lock.json, leading to inconsistent builds.

Also I've updated package-lock.json to be based off npm v7, which uses [lockfileVersion 2](https://github.blog/2021-02-02-npm-7-is-now-generally-available/#changes-to-the-lockfile). npm v7 has been out for a while now and I prefer to stay on later versions.

Finally, I tried removing the grpc-node workaround just to see if the issue had been fixed and the build is passing for me.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -62,10 +62,8 @@ jobs:
     inputs:
       versionSpec: $(NODE_VERSION)
     displayName: 'Install Node.js for test'
-  - script: npm install -g request
-    displayName: 'Workaround to https://github.com/grpc/grpc-node/issues/922'
-  - script: npm install
-    displayName: 'npm install'
+  - script: npm ci
+    displayName: 'npm ci'
   - script: npm run build
     displayName: 'npm run build'
   - script: npm test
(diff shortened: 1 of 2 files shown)


alrod:
do we need to update https://github.com/Azure/azure-functions-nodejs-worker#getting-started ?

alrod (changes requested):
do we need to update https://github.com/Azure/azure-functions-nodejs-worker#getting-started ?

ejizba:
Per the [docs](https://docs.npmjs.com/cli/v7/commands/npm-ci), here is when `npm ci` should be used:

> This command is similar to npm install, except it's meant to be used in automated environments such as test platforms, continuous integration, and deployment -- or any situation where you want to make sure you're doing a clean install of your dependencies.

Otherwise, `npm install` is still the default, including for local development. `npm ci` would prevent you from ever changing your dependencies, which is a pretty common thing to do when developing. If you _do_ change a dependency, it'll correspond with a change in the package.json and/or package-lock.json files which has to get committed for the automated build to pass (when using `npm ci`).

alrod (approved):

PR #437, opened 2021-10-14
Add basic linting to the repo

ejizba:
Formatting is done using `prettier` and linting is done with `eslint`. There were a lot of existing lint errors, so I disabled all rules that could not be fixed automatically and plan to fix those in smaller pieces in the future. I've also used `husky` to run the auto-fix when doing a commit - it adds a bit of a delay when committing which I'm not sure I like, but figured I'd give it a try.

I split this PR up into two commits:
1. Add the linting configuration: https://github.com/Azure/azure-functions-nodejs-worker/pull/437/commits/44b5fa76d94346e26dd6d1f11552bae72caae6e9
2. Everything that was auto-fixed: https://github.com/Azure/azure-functions-nodejs-worker/pull/437/commits/93d47e9e1123780df63e564c986927a6e8191a3e


alrod (changes requested):
Can you please also add a commit for removing Node8 or send a different PRs for all changes. It would be nice to have one PR for one change- reviewing is easier and history is cleaner.

alrod:
Do I understand right that most of the changes were done by the tools automatically?

ejizba:
> Can you please also add a commit for removing Node8 or send a different PRs for all changes. It would be nice to have one PR for one change- reviewing is easier and history is cleaner.

I removed Node 8 in azure-pipelines.yml in this PR because the build fails without it. Some of the lint stuff has a minimum node version of Node 10. I plan to 'rebase' instead of 'squash' on this PR so that the history stays clean, but I can't really split it into separate PRs otherwise the build will fail.

> Do I understand right that most of the changes were done by the tools automatically?

Yes. The first two commits I did manually and the last commit (which affects more files) was all auto-fixed. When I port this to v3.x, I will be able to cherry-pick the manual commits and re-run the tool for the auto-fixes.

alrod (approved):

ejizba:
I supposed I can do a separate PR to remove Node 8 - I'll just have to merge it before this PR. See https://github.com/Azure/azure-functions-nodejs-worker/pull/441

PR #438, opened 2021-10-15
Simplify TypeScript types package

ejizba:
The package felt overcomplicated in a few ways, so I refactored to make our lives simpler.
- Stop auto-generating `*.d.ts` files. The vast majority of auto-generated files were not used and its easier to directly maintain the `*.d.ts` file. Other people like [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped#create-a-new-package) and [VS Code](https://github.com/microsoft/vscode) do it this way, too
- Get rid of the generic `AzureFunction<>` type which forced us to have a separate ts3.1 definition. It has not been released yet and I don't see the value in doing this:
    ```typescript
     export const runTypedReturn: AzureFunction<CustomOutput> = async (context, request: HttpRequest) => {
    ```
    When you can just do this:
    ```typescript
    export const runTypedReturn: AzureFunction = async (context, request: HttpRequest): Promise<CustomOutput> => {
    ```

I also added unit tests to make sure our types work with both v3.0.x and v4.0.x of TypeScript.


alrod:
Having generic `AzureFunction<> ` is one of the reason for upgrading ts version:
https://github.com/Azure/azure-functions-nodejs-worker/pull/438

@anthonychu, do you remember why did we want generic `AzureFunction<> ` ? or we can can go without it?

ejizba:
The generic `AzureFunction<>` was first suggested here: https://github.com/Azure/azure-functions-nodejs-worker/pull/372, but IMO it did not have proper discussion prior to being adopted [here](https://github.com/Azure/azure-functions-nodejs-worker/pull/374#discussion_r573406165). For example, the "DefinitelyTyped" repo discusses in it's "Common mistakes" section how people often use generics more than they need to when authoring types: https://github.com/DefinitelyTyped/DefinitelyTyped#common-mistakes

ejizba:
@alrod @anthonychu anything else you need from me on this? It's blocking me from making any more progress on the types package

alrod (approved):

PR #441, opened 2021-10-19
Remove old references to Node 8

ejizba:
Node 8 is not supported on v2 of the worker.


alrod (approved):

ejizba:
Turns out Node 8 can still be used with `FUNCTIONS_V2_COMPATIBILITY_MODE`. Will follow up with some folks offline to figure out the best way forward on this

PR #444, opened 2021-10-21
Always run BuildArtifacts job in Dev Ops and fix errors in package step

ejizba:
The BuildArtifacts job started failing after https://github.com/Azure/azure-functions-nodejs-worker/pull/434 and I want to make sure we prevent failures from happening in the future by always running this job. The only thing that is still conditioned is "Push NuGet package to the AzureFunctionsPreRelease feed".

In order to fix the [errors](https://azfunc.visualstudio.com/Azure%20Functions/_build/results?buildId=38939&view=logs&j=02c445e2-ff43-58b5-026b-e17d54906f35&t=1128938b-5223-5ec9-11e1-ca0df3f8db03), I hard-coded the versions of some dependencies back to the old versions before the above-mentioned PR rather than trying to fix with the new versions. The v3.x branch is already using completely different dependencies anyways (grpc-js, etc.)

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -114,7 +114,6 @@ jobs:
       failTaskOnFailedTests: true
 
 - job: BuildArtifacts  
-  condition: and(succeeded(), eq(variables['Build.SourceBranch'], 'refs/heads/v2.x'), eq(variables['UPLOADPACKAGETOPRERELEASEFEED'], true))
   pool:
     vmImage: 'vs2017-win2016'
   steps:
@@ -122,6 +121,10 @@ jobs:
     inputs:
       versionSpec: $(NODE_10)
     displayName: 'Install Node.js for test'
+  - script: npm ci
+    displayName: 'npm ci'
+  - script: npm run build-nomaps
+    displayName: 'npm run build-nomaps'
   - powershell: .\package.ps1
   - task: NuGetCommand@2
     inputs:
@@ -133,7 +136,7 @@ jobs:
       ArtifactName: 'drop'
       publishLocation: 'Container'
   - task: NuGetCommand@2
-    condition: eq(variables['UPLOADPACKAGETOPRERELEASEFEED'], 'True')
+    condition: and(succeeded(), eq(variables['Build.SourceBranch'], 'refs/heads/v2.x'), eq(variables['UPLOADPACKAGETOPRERELEASEFEED'], true))
     inputs:
       command: 'push'
       packagesToPush: '$(Build.ArtifactStagingDirectory)/**/*.nupkg;!$(Build.ArtifactStagingDirectory)/**/*.symbols.nupkg'
--- a/scripts/generateProtos.js
+++ b/scripts/generateProtos.js
@@ -1,6 +1,6 @@
-const util = require("util");
-const exec = util.promisify(require("child_process").exec);
-const path = require("path");
+const util = require('util');
+const exec = util.promisify(require('child_process').exec);
+const path = require('path');
 
 async function generateProtos() {
     try {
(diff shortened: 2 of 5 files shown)


scripts/generateProtos.js
```diff
@@ -10,31 +10,38 @@ async function generateProtos() {
         const protoFiles = [
             path.join(protoRoot, 'shared', 'NullableTypes.proto'),
             path.join(protoRoot, 'identity', 'ClaimsIdentityRpc.proto'),
-            path.join(protoRoot, 'FunctionRpc.proto')
+            path.join(protoRoot, 'FunctionRpc.proto'),
```
alrod, line 13: remove comma

azure-pipelines.yml
```diff
@@ -114,14 +114,17 @@ jobs:
...
 - job: BuildArtifacts  
-  condition: and(succeeded(), eq(variables['Build.SourceBranch'], 'refs/heads/v2.x'), eq(variables['UPLOADPACKAGETOPRERELEASEFEED'], true))
   pool:
     vmImage: 'vs2017-win2016'
   steps:
   - task: NodeTool@0
     inputs:
       versionSpec: $(NODE_10)
     displayName: 'Install Node.js for test'
+  - script: npm ci
```
alrod, line 124: why did you move the  calls from `package.ps1`?

azure-pipelines.yml
```diff
@@ -133,7 +136,7 @@ jobs:
       ArtifactName: 'drop'
       publishLocation: 'Container'
   - task: NuGetCommand@2
-    condition: eq(variables['UPLOADPACKAGETOPRERELEASEFEED'], 'True')
+    condition: and(succeeded(), eq(variables['Build.SourceBranch'], 'refs/heads/v2.x'), eq(variables['UPLOADPACKAGETOPRERELEASEFEED'], true))
```
alrod, line 139: can we remove eq(variables['Build.SourceBranch'], 'refs/heads/v2.x') - sometime we need to build the artifacts based on a PR

alrod (changes requested):

scripts/generateProtos.js
ejizba: This is by design. The auto-formatter adds trailing commas wherever possible (which is almost always), which makes it slightly easier to reorder lines, copy/paste lines, add new lines, etc.

https://prettier.io/docs/en/options.html#trailing-commas

azure-pipelines.yml
ejizba: Because package.ps1 should be using `npm ci` and the only reason I didn't update it in https://github.com/Azure/azure-functions-nodejs-worker/pull/434 was because I had no idea package.ps1 was running `npm install` in the first place. I prefer if the build/test job is as consistent with the package job as possible so that we know we're shipping the same thing we're testing.

azure-pipelines.yml
ejizba: With my change, the artifact is now available to be downloaded from the build. I haven't tried it, but I assume that can be installed locally for testing purposes. For example, see [here](https://azfunc.visualstudio.com/Azure%20Functions/_build/results?buildId=39013&view=artifacts&pathAsName=false&type=publishedArtifacts)

The only thing that checks against the v2.x branch is for pushing that artifact to the prerelease feed. Are you saying you also want the nupkg pushed to the feed from a PR? If yes, I'd be curious what exactly uses the prerelease feed (integration tests, etc.?) before removing

alrod: oh sorry, I thought the condition was for the building artifacts

alrod (approved):

PR #457, opened 2021-11-03
Fix Cosmos e2e test

ejizba:
A Cosmos DB error code changed over the weekend and our e2e test fails because it's not retrying `ReadDocument` when it should be.

cc @Francisco-Gamino


Francisco-Gamino (approved):
LGTM

PR #409, opened 2021-08-21
Add type definitions for Timer 

jamiehaywood:
relates #398 

I wasn't entirely sure about the definitions, but I had a look at the source code, which does have comments, and lifted and shifted from there:
https://github.com/Azure/azure-webjobs-sdk-extensions/blob/9c87af74f90159f8b987c77aa77546f55a62a84e/src/WebJobs.Extensions/Extensions/Timers/Scheduling/ScheduleStatus.cs

PR #463, opened 2021-11-03
Add types test for Timer trigger

ejizba:
Nothing fancy, just adding the default timer template referencing the new `Timer` type added [here](https://github.com/Azure/azure-functions-nodejs-worker/pull/409)


hossam-nasr (approved):
Lgtm!

PR #464, opened 2021-11-09
Update version info on v2.x README

ejizba:
Now that v4 of the runtime GA-ed, I will make v3.x the default branch for the worker. That means the v2.x branch only has to have info on its particular version stuff since IMO it's only necessary to keep the full table on the default branch.


hossam-nasr (approved):
Looks good!

PR #476, opened 2021-11-22
Update README now that Node 16 is officially in public preview

ejizba:
https://azure.microsoft.com/updates/public-preview-nodejs-16-in-azure-functions/


hossam-nasr (approved):
Lgtm!

Issue #428, opened 2021-10-05
'@azure/functions' npm package versioning


ejizba:
As of right now, the only version of the `@azure/functions` [npm package](https://www.npmjs.com/package/@azure/functions) that has been published is v1.x. However, each new version of the Node.js worker can introduce breaking changes that would affect the TypeScript types. My proposal is that the TypeScript types should be directly tied to the worker version and we should release/maintain each major version that is still supported by the runtime.

|@azure/functions version|Worker Version|Runtime Version|
|---|---|---|
|1.x|1.x|2.x|
|2.x|2.x|3.x|
|3.x|3.x|4.x|

Some notes:
- With v4 of the runtime, we considered 'aligning' the worker and runtime versions, but long term we want to be able to release the worker separately from the runtime, so the versions would likely diverge eventually anyways. Plus skipping from 2 to 4 could just add confusion.
- Most users are on v3 of the host by now, but are still using v1 of the typescript types because that's all that we've released. We should refrain from updating v1 types in such a way that may break those users.

ejizba:
Both v2 and v3 of the npm package have been released. Since we are now caught up with the worker versions, I will go ahead and close this issue.

PR #478, opened 2021-11-22
Bump types package to 3.0.0

ejizba:
I don't see the need to do a beta release because it's so similar to v2.

Here's the general issue on versioning: https://github.com/Azure/azure-functions-nodejs-worker/issues/428


hossam-nasr (approved):
Lgtm!

Issue #421, opened 2021-10-01
Automatically decode x-ms-client-principal header


anthonychu:
Add a `user` property to the `HttpRequest` that contains an object representing the logged in user.

If a user is logged in, the platform sets the `x-ms-client-principal` header, which is in base64 encoded JSON. Both AppService/Functions authentication and Static Web Apps authentication support this header. However, they each have a slightly different schema.

Option 1: automatically decode the header into `req.user` and leave it to the developer to figure out which type they are working with.

Option 2: automatically decode the header into `req.user` in a common type. The worker takes care of mapping fields from each schema into the corresponding fields in the common type. The developer can expect the same type no matter which auth the app is using.

anthonychu:
Option 3, which I think is the one I'd suggest:

- Add a `user` property to the http request object if `x-ms-client-principal` exists:
    ```js
    {
      "type": "AppService", // or "StaticWebApps"
      "id": "b18031c7-46de-49c5-b68a-b562af6876a7",
      "username": "connie.wilson@contoso.com",
      "identityProvider": "aad",
      "claimsPrincipalData": { ... }
    }
- `claimsPrincipalData` - base64 decoded and JSON parsed `x-ms-client-principal`
- `type` - determine based on the properties in `claimsPrincipalData`

| Property | AppService | StaticWebApps |
| --- | --- | --- |
| `id` | `x-ms-client-principal-id` header | `userId` in client principal |
| `username` | `x-ms-client-principal-name` header | `userDetails` in client principal |
| `identityProvider` | `x-ms-client-principal-idp` header | `identityProvider` in client principal |

hossam-nasr:
Closing this issue, implemented in this PR #483.

Basically following @anthonychu's option 3. In `HttpRequest`, the `user` property is added:

```TypeScript
/**
  *  Object representing logged-in user, either through
  *  AppService/Functions authentication, or SWA Authentication
  *  null when no such user is logged in.
  */
user: HttpRequestUser | null;
```

which has the following interface:

```TypeScript
/**
 *  Object representing logged-in user, either through
 *  AppService/Functions authentication, or SWA Authentication
 */
export interface HttpRequestUser {
    /**
     * Type of authentication, either AppService or StaticWebApps
     */
    type: HttpRequestUserType;
    /**
     * unique user GUID
     */
    id: string;
    /**
     * unique username
     */
    username: string;
    /**
     * provider of authentication service
     */
    identityProvider: string;
    /**
[...]

PR #483, opened 2021-12-01
Automatically decode x-ms-client-principal header

hossam-nasr:
Automatically decode `x-ms-client-principal` header #421:

- Add `HttpRequestUser` interface and `HttpRequestUserType` type, according to the definitions in #421 
- Add `user` property in `HttpRequest` interface
- Based on data in` x-ms-client-principal` header, figure out the type of authentication used, and add appropriate user field in the `HttpRequest` constructor


src/http/Request.ts
```diff
@@ -19,6 +20,29 @@ export class Request extends RequestProperties {
     constructor(httpInput: RequestProperties) {
         super();
         Object.assign(this, httpInput);
+        if (httpInput.headers['x-ms-client-principal']) {
```
ejizba: If there's any issues parsing the client principal, I don't think that should throw an error here (basically preventing the entire execution from happening). Instead, how about we lazy-parse this by putting it in a getter - that way only people that try to access the user property would see the error.

Only downside to a getter is by default it would re-do the parsing each time. However, we can get around that pretty easily by just caching the result of our parse. Something like this:
```typescript
private _cachedUser?: HttpRequestUser;
public get user(): HttpRequestUser | null {
    if (this._cachedUser === undefined) {
        this._cachedUser = // do all your stuff to parse user
    }

    return this._cachedUser;
}
```

src/http/Request.ts
```diff
@@ -19,6 +20,29 @@ export class Request extends RequestProperties {
     constructor(httpInput: RequestProperties) {
         super();
         Object.assign(this, httpInput);
+        if (httpInput.headers['x-ms-client-principal']) {
+            const claimsPrincipalData = JSON.parse(
+                Buffer.from(httpInput.headers['x-ms-client-principal'], 'base64').toString('utf-8')
+            );
+
+            if (claimsPrincipalData['identityProvider']) {
+                this.user = {
```
ejizba: Please add a few unit tests. Add at least one real-life test for both "StaticWebApps" and "AppService" (making sure to obfuscate any important secrets/keys) and then add any edge-case tests as you see fit.

types/index.d.ts
```diff
@@ -131,6 +160,11 @@ declare module '@azure/functions' {
          * Route parameter keys and values.
          */
         params: HttpRequestParams;
+        /**
+         *  Object representing logged-in user, either through
+         *  AppService/Functions authentication, or SWA Authentication
+         */
+        user?: HttpRequestUser | null;
```
ejizba: We have to be careful about `undefined` vs `null`, especially because we do not have linting properly set up to cover this area yet. The differences are subtle, but definitely there. In this case, I would get rid of the question mark.
```suggestion
        user: HttpRequestUser | null;
```

`undefined` means we have not figured out what the value of user is yet, whereas `null` means we have figured it out and there is no user. We should never return `undefined` here and the types should reflect that

src/http/ExtractHttpUser.ts
```diff
@@ -0,0 +1,34 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License.
+
+import { HttpRequestHeaders, HttpRequestUser } from '@azure/functions';
+
+export function extractHttpUserFromHeaders(headers: HttpRequestHeaders): HttpRequestUser | null {
```
ejizba: As a general rule (which is not being followed in this repo yet), the file name should match the main exported function name. Aka rename it to `extractHttpUserFromHeaders.ts`

test/ExtractHttpUserTests.ts
```diff
@@ -0,0 +1,101 @@
...
+        };
+
+        const headers: HttpRequestHeaders = {
+            'x-ms-client-principal': Buffer.from(JSON.stringify(claimsPrinicipalData)).toString('base64'),
+        };
+
+        const user: HttpRequestUser | null = extractHttpUserFromHeaders(headers);
+
+        expect(user).to.not.be.null;
+        expect(user?.id).to.equal(id);
```
ejizba, line 71: You should also verify `user.type`.

In general, validate at least every property you added in the `types/index.d.ts` file

test/ExtractHttpUserTests.ts
```diff
@@ -0,0 +1,101 @@
...
+            'x-ms-client-principal': Buffer.from(JSON.stringify(claimsPrincipalData)).toString('base64'),
+        };
+
+        const user: HttpRequestUser | null = extractHttpUserFromHeaders(headers);
+
+        expect(user).to.not.be.null;
+        expect(user?.id).to.equal(id);
+        expect(user?.username).to.equal(username);
+        expect(user?.identityProvider).to.equal(provider);
+        expect(user?.claimsPrincipalData).to.eql(claimsPrincipalData);
```
ejizba: nit: Not a big fan of `eql`. I think `.to.deep.equal` does the same thing and is more obvious

test/ExtractHttpUserTests.ts
```diff
@@ -0,0 +1,101 @@
...
+describe('Extract Http User Claims Principal from Headers', () => {
+    it('Correctly parses AppService headers', () => {
+        const username = 'test@example.com';
+        const id: string = uuid();
+        const provider = 'aad';
+        const claimsPrincipalData = {
+            auth_typ: provider,
+            claims: [
+                { typ: 'aud', val: uuid() },
+                { typ: 'iss', val: `https://login.microsoftonline.com/${uuid()}/v2.0` },
```
ejizba: For the sake of readability, you can delete some of this junk in the "claims" array. I know I said make the unit test as real as possible, but often the real data is way too verbose and we can get away with trimming it down a bit. Just keep the general structure as-is and obviously keep any important fields

types/index.d.ts
```diff
@@ -131,6 +160,11 @@ declare module '@azure/functions' {
          * Route parameter keys and values.
          */
         params: HttpRequestParams;
+        /**
+         *  Object representing logged-in user, either through
```
ejizba, line 164: You should ideally describe when `null` is returned. Something like https://github.com/Azure/azure-functions-nodejs-worker/pull/486/files#diff-093ad82a25aee498b11febf1cdcb6546e4d223ffcb49ed69cc275ac27ce0ccceR152

ejizba (commented):
Wrote most of these comments a while ago, not sure why I left them in a "pending" state. Also please update with the latest v3.x

src/http/ExtractHttpUser.ts
hossam-nasr: Renamed the file

test/ExtractHttpUserTests.ts
hossam-nasr: Added checking `user.type` for both tests

test/ExtractHttpUserTests.ts
hossam-nasr: Done

test/ExtractHttpUserTests.ts
hossam-nasr: Thanks! Done

types/index.d.ts
hossam-nasr: Done

ejizba (approved):

PR #484, opened 2021-12-02
Add launch config for unit tests

ejizba:
And add reference to types for chai (used in unit tests)

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,25 +1,23 @@
 {
-    // Use IntelliSense to learn about possible Node.js debug attributes.
-    // Hover to view descriptions of existing attributes.
-    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
     "version": "0.2.0",
     "configurations": [
-
-        {
-            "type": "node",
-            "request": "launch",
-            "name": "Launch Program",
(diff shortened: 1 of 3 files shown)


.vscode/launch.json
```diff
@@ -1,25 +1,23 @@
...
-        {
-            "type": "node",
-            "request": "launch",
-            "name": "Launch Program",
-            "program": "${file}",
-            "outFiles": [
-                "${workspaceRoot}/dist/src/*.js"
-            ],
-            "args": ["--host", "127.0.0.1", "--port", "50051", "--requestId", "100a", "--workerId", "worker1"]
-        },
```
hossam-nasr, line 17: Why are we removing this?

ejizba: IMO this one is pretty useless since it doesn't use a real function app and just fakes the args. I always test with the "Attach by Process ID" one as described in the README. If you find this one useful, though, we can leave it in

hossam-nasr: No, I also just debug using the "Attach by Process ID," but I wasn't sure if maybe this was used for other purposes and removed by mistake. Thanks!

hossam-nasr (approved):

PR #485, opened 2021-12-02
Combine Request and RequestProperties

ejizba:
See this code for an http trigger:
```typescript
const httpTrigger: AzureFunction = async function (context: Context, req: HttpRequest): Promise<void> {
    context.log('HTTP trigger function processed a request.');

    console.log(context.req === req);
```

Currently, `false` gets printed. That makes it difficult for us to add things to Request (like the form and client-principal work) because we'd have to add it in both places. My fix combines the two and ensures people will get the same experience no matter which `req` they use.

### DIFF
--- a/src/http/Request.ts
+++ b/src/http/Request.ts
@@ -1,27 +1,33 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License.
 
-import { HttpMethod, HttpRequest } from '@azure/functions';
+import { HttpMethod, HttpRequest, HttpRequestHeaders, HttpRequestParams, HttpRequestQuery } from '@azure/functions';
+import { AzureFunctionsRpcMessages as rpc } from '../../azure-functions-language-worker-protobuf/src/rpc';
+import { fromTypedData } from '../converters/RpcConverters';
+import { fromNullableMapping, fromRpcHttpBody } from '../converters/RpcHttpConverters';
 
-export class RequestProperties implements HttpRequest {
-    method: HttpMethod | null = null;
-    url = '';
-    originalUrl = '';
-    headers: { [key: string]: string } = {};
-    query: { [key: string]: string } = {};
-    params: { [key: string]: string } = {};
-    body?: any;
-    rawBody?: any;
-    [key: string]: any;
-}
+export class Request implements HttpRequest {
+    public method: HttpMethod | null;
+    public url: string;
+    public originalUrl: string;
+    public headers: HttpRequestHeaders;
+    public query: HttpRequestQuery;
+    public params: HttpRequestParams;
+    public body?: any;
+    public rawBody?: any;
 
-export class Request extends RequestProperties {
-    constructor(httpInput: RequestProperties) {
-        super();
-        Object.assign(this, httpInput);
+    public constructor(rpcHttp: rpc.IRpcHttp) {
+        this.method = <HttpMethod>rpcHttp.method;
+        this.url = <string>rpcHttp.url;
+        this.originalUrl = <string>rpcHttp.url;
+        this.headers = fromNullableMapping(rpcHttp.nullableHeaders, rpcHttp.headers);
+        this.query = fromNullableMapping(rpcHttp.nullableQuery, rpcHttp.query);
+        this.params = fromNullableMapping(rpcHttp.nullableParams, rpcHttp.params);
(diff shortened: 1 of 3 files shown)


src/http/Request.ts
```diff
@@ -1,27 +1,33 @@
...
+        this.originalUrl = <string>rpcHttp.url;
+        this.headers = fromNullableMapping(rpcHttp.nullableHeaders, rpcHttp.headers);
+        this.query = fromNullableMapping(rpcHttp.nullableQuery, rpcHttp.query);
+        this.params = fromNullableMapping(rpcHttp.nullableParams, rpcHttp.params);
+        this.body = fromTypedData(<rpc.ITypedData>rpcHttp.body);
+        this.rawBody = fromRpcHttpBody(<rpc.ITypedData>rpcHttp.body);
     }
 
-    get(field: string): string | undefined {
+    public get(field: string): string | undefined {
```
hossam-nasr, line 30: Just out of curiosity, what's the difference between specifying `public` and not having it at all? I see you added `public` to everything

src/http/Request.ts
```diff
@@ -1,27 +1,33 @@
...
+    public rawBody?: any;
 
-export class Request extends RequestProperties {
-    constructor(httpInput: RequestProperties) {
-        super();
-        Object.assign(this, httpInput);
+    public constructor(rpcHttp: rpc.IRpcHttp) {
+        this.method = <HttpMethod>rpcHttp.method;
+        this.url = <string>rpcHttp.url;
+        this.originalUrl = <string>rpcHttp.url;
```
hossam-nasr, line 22: I know this was also in the original version, but is there a difference between `this.url` and `this.originalUrl`?

src/http/Request.ts
```diff
@@ -1,27 +1,33 @@
...
-export class RequestProperties implements HttpRequest {
-    method: HttpMethod | null = null;
-    url = '';
-    originalUrl = '';
-    headers: { [key: string]: string } = {};
-    query: { [key: string]: string } = {};
-    params: { [key: string]: string } = {};
-    body?: any;
-    rawBody?: any;
-    [key: string]: any;
```
hossam-nasr, line 15: Is there a reason this part is left out now?

hossam-nasr (commented):
Thanks for this! I wanted to do something similar for the `x-ms-client-principal` ticket too, but seemed a bit outside the scope of that issue. But it didn't make a lot of sense to me that `context.req` and `req` wouldn't have the same context. This looks good to me, but I left just a few comments out of curiosity/for my learning. Thanks!

src/http/Request.ts
ejizba: There's no difference. It's a stylistic thing that I'm used to doing. Eventually we should decide what we prefer and get the linter to automatically handle this for us. Until then, I think it's up to whatever the person wants

src/http/Request.ts
ejizba: Nope. There's quite a bit of duplicate stuff, but we can't clean it up because it might break people

src/http/Request.ts
ejizba: All it's saying is that the type for any other properties not explicitly mentioned is `any`, but `any` could literally be anything so there's no point. It's actually pretty discouraged to ever use `any` in TypeScript (it kind of defeats the point of _Type_Script), but unfortunately this repo uses it in a lot of places. I plan to slowly clean it up, but it'll take a while

hossam-nasr: It does also specify that the key type must be `string` though, right? Or is that always the case anyway?

hossam-nasr (approved):

src/http/Request.ts
ejizba: Here are a few examples showing the nuances

With `[key: string]: any;`, you basically never get errors. TypeScript lets you do anything:
Screen Shot 2021-12-02 at 11 16 52 AM

If I did `[key: string]: string;` instead, TypeScript assumes `value` is `string` regardless of the key type:
Screen Shot 2021-12-02 at 11 17 34 AM

Lastly, if you get rid of `[key: string]: ...` TypeScript only enforces types if you access a property with a `.` (which is the most common way to access properties). This is the behavior I prefer. We should get an error unless we define the type for the `test` property.
Screen Shot 2021-12-02 at 11 17 14 AM

hossam-nasr: That makes sense, thanks!

PR #490, opened 2021-12-04
Generate SBOM for both worker and types package

ejizba:
Previously the only artifact in this pipeline was the worker nuget package. I've updated it so that we also generate the types package as a *.tgz file in the same pipeline. I will update the release pipeline for the types package to use this tar file instead of generating it's own tar file.

Finally, I modified the default SBOM generation to be more accurate for our use case. Previously it would include all dependencies (prod & dev) in the manifest, but it should only include prod.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -131,12 +131,21 @@ jobs:
   - task: NuGetCommand@2
     inputs:
       command: pack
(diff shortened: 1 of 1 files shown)


davidmrdavid (approved):
LGTM, good idea on the --prod prune step

PR #492, opened 2021-12-08
Add release pipeline yaml for types package

ejizba:
Related to SBOM work (https://github.com/Azure/azure-functions-nodejs-worker/pull/490), we need to actually use the types package artifact from that build for releasing. Previously it was a non-yaml based release pipeline and it basically created its own artifact.

Example pipeline run: https://azfunc.visualstudio.com/Azure%20Functions/_build/results?buildId=44253&view=results

(I've only tried the publish with `--dry-run` so far)

### DIFF
--- a/azure-pipelines-release.yml
+++ b/azure-pipelines-release.yml
@@ -27,5 +27,5 @@ jobs:
       command: custom
       workingDir: '$(Pipeline.Workspace)/nodeWorkerCI/types'
       verbose: true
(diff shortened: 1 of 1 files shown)


davidmrdavid (approved):
LGTM, I even tested the rough logic here by porting it to DurableJS and DurablePython and it seems to work there as well. Thank you for the guiding example, it's saved me quite a bit of time :)

PR #494, opened 2021-12-14
Move build/release to 1ES hosted agents

ejizba:
We need to use 1ES hosted agents for anything build/release related, but can continue to use Microsoft Hosted agents for test purposes (which is nice because 1ES doesn't support mac). That being said, we have to make sure the test jobs can't affect the build/release, so I thought it would be safest to split the build/test pipelines into completely different definitions.

The bulk of this PR is just splitting/moving files. The only actual definition changes are listed here:
- Change "pool" in build.yml and types-release.yml to 1ES pool
- Add optional `NPM_PUBLISH_ARGS` variable to types-release so I can more easily do `--dry-run`


Francisco-Gamino (approved):
LGTM

PR #498, opened 2021-12-15
Include sbom manifest in packages

ejizba:
Basic strategy for each package:
- Copy the files to a staging directory (this ensures the manifest won't have extra files that aren't actually shipped)
- Generate manifest in staging directory
- Create package from staging directory

I got rid of the package.ps1 script because its easier to get the ordering right if you do each item (copy files, run webpack, etc.) as its own azure pipelines task.

Here's a sample build drop. If you unzip the packages the manifests are inside and are very accurate: https://azfunc.visualstudio.com/Azure%20Functions/_build/results?buildId=45434&view=artifacts&pathAsName=false&type=publishedArtifacts


azure-pipelines/build.yml
```diff
@@ -23,29 +23,40 @@ jobs:
...
   - script: npm ci
     displayName: 'npm ci'
-  - script: npm run build-nomaps
-    displayName: 'npm run build-nomaps'
-  - powershell: .\package.ps1
-  - task: NuGetCommand@2
+  - script: npm run build
+    displayName: 'npm run build'
+  - script: npm run webpack
+    displayName: 'npm run webpack'
```
davidmrdavid, line 29: Why did we drop the `-nomaps` option, and was `webpack` always part of this pipeline?

azure-pipelines/build.yml
```diff
@@ -23,29 +23,40 @@ jobs:
...
+      targetFolder: '$(Build.ArtifactStagingDirectory)/worker'
+      cleanTargetFolder: true
   - script: npm prune --production
     displayName: 'npm prune --production' # so that only production dependencies are included in SBOM
   - task: ManifestGeneratorTask@0
     displayName: 'Generate SBOM for worker'
     inputs:
       BuildDropPath: '$(Build.ArtifactStagingDirectory)/worker'
-      BuildComponentPath: '$(System.DefaultWorkingDirectory)/node_modules'
+      BuildComponentPath: '$(Build.SourcesDirectory)/node_modules'
```
davidmrdavid, line 50: Why is this working of the `SourcesDirectory`? I thought we had just copied the files to "staging"?

azure-pipelines/build.yml
```diff
@@ -23,29 +23,40 @@ jobs:
...
       PackageName: 'Azure Functions Node.js Worker'
-  - task: PublishBuildArtifacts@1
-    displayName: 'Publish worker artifact'
+  - task: NuGetCommand@2
+    displayName: 'NuGet pack worker'
     inputs:
-      PathtoPublish: '$(Build.ArtifactStagingDirectory)/worker'
-      ArtifactName: 'worker'
+      command: pack
+      packagesToPack: '$(Build.SourcesDirectory)/Worker.nuspec'
```
davidmrdavid, line 56: Also wondering why we're not packing from the "staging" directory

davidmrdavid (commented):
I have some questions prior to an approval :)

azure-pipelines/build.yml
ejizba: > Why did we drop the -nomaps option

I assume someone added "-nomaps" to prevent cluttering up the package with "map" files, but they're never included anyway since we explicitly copy each file we want to a staging directory before packaging.

> was webpack always part of this pipeline?

Yes, it was in the `package.ps1` script before (which felt pretty hidden to me)

azure-pipelines/build.yml
ejizba: Just to be clear, `System.DefaultWorkingDirectory` and `Build.SourcesDirectory` are the same, I just prefer the shorter name

To answer your question: Once we webpack the code, anything we used from the "node_modules" folder is now in that webpacked file (worker-bundle.js) and is pretty mangled up. The sbom manifest tool can't determine the components from the webpacked file in the staging dir, so I'm referencing the original un-webpacked modules in the sources directory

azure-pipelines/build.yml
ejizba: The basePath is the staging directory, which is where all the files are.

I could've copied the nuspec to the staging directory and then I wouldn't need to specify "packagesToPack" _and_ "basePath", but then the nuspec is listed in the manifest file which didn't seem 100% accurate considering when you unpack the nuget package you get a folder like this:
Screen Shot 2021-12-16 at 11 55 06 AM

It may or may not matter, but the nuspec is not in that folder so I felt like it shouldn't be in the manifest.

azure-pipelines/build.yml
davidmrdavid: Makes sense. Do you mind documenting this in the YAML itself? It's not immediately obvious, at least to me.

davidmrdavid:
Thanks for responding to my questions. @ejizba, I'd be happy to approve if we document that the "The sbom manifest tool can't determine the components from the webpacked file" and that this is why the SBOM is referencing the _original_ source code :-)

davidmrdavid (approved):
LGTM!

PR #501, opened 2021-12-16
Adding package.json warning when using AzureFiles

alrod:
On [reading function code](https://github.com/Azure/azure-functions-nodejs-worker/blob/v3.x/src/FunctionLoader.ts#L50) NodeJs tries to find 'package.json' all the way up to the file system root
 In Azure files it causes a delay during cold start as connection to Azure Files is an expensive operation.

We want to add a log to detect the issue.

### DIFF
--- a/src/WorkerChannel.ts
+++ b/src/WorkerChannel.ts
@@ -2,6 +2,7 @@
 // Licensed under the MIT License.
 
 import { Context } from '@azure/functions';
+import { access, constants } from 'fs';
 import { format, isFunction } from 'util';
 import { AzureFunctionsRpcMessages as rpc } from '../azure-functions-language-worker-protobuf/src/rpc';
 import { CreateContextAndInputs, LogCallback, ResultCallback } from './Context';
@@ -166,6 +167,8 @@ export class WorkerChannel implements IWorkerChannel {
                 });
                 err = exception;
             }
+
+            this.logColdStartWarning();
 
             this._eventStream.write({
                 requestId: requestId,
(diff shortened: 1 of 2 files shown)


src/WorkerChannel.ts
```diff
@@ -407,4 +410,32 @@ export class WorkerChannel implements IWorkerChannel {
...
+            process.env.WEBSITE_CONTENTAZUREFILECONNECTIONSTRING &&
+            process.env.WEBSITE_CONTENTSHARE &&
+            process.env.AzureWebJobsScriptRoot
+        ) {
+            // Add delay to avoid affecting coldstart
+            if (!delayInMs) {
+                delayInMs = 5000;
+            }
+            setTimeout(() => {
+                const path = require('path');
```
ejizba: nit: I would put this "path" import at the top of the file with all the other imports

src/WorkerChannel.ts
```diff
@@ -407,4 +410,32 @@ export class WorkerChannel implements IWorkerChannel {
...
+                access(path.join(process.env.AzureWebJobsScriptRoot, 'package.json'), constants.F_OK, (e) => {
+                    if (e) {
+                        this.log({
+                            message:
+                                'package.json is not found at the root of the Function App in Azure Files - cold start for NodeJs can be affected.',
+                            level: LogLevel.Debug,
```
ejizba, line 433: Should this be `LogLevel.Warning`?

src/WorkerChannel.ts
```diff
@@ -167,6 +168,8 @@ export class WorkerChannel implements IWorkerChannel {
                 err = exception;
             }
 
+            this.logColdStartWarning();
```
ejizba: I believe by putting this in `functionLoadRequest`, the log will get written once per function. However, we really only need to log it once per app. With that in mind, I think `workerInitRequest` might be a better place

src/WorkerChannel.ts
alrod: Let's leave it as LogLevel. Debug as we do not want to scary customers with sudden unknown warning.

src/WorkerChannel.ts
alrod: yeah, you are right. moved to `workerInitRequest`

ejizba (approved):

PR #506, opened 2021-12-17
Port "Include sbom manifest in packages" to v2.x

ejizba:
Original PR: https://github.com/Azure/azure-functions-nodejs-worker/pull/498

The two differences:
- This branch is using the old grpc libraries and installs them during the `package.ps1` script (which I deleted on v3.x). I still deleted the package.ps1 script, but I moved all the grpc stuff into a new script `scripts/installGrpc.ps1`
- This branch is on an older version of webpack, so the args aren't the same in the package.json script


package.json
```diff
@@ -50,12 +50,12 @@
...
         "clean": "rimraf dist && rimraf azure-functions-language-worker-protobuf/src/rpc*",
         "build": "rimraf dist && npm run gen && shx mkdir -p dist/azure-functions-language-worker-protobuf/src && shx cp azure-functions-language-worker-protobuf/src/rpc.* dist/azure-functions-language-worker-protobuf/src/. && node ./node_modules/typescript/bin/tsc",
-        "build-nomaps": "npm run build -- --sourceMap false",
         "gen": "node scripts/generateProtos.js",
         "test": "mocha ./dist/test/**/*.js --reporter mocha-multi-reporters --reporter-options configFile=test/mochaReporterOptions.json",
         "lint": "eslint .",
         "lint-fix": "eslint . --fix",
-        "watch": "node ./node_modules/typescript/bin/tsc --watch"
+        "watch": "node ./node_modules/typescript/bin/tsc --watch",
+        "webpack": "webpack"
```
davidmrdavid, line 58: I suppose this is what you mean by "This branch is on an older version of webpack, so the args aren't the same in the package.json script" as it's missing the `--mode production` flags.

I'm assuming that `--mode production` used to be the default value in this older version, right? What confuses me is that the documentation [here](https://webpack.js.org/configuration/mode/) says that `production` is still the default, so I guess I'm just unsure as to why we needed that flag in the `v3.x` PR.

Regardless, this seems like a non-blocking concern, but I'd appreciate a response :-)

davidmrdavid (approved):
LGTM but have 1 non-blocking question in my comments

package.json
ejizba: Yeah `production` is still the default, but if you run it without the `--mode` argument, webpack will print these annoying warnings:

> WARNING in configuration
> The 'mode' option has not been set, webpack will fallback to 'production' for this value.
> Set 'mode' option to 'development' or 'production' to enable defaults for each environment.
> You can also set it to 'none' to disable any default behavior. Learn more: https://webpack.js.org/configuration/mode/

davidmrdavid: Fair enough, good to merge then

PR #511, opened 2022-01-12
Allow Node.js v17.x and v15.x to be used during local development with warning messages

### LINKED ISSUE #363: Improve local dev experience for unsupported versions of Node.js
It's a better local development experience if warnings are issued on local development rather than outright failure for missing Node.js versions

hossam-nasr:
Fixes #363 . Allows Node.js `v17.x` and `v15.x` to be used if function is being run locally (if env variable `AZURE_FUNCTIONS_ENVIRONMENT == "Development"`). Displays a warning to the customer that this version is not officially supported and that the function will be run using a different Node.js version when deployed to Azure.

### DIFF
--- a/src/WorkerChannel.ts
+++ b/src/WorkerChannel.ts
@@ -119,11 +119,14 @@ export class WorkerChannel implements IWorkerChannel {
     public workerInitRequest(requestId: string, msg: rpc.WorkerInitRequest) {
         // Validate version
         const version = process.version;
-        if (version.startsWith('v17.') && process.env.AZURE_FUNCTIONS_ENVIRONMENT == 'Development') {
+        if (
+            (version.startsWith('v17.') || version.startsWith('v15.')) &&
+            process.env.AZURE_FUNCTIONS_ENVIRONMENT == 'Development'
+        ) {
             const msg =
                 'Node.js version used (' +
                 version +
-                ') is not officially supported. You may use it during local development, but your function will be deployed to a different Node.js version on Azure.' +
+                ') is not officially supported. You may use it during local development, but must use an officially supported version on Azure.' +
                 ' The version of the Azure Functions runtime you are using (v4) officially supports Node.js v14.x or Node.js v16.x.' +
                 ' Refer to our documentation to see the Node.js versions supported by each version of Azure Functions: https://aka.ms/functions-node-versions';
             systemWarn(msg);
--- a/src/WorkerChannel.ts
+++ b/src/WorkerChannel.ts
@@ -126,8 +126,7 @@ export class WorkerChannel implements IWorkerChannel {
             const msg =
                 'Node.js version used (' +
                 version +
-                ') is not officially supported. You may use it during local development, but must use an officially supported version on Azure.' +
-                ' The version of the Azure Functions runtime you are using (v4) officially supports Node.js v14.x or Node.js v16.x:' +
+                ') is not officially supported. You may use it during local development, but must use an officially supported version on Azure:' +
                 ' https://aka.ms/functions-node-versions';
             this.log({
                 message: msg,
--- a/src/WorkerChannel.ts
+++ b/src/WorkerChannel.ts
@@ -11,7 +11,7 @@ import { toTypedData } from './converters';
 import { IFunctionLoader } from './FunctionLoader';
 import { IEventStream } from './GrpcService';
 import { InternalException } from './utils/InternalException';
-import { systemError, systemWarn } from './utils/Logger';
+import { systemError } from './utils/Logger';
 import LogCategory = rpc.RpcLog.RpcLogCategory;
 import LogLevel = rpc.RpcLog.Level;
 
@@ -127,9 +127,13 @@ export class WorkerChannel implements IWorkerChannel {
                 'Node.js version used (' +
                 version +
                 ') is not officially supported. You may use it during local development, but must use an officially supported version on Azure.' +
-                ' The version of the Azure Functions runtime you are using (v4) officially supports Node.js v14.x or Node.js v16.x.' +
-                ' Refer to our documentation to see the Node.js versions supported by each version of Azure Functions: https://aka.ms/functions-node-versions';
-            systemWarn(msg);
+                ' The version of the Azure Functions runtime you are using (v4) officially supports Node.js v14.x or Node.js v16.x:' +
+                ' https://aka.ms/functions-node-versions';
+            this.log({
+                message: msg,
+                level: LogLevel.Warning,
+                logCategory: LogCategory.System,
+            });
         } else if (!(version.startsWith('v14.') || version.startsWith('v16.'))) {
             const msg =
                 'Incompatible Node.js version' +
--- a/src/WorkerChannel.ts
+++ b/src/WorkerChannel.ts
@@ -11,7 +11,7 @@ import { toTypedData } from './converters';
 import { IFunctionLoader } from './FunctionLoader';
 import { IEventStream } from './GrpcService';
 import { InternalException } from './utils/InternalException';
-import { systemError } from './utils/Logger';
+import { systemError, systemWarn } from './utils/Logger';
 import LogCategory = rpc.RpcLog.RpcLogCategory;
 import LogLevel = rpc.RpcLog.Level;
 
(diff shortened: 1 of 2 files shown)


src/nodejsWorker.ts
```diff
@@ -19,6 +19,13 @@ function validateNodeVersion(version) {
         if (versionSplit.length != 3) {
             message = "Could not parse Node.js version: '" + version + "'";
             // Unsupported version note: Documentation about Node's stable versions here: https://github.com/nodejs/Release#release-plan and an explanation here: https://medium.com/swlh/understanding-how-node-releases-work-in-2018-6fd356816db4
+        } else if (process.env.AZURE_FUNCTIONS_ENVIRONMENT == 'Development' && major == 'v17') {
```
ejizba: Let's have this logic directly mirror the `supportedVersions` logic. So create a new array (maybe called `devOnlyVersions`?) and do `indexOf` just like below.

src/nodejsWorker.ts
```diff
@@ -19,6 +19,13 @@ function validateNodeVersion(version) {
         if (versionSplit.length != 3) {
             message = "Could not parse Node.js version: '" + version + "'";
             // Unsupported version note: Documentation about Node's stable versions here: https://github.com/nodejs/Release#release-plan and an explanation here: https://medium.com/swlh/understanding-how-node-releases-work-in-2018-6fd356816db4
+        } else if (process.env.AZURE_FUNCTIONS_ENVIRONMENT == 'Development' && major == 'v17') {
+            const warning =
+                'Node.js version used (' +
```
ejizba: IMO we should allow v15 as well. I don't really see a downside and it's less confusing (otherwise people might ask why we skipped one number).

src/nodejsWorker.ts
```diff
@@ -19,6 +19,13 @@ function validateNodeVersion(version) {
         if (versionSplit.length != 3) {
             message = "Could not parse Node.js version: '" + version + "'";
             // Unsupported version note: Documentation about Node's stable versions here: https://github.com/nodejs/Release#release-plan and an explanation here: https://medium.com/swlh/understanding-how-node-releases-work-in-2018-6fd356816db4
+        } else if (process.env.AZURE_FUNCTIONS_ENVIRONMENT == 'Development' && major == 'v17') {
+            const warning =
+                'Node.js version used (' +
+                version +
+                ') is not officially supported. You may use it during local development, but your function will be deployed to a different Node.js version on Azure.' +
+                ' Refer to our documentation to see the Node.js versions supported by each version of Azure Functions: https://aka.ms/functions-node-versions';
```
ejizba: nit: maybe we can shorten this a bit

```suggestion
                ') is not officially supported. You may use it during local development, but must use an officially supported version on Azure: https://aka.ms/functions-node-versions';

```

src/WorkerChannel.ts
```diff
@@ -119,7 +119,15 @@ export class WorkerChannel implements IWorkerChannel {
...
-        if (!(version.startsWith('v14.') || version.startsWith('v16.'))) {
+        if (version.startsWith('v17.') && process.env.AZURE_FUNCTIONS_ENVIRONMENT == 'Development') {
+            const msg =
+                'Node.js version used (' +
+                version +
+                ') is not officially supported. You may use it during local development, but your function will be deployed to a different Node.js version on Azure.' +
+                ' The version of the Azure Functions runtime you are using (v4) officially supports Node.js v14.x or Node.js v16.x.' +
+                ' Refer to our documentation to see the Node.js versions supported by each version of Azure Functions: https://aka.ms/functions-node-versions';
+            systemWarn(msg);
+        } else if (!(version.startsWith('v14.') || version.startsWith('v16.'))) {
```
ejizba, line 136: Huh...any thoughts on why we have duplicated logic to check the version in this file and the other?

My hunch is that it was only because of the "_v1WorkerBehavior" flag on the v2.x branch: https://github.com/Azure/azure-functions-nodejs-worker/blob/v2.x/src/WorkerChannel.ts#L131 and that we can get rid of the logic in this file since v3.x branch doesn't have any backwards compat flag like that.

hossam-nasr: I'm not sure why there is this double logic, but it is needed on this branch too. I only modified this file because during my tests, function apps with node version ~17 would error out here.

ejizba: Yeah I agree you'll probably need to adjust the logic in both places, but let's still try to avoid duplicating as much as we can. For example, we want to make sure the warning is only printed once

src/WorkerChannel.ts
```diff
@@ -119,7 +119,18 @@ export class WorkerChannel implements IWorkerChannel {
     public workerInitRequest(requestId: string, msg: rpc.WorkerInitRequest) {
         // Validate version
         const version = process.version;
-        if (!(version.startsWith('v14.') || version.startsWith('v16.'))) {
+        if (
+            (version.startsWith('v17.') || version.startsWith('v15.')) &&
+            process.env.AZURE_FUNCTIONS_ENVIRONMENT == 'Development'
+        ) {
+            const msg =
+                'Node.js version used (' +
```
ejizba, line 127: This message is like _really_ long haha. Can we use the message from the other file instead please?

src/WorkerChannel.ts
```diff
@@ -119,7 +119,18 @@ export class WorkerChannel implements IWorkerChannel {
...
+            (version.startsWith('v17.') || version.startsWith('v15.')) &&
+            process.env.AZURE_FUNCTIONS_ENVIRONMENT == 'Development'
+        ) {
+            const msg =
+                'Node.js version used (' +
+                version +
+                ') is not officially supported. You may use it during local development, but must use an officially supported version on Azure.' +
+                ' The version of the Azure Functions runtime you are using (v4) officially supports Node.js v14.x or Node.js v16.x.' +
+                ' Refer to our documentation to see the Node.js versions supported by each version of Azure Functions: https://aka.ms/functions-node-versions';
+            systemWarn(msg);
```
ejizba: Try `this.log` (there's other examples in this file) and see if that works better

src/WorkerChannel.ts
hossam-nasr: Yeah I agree with you. I just adjusted the code so that here it's a gRPC log (makes it to the host) and the other one is a node process log

src/WorkerChannel.ts
hossam-nasr: Haha I've shortened the message a bit, but I think it's useful to explicitly mention which node versions are supported on this runtime, following the lead of the other error message in this file. I also think it's good to have a somewhat long warning to attract attention. If you feel strongly about this, though, I can just copy the exact message from the other file.

ejizba: > I think it's useful to explicitly mention which node versions are supported on this runtime

The reason I removed the explicit versions in my suggestion was because it's too easy for those to get out of date. If you want to keep the versions, you could address that concern with a `const` used in both the message and the `if` statement, which ensures they stay in sync. Basically, a more complicated version of this:
```typescript
const supportedVersion = '1';
if (version !== supportedVersion) {
    const message = `You gotta be on version ${supportedVersion}`;
}
```

> I also think it's good to have a somewhat long warning to attract attention.

At least in core tools, it'll show up bright yellow which serves a similar purpose. Subjectively, I would argue users are more likely to skip long messages because they (me too) are lazy. Objectively, we should keep messages short to keep the logs (as a whole) more manageable. This message likely feels more important because of recency bias, but ultimately it's no more or less important than existing logs/warnings. I wouldn't block a PR on this, but I would likely continue to comment if I see long messages

src/WorkerChannel.ts
ejizba: I'm concerned that the only reason these logs are currently separate is because we're running in a development environment with our local worker build. I worry once we integrate a real version of the node worker into the core tools, it'll show the warning twice in one place. Are you confident that won't happen? I just don't know enough about how the logs work to be sure.

Otherwise, we could do something like this in `nodejsWorker.ts`:
```
} else if (process.env.AZURE_FUNCTIONS_ENVIRONMENT == 'Development' && devOnlyVersions.indexOf(major) >= 0) {
    // Let users continue here. Warning will be shown in the workerInitRequest
}
```

hossam-nasr: I'm fairly confident that it won't happen. I tried starting my function app from core tools without specifying a local worker directory, and I didn't get any logs from the worker's console, which I would expect to see (e.g. ```systemLog(`Worker ${workerId} connecting on ${connection}`);``` or ```systemLog(`Worker ${workerId} exited with code ${code}`);``` in `Worker.ts`.) Also, when I trigger an exception by using `v17.x` of node, I don't see the `logPreix = "LanguageWorkerConsoleLog"` at the beginner of the error message, indicating the error in the core tools is coming from `throw` statement, not from the `console.error`.

ejizba: Thanks for the investigation! It felt wrong that we weren't seeing any of those logs, so I looked into it further. Turns out you can view them by adding this to your local.settings.json: `"logging:logLevel:Worker": "debug"` (more details [here](https://github.com/anthonychu/functions-log-suppression)). I think it's a bug that all worker console logs are currently categorized as "Debug" so I submitted this PR to fix: https://github.com/Azure/azure-functions-host/pull/8102

Assuming that gets merged, the console warnings _would_ be displayed by default

hossam-nasr: Thanks for looking into this! I confirmed this by adding that setting to my local.settings.json, and indeed the warning message was displayed twice (although the one from `console.warn` was not yellow or clear at all that it was a warning -- hopefully this gets fixed by your PR). I went ahead and removed this duplicate warning. Good find!

ejizba (approved):

PR #525, opened 2022-01-28
Add tip to show worker logs to README

ejizba:


README.md
```diff
@@ -23,7 +23,9 @@
 - In the local function app, add the following settings to your "local.settings.json" file or configure them directly as environment variables
   - `languageWorkers__node__workerDirectory`: `<path to the root of this repository>`
   - `languageWorkers__node__arguments`: `--inspect`
-    > NOTE: If you need to debug worker initialization, use `--inspect-brk` instead of `--inspect`. Just keep in mind you need to attach the debugger within 30 seconds or the host process will timeout.
+    > 💡 Tip #1: Set `logging__logLevel__Worker` to `debug` if you want to view worker-specific logs in the output of `func start`
```
hossam-nasr, line 26: Isn't it `logging:logLevel:Worker`? That's what got it to work for me in `local.settings.json` at least.

ejizba: `:` and `__` are usually equivalent, except `:` doesn't work in all situations. Here's an example doc that mentions it: https://docs.microsoft.com/en-us/aspnet/core/fundamentals/configuration/?view=aspnetcore-6.0#environment-variables

hossam-nasr (approved):

Issue #168, opened 2019-01-04
The charset of content-type request header is ignored


yoichiro:
#### Abstract

It seems that a charset of the content-type response header is ignored in specific case. If a request has an Accept-Charset request header including "utf-16, utf-8" value, the Azure Functions returns a response that has a response body encoded by UTF-16LE ignoring a charset in the "content-type" response header and has a "charset=utf-16" (this means "UTF-16BE") in the "Content-Type" response header, therefore the client cannot handle the response.

#### Expected Behavior

* The Azure Functions should determines the charset of the response body according to the charset that is set as the "content-type" response header by a code without using the "Accept-Charset" request header value. That is, if the "charset=utf-8" specified, the response body charset should become UTF-8. If the "charset=utf-16" or "charset=utf-16be specified, the response body charset should become UTF-16BE. And, if the "charset=utf-16le" specified, the response body charset should become UTF-16LE.
* Otherwise, the charset of the content-type response header should become "charset=utf-16le" to fit the response body UTF-16LE charset. Or, the charset of the response body is changed to "UTF-16BE".

#### Detail of Current Behavior

For example, it assumes that I write the following code to return the UTF-8 content:

```js
module.exports = (context, req) => {
    context.res = {
        status: 200,
        headers: {
            'content-type': 'application/json; charset=utf-8'
        },
        body: {
            'fulfillmentText': 'Hi, Yoichiro!'
        }
    };
    context.done();
};
```

If a request has an Accept-Charset request header with "utf-8" value only, the response is the following:

* Content-Type: application/json; charset=utf-8
* Response Body Charset: UTF-8

This above is a correct behavior.

Next, if a request has an Accept-Charset request header including "utf-16,utf-8", the response is the following:

* Content-Type: application/json; charset=utf-16
* Response Body Charset: UTF-16LE

This is a bad behavior. The reason is follows:

If the response body charset (UTF-16LE) is correct (that is, the content-type response header set by the code should be ignored) as the specification, the "charset=utf-16" of the content type is wrong. Because, the "utf-16" means "UTF-16BE (Big Endian)", not "UTF-16LE (Little Endian)". This specification is described by the [RFC 2781](https://tools.ietf.org/html/rfc2781#section-4.3):

[...]

PR #529, opened 2022-02-10
Add interface for HTTP Respones in types package

hossam-nasr:
Adds two interfaces for HTTP responses. Implements #169 

### **Returning a response as an object**

The `HttpResponseObject` interface defines the properties that the user can set when returning a response object in the HTTP module-like way. Those properties are passed to the host and define the response sent back to the client. An example of using this interface:

```TypeScript
const httpTrigger: AzureFunction = async function(context: Context, req: HttpRequest): Promise<void> {
    context.res = {
        status: 200,
        body: "hello, world",
        headers: {
            hello: "world"
        }
   }
};
````

### **Returning a response using methods**

The `HttpResponseApi` interface defines the methods exposed by the `context.res` object passed to the function when using HTTP triggers. This can be used to return a response in the express-like way. An example of using this interface:

```TypeScript
const httpTrigger: AzureFunction = async function (context: Context, req: HttpRequest): Promise<void> {
    context.res = context.res as HttpResponseApi;
    context.res.setHeader("hello", "world");
    context.res.status(200);
    context.res.body = "hello, world";
}; 
```

### **Design choices**
The `res` property of the `Context` interface is set to be `type HttpResponse = HttpResponseApi | HttpResponseObject`

One challenge with implementing this interface is we need to support both ways of returning an HTTP response, both of which use the `Context.res` property. This means that `context.res.setHeader()`, for example, *could* be `undefined`. As a result, while JS code could use either ways of returning an HTTP response without an issue, TS code would throw an error when trying to call `context.res.setHeader()` without some sort of type guard/check to make sure the property is callable. 

There are several ways to mitigate this. One way is to cast `context.res` to an `HttpResponseApi` object, as used above:

```TypeScript
context.res = context.res as HttpResponseApi;
context.res.setHeader("hello", "world");
```

Another way would be to explicitly check the type of methods before calling them:

```TypeScript
if(context.res?.status instanceof Function) {
    context.res.status(200)
}
```

Users can also define their own types and use those throughout their project:

```TypeScript
export type ActualContext = Context & { res: HttpResponseApi }

const httpTrigger: AzureFunction = async function (context: ActualContext, req: HttpRequest): Promise<void> {
    context.res.status(200);
}; 
```

Whichever way this is done, TS code that called those functions without any checks, relying on the type of `context.res` being `[key: string]: any` will need at least some changes to continue working with no errors.


types/index.d.ts
```diff
@@ -85,16 +85,20 @@ declare module '@azure/functions' {
         /**
          * HTTP response object. Provided to your function when using HTTP Bindings.
          */
-        res?: {
-            [key: string]: any;
-        };
+        res?: HttpResponse;
```
ejizba: As discussed, let's leave this as-is for now

types/index.d.ts
```diff
@@ -183,6 +187,138 @@ declare module '@azure/functions' {
...
+         * @param type the value to set header 'Content-Type' to
+         * @returns the updated HttpResponseApi instance
+         */
+        type(type: string): HttpResponseApi;
+        /**
+         * Automatically sets the content-type then calls context.done()
+         * @param body (optional) body content to send
+         * @returns updated HttpResponseApi instance
+         * @deprecated this method calls context.done() which is deprecated, use async/await and pass the response as the return value instead.
+         * See the docs here for more information: https://docs.microsoft.com/en-us/azure/azure-functions/functions-reference-node?tabs=v2#use-async-and-await
```
ejizba: Let's create an aka.ms link for this, ping me for details if you need help since I assume you haven't done that before

Also - always make sure to remove the "en-us" part of the link so that the user's language will be respected

types/index.d.ts
```diff
@@ -183,6 +187,138 @@ declare module '@azure/functions' {
      * Possible values for an HTTP Request user type
      */
     export type HttpRequestUserType = 'AppService' | 'StaticWebApps';
+    /**
+     * HTTP response object
+     */
+    export interface IResponse {
```
ejizba: We avoid prefixing interfaces with `I`. Some people think it's helpful (not me), but it's 
a lot more popular in .NET, not Node.js. There's whole debates about it if you google it and want to know more, for example https://github.com/microsoft/TypeScript-Handbook/issues/121

Maybe use `HttpResponseBase` or `HttpResponseShared`

types/index.d.ts
```diff
@@ -183,6 +187,138 @@ declare module '@azure/functions' {
...
+        statusCode?: number | string;
+        /**
+         * Enables content negotiation using ASP.NET core if true
+         * If false, treat response as raw (default)
+         * @default false
+         */
+        enableContentNegotiation?: boolean;
+    }
+    /**
+     * Http response object and methods. Provided to your function when using HTTP triggers.
```
ejizba: We need to make sure we touch on the differences between the two interfaces. Maybe something like 
> This is the default type on the `Context` object provided to your function when using HTTP triggers.

And maybe for the other one:

> This is not the default on the `Context` object, but you may replace `context.res` with an object of this type

types/index.d.ts
```diff
@@ -183,6 +187,138 @@ declare module '@azure/functions' {
...
+         * HTTP response status code.
+         * @default 200
+         */
+        statusCode?: number | string;
+        /**
+         * Enables content negotiation using ASP.NET core if true
+         * If false, treat response as raw (default)
+         * @default false
+         */
+        enableContentNegotiation?: boolean;
```
ejizba, line 265: We may have discussed this already, but unless we see this used a lot in issue reporting, etc. - I'd leave it off. It's just a bit confusing how it's related to ASP.NET core in a Node.js environment

We can always add it back if anyone asks for it

types/index.d.ts
```diff
@@ -183,6 +187,138 @@ declare module '@azure/functions' {
...
+    }
+    /**
+     * Http response object and methods. Provided to your function when using HTTP triggers.
+     */
+    export interface HttpResponseApi extends IResponse {
```
ejizba: @anthonychu @stefanushinardi we would love your help picking names for the two types of response interfaces (see Hossam's PR description for examples of both).

Marie has described them in the past as "express-like" and "http-module-like".

One option I thought of was `HttpResponseDefault` (for the express like) and `HttpResponseAlt` for the other one.

Of course, I'm hoping we clean all of this up with the new programming model anyways

types/index.d.ts
```diff
@@ -183,6 +187,138 @@ declare module '@azure/functions' {
...
+        getHeader(field: string): any;
+        /**
+         * Has the same functionality as getHeader
+         * Get the value of a particular header field
+         * @param field the name of the header field to get
+         */
+        get(field: string): any;
+        /**
+         * Removes a particular header field
+         * @param field the name of the header field to remove
```
ejizba: Do we find these lines helpful? I feel like a method description is always necessary, but a lot of these param lines don't tell me anything more than the param name itself

types/index.d.ts
```diff
@@ -183,6 +187,138 @@ declare module '@azure/functions' {
...
+         * @param statusCode status code to send
+         * @returns updated HttpResponseApi instance
+         * @deprecated this method calls context.done() which is deprecated, use async/await and pass the response as your function's return value instead.
+         * See the docs here for more information: https://docs.microsoft.com/en-us/azure/azure-functions/functions-reference-node?tabs=v2#use-async-and-await
+         */
+        sendStatus(statusCode: string | number): HttpResponseApi;
+        /**
+         * Sets the 'Content-Type' header to 'application/json' then calls send(body)
+         * @param body (optional) body content to send
+         * @deprecated this method calls context.done() which is deprecated, use async/await and pass the response as your function's return value instead.
```
ejizba, line 331: Doesn't have to be this PR, but we should add similar deprecated tags to the other context.done stuff in this file (at least context.done itself haha)

hossam-nasr: Agreed. I was gonna do this in a different PR so this PR isn't doing too much

types/index.d.ts
hossam-nasr: Done

types/index.d.ts
hossam-nasr: Done. I wasn't sure what to name the aka.ms link, so feel free to suggest alternate names

types/index.d.ts
hossam-nasr: Wow, that was an interesting rabbit hole to go down lol, thanks for sharing. I named it `IResponse` just because that's what `Context.res` used before. I didn't know it was that deep lol. Anyway I changed it to `HttpResponseBase` now.

types/index.d.ts
hossam-nasr: Yeah good callout. I added those

types/index.d.ts
hossam-nasr: Honestly, for the sake of completeness, I would argue to keep it. If any user, even if 1% of the user base, is already using this property in their code (I assume it had a valid use-case, even for node), then attempting to use the new interface would break their existing code. If the description is confusing, then I would opt for having a more accurate description or a less confusing one over removing it from the interface entirely.

types/index.d.ts
hossam-nasr: @anthonychu @stefanushinardi ping

types/index.d.ts
hossam-nasr: Yeah good point, I removed the `@param` lines.

types/index.d.ts
ejizba: I don't care about the name - usually I just use the generated url. The reason I wanted an aka.ms link was so that we can change the destination url at any time without having to re-release the types package.

types/index.d.ts
ejizba: > I assume it had a valid use-case, even for node

This comment implies it doesn't actually do anything: https://github.com/Azure/azure-functions-nodejs-worker/issues/168#issuecomment-454657844. If you can verify it actually works, then I'm okay including it based on your reasoning

types/index.d.ts
```diff
@@ -229,6 +235,125 @@ declare module '@azure/functions' {
      * Possible values for an HTTP Request user type
      */
     export type HttpRequestUserType = 'AppService' | 'StaticWebApps';
+    /**
+     * HTTP response object
```
ejizba: I think we should copy these 4-5 properties to both `HttpResponseApi` and `HttpResponseObject` and get rid of the base class. I just worry we're going to confuse people with too many `HttpResponse...` interfaces and I don't think users should ever directly use this one.

IMO we should treat this file more as "docs" than "source code", meaning readability can outweigh "NO COPYING CODE".

types/index.d.ts
hossam-nasr: Yeah makes sense :)

types/index.test.ts
```diff
@@ -3,7 +3,7 @@
 
 // This file will be compiled by multiple versions of TypeScript as decribed in ./test/TypesTests.ts to verify there are no errors
 
-import { AzureFunction, Context, Cookie, HttpMethod, HttpRequest, Timer } from '@azure/functions';
+import { AzureFunction, Context, Cookie, HttpMethod, HttpRequest, HttpResponseApi, HttpResponseObject, Timer } from '@azure/functions';
```
ejizba: I suggest you disable this rule for the whole file:
```
/* eslint-disable @typescript-eslint/no-unused-vars */
```
This file is a bit weird, so some of the rules don't make sense.

types/index.test.ts
```diff
@@ -77,6 +77,60 @@ const cookieFunction: AzureFunction = async function (context: Context) {
...
+    context.res.setHeader('hello', 'world');
+    context.res.set('hello', 'world');
+    context.res.header('hello', 'world');
+    const hello: string = context.res.get('hello');
+    const hello2: string = context.res.getHeader('hello');
+    context.res.removeHeader('hello');
+    context.res.type('application/json');
+    context.res.body = {
+        hello,
+        hello2
```
ejizba: Are you using VS Code for your local development? For me at least, it auto-fixes lint errors like this one (where it wants a comma after hello2) when I save the file. If that's not happening for you, we might want to double check the repo settings in `.vscode`. Also double check you have the recommended extensions installed (specifically the one for eslint)

types/index.test.ts
```diff
@@ -88,7 +142,45 @@ const returnWithContextDone: AzureFunction = function (context: Context, _req: H
...
 
-export { runHttp, cookieFunction, runHttpReturn, runServiceBus, runFunction, runHttpWithQueue, returnWithContextDone };
+const returnWithSend: AzureFunction = function (context: Context, req: HttpRequest) {
+    context.res = context.res as HttpResponseApi;
+    context.res.send('hello world');
+};
+
+const returnWithEnd: AzureFunction = function (context: Context, req: HttpRequest) {
+    context.res = context.res as HttpResponseApi;
+    context.res.end('hello world');
```
ejizba, line 159: The point of this file is to verify the types, not run any of the code. So you could just move these `end`, `send`, etc. calls into one function instead of separating it out like this

types/index.d.ts
hossam-nasr: This is just a silly example, but I am sure there are other ways it could be used.

Using `enableContentNegotiation: false` (default):

```TS
context.res = {
    status: 200,
    body: "{'hello': 'world', 'test': 123}",
    headers: {
        'content-type': 'application/json'
    },
    enableContentNegotiation: false
};
```

Returns the response body as `{'hello': 'world', 'test': 123}`, which is an invalid `application/json` response.

Using `enableContentNegotiation: true`:

```TS
context.res = {
    status: 200,
    body: "{'hello': 'world', 'test': 123}",
    headers: {
        'content-type': 'application/json'
    },
    enableContentNegotiation: true
};
```

returns the body as `"{'hello': 'world', 'test': 123}"` (added double quotes), which is a valid `application/json` response.

types/index.d.ts
hossam-nasr: Yeah that makes sense. Do you think it'd be helpful to not export the base interface so that users can't directly use it? Or is it more about the readability of this file itself?

ejizba: What do you mean "not export the base interface"? I don't think we need it at all

types/index.test.ts
hossam-nasr: Done :)

types/index.test.ts
hossam-nasr: Yeah good callout. I am using VSCode and I have the extensions installed but I didn't have the npm package `eslint-plugin-deprecation`, so when you added that it broke ESLint for me. Running `npm i` fixed the issue, so I fixed the linting errors too

types/index.test.ts
hossam-nasr: Makes sense

types/index.d.ts
ejizba: Yes I was talking about the readability of this file. I use "go to definition" and read typescript definition files all the time. I basically prefer those over online docs. Plus if the file itself is readable, that usually means the intellisense works better

hossam-nasr: Cool yeah makes sense.

src/http/Response.ts
```diff
@@ -1,23 +1,12 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License.
 
-import { Cookie } from '@azure/functions';
-import { HeaderName, MediaType } from '../constants';
+/* eslint-disable deprecation/deprecation */
```
ejizba: I'm okay disabling this rule on a per-file basis for a _test_ file, but not a _src_ file. Please disable on a per-line basis in src

types/index.d.ts
```diff
@@ -229,6 +235,143 @@ declare module '@azure/functions' {
...
+         */
+        body?: any;
+        /**
+         * HTTP response status code.
+         * @default 200
+         */
+        statusCode?: number | string;
+        /**
+         * HTTP response status code
+         * @default 200
```
ejizba, line 362: Please add notes like this

to statusCode:
> This property takes priority over the `status` property

to status:
> The same as `statusCode`. This property is ignored if `statusCode` is set

ejizba (commented):
Other than the names and one last comment, I think we're g2g

types/index.d.ts
hossam-nasr: Done

types/index.d.ts
hossam-nasr: As discussed, I have renamed `HttpResponseApi` and `HttpResponseObject` to `HttpResponseFull` and `HttpResponseSimple` respectively.

ejizba (approved):

PR #535, opened 2022-02-14
Turn on several lint rules 🧹

ejizba:
Unlike the [last round](https://github.com/Azure/azure-functions-nodejs-worker/pull/437) of lint rules I added, none of these could be auto-fixed. However, they were still pretty trivial manual changes. I split each rule up into its own separate commit - I figured that might be easier to view.

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,6 +14,8 @@
     "devDependencies": {
         "@types/mocha": "^2.2.48",
         "@types/chai": "^4.2.22",
+        "@types/chai-as-promised": "^7.1.5",
+        "@types/mock-require": "^2.0.1",
         "@types/node": "^16.9.6",
         "@types/sinon": "^2.3.7",
         "@typescript-eslint/eslint-plugin": "^4.33.0",
--- a/package.json
+++ b/package.json
@@ -25,6 +25,7 @@
         "eslint-config-prettier": "^8.3.0",
         "eslint-plugin-header": "^3.1.1",
         "eslint-plugin-prettier": "^4.0.0",
+        "eslint-plugin-deprecation": "^1.3.2",
         "eslint": "^7.32.0",
         "chalk": "^4.0.0",
         "jsdoc": "^3.6.3",
--- a/src/FunctionLoader.ts
+++ b/src/FunctionLoader.ts
@@ -31,7 +31,6 @@ export class FunctionLoader implements IFunctionLoader {
         if (scriptFilePath.endsWith('.mjs')) {
             if (this.allowESModules) {
                 // IMPORTANT: pathToFileURL is only supported in Node.js version >= v10.12.0
-                // @ts-ignore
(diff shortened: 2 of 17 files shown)


package.json
```diff
@@ -14,6 +14,8 @@
     "devDependencies": {
         "@types/mocha": "^2.2.48",
         "@types/chai": "^4.2.22",
+        "@types/chai-as-promised": "^7.1.5",
+        "@types/mock-require": "^2.0.1",
```
hossam-nasr, line 18: Are the types packages required to satisfy the linting rules? If it's just for added elegance can we move this to a different PR?

src/Worker.ts
```diff
@@ -39,7 +36,7 @@ export function startNodeWorker(args) {
         throw new InternalException(exception);
     }
 
-    const workerChannel = new WorkerChannel(workerId, eventStream, new FunctionLoader());
+    new WorkerChannel(workerId, eventStream, new FunctionLoader());
```
hossam-nasr, line 39: If it's created and never used do we even need this line?

src/FunctionLoader.ts
```diff
@@ -50,7 +49,7 @@ export class FunctionLoader implements IFunctionLoader {
         }
         const entryPoint = <string>(metadata && metadata.entryPoint);
         const userFunction = getEntryPoint(script, entryPoint);
-        if (!isFunction(userFunction)) {
+        if (typeof userFunction !== 'function') {
```
hossam-nasr, line 52: Just out of curiosity, is there a difference between using `typeof userFunction !== 'function'` and `!(userFunction instanceof Function)`?

src/Worker.ts
ejizba: This is an anti-pattern that I would like to fix in a separate PR. The constructor for the WorkerChannel has side effects, even though the workerChannel is never used - which means it should be a function, not a class

package.json
ejizba: They tell me if I'm importing the modules correctly which was very useful when fixing the "no-var-requires" rule: https://github.com/Azure/azure-functions-nodejs-worker/pull/535/commits/86cba78e9a42f216fa7f9ba097eb45c54071e180

I plan to rebase when merging this PR (unless you have major objections) and I feel like adding the types when turning on that rule makes sense. Each commit by itself is relatively small and encapsulated

src/FunctionLoader.ts
ejizba: Tbh, I've never seen anyone use `instanceof Function` 😝  I don't know all the nuances, but here are a few points of reference:
- The docs for "isFunction" say to use "typeof" https://nodejs.org/api/all.html#all_util_utilisfunctionobject
- https://stackoverflow.com/questions/899574/what-is-the-difference-between-typeof-and-instanceof-and-when-should-one-be-used

hossam-nasr (approved):

PR #536, opened 2022-02-14
Revert "Serialize falsy values for Durable Functions (#383)" in v2.x

ejizba:
Reverts https://github.com/Azure/azure-functions-nodejs-worker/pull/383, which has never been released in the host for this major version. The reason we're reverting is because this could potentially break users who depend on the existing behavior. There was originally some discussion that this may be "worth it" for v2.x because of how incorrect the behavior was, but that discussion is somewhat moot now because people can migrate to v3.x of the worker which had this change from the start.


davidmrdavid (approved):
LGTM

PR #544, opened 2022-02-23
Move event stream setup out of WorkerChannel

ejizba:
This is a continuation of https://github.com/Azure/azure-functions-nodejs-worker/pull/543 where I try to slim down `WorkerChannel`. The main problems addressed by this PR:
1. The constructor of `WorkerChannel` has side effects, which is confusing and an anti-pattern in Node.js (as discussed [here](https://github.com/Azure/azure-functions-nodejs-worker/pull/535#discussion_r807195686))
2. The `eventStream.on('data'` logic felt too "auto-magic" to me, which made it confusing. My new data handler is intended to be more straightforward and it has better type safety (aka no more stuff like `<string>`, `<any>`, `.apply()`)

### DIFF
--- a/src/eventHandlers/invocationRequest.ts
+++ b/src/eventHandlers/invocationRequest.ts
@@ -5,6 +5,7 @@ import { format } from 'util';
 import { AzureFunctionsRpcMessages as rpc } from '../../azure-functions-language-worker-protobuf/src/rpc';
 import { CreateContextAndInputs, LogCallback, ResultCallback } from '../Context';
 import { toTypedData } from '../converters';
+import { nonNullProp } from '../utils/nonNull';
 import { toRpcStatus } from '../utils/toRpcStatus';
 import { WorkerChannel } from '../WorkerChannel';
 import LogCategory = rpc.RpcLog.RpcLogCategory;
@@ -15,8 +16,8 @@ import LogLevel = rpc.RpcLog.Level;
  * @param requestId gRPC message request id
  * @param msg gRPC message content
  */
-export function invocationRequest(channel: WorkerChannel, requestId: string, msg: rpc.InvocationRequest) {
-    const info = channel.functionLoader.getInfo(msg.functionId);
+export function invocationRequest(channel: WorkerChannel, requestId: string, msg: rpc.IInvocationRequest) {
+    const info = channel.functionLoader.getInfo(nonNullProp(msg, 'functionId'));
     const logCallback: LogCallback = (level, category, ...args) => {
         channel.log({
             invocationId: msg.invocationId,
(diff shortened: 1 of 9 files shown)


src/eventHandlers/invocationRequest.ts
```diff
@@ -15,8 +16,8 @@ import LogLevel = rpc.RpcLog.Level;
  * @param requestId gRPC message request id
  * @param msg gRPC message content
  */
-export function invocationRequest(channel: WorkerChannel, requestId: string, msg: rpc.InvocationRequest) {
-    const info = channel.functionLoader.getInfo(msg.functionId);
+export function invocationRequest(channel: WorkerChannel, requestId: string, msg: rpc.IInvocationRequest) {
+    const info = channel.functionLoader.getInfo(nonNullProp(msg, 'functionId'));
```
hossam-nasr, line 20: I don't necessarily see the benefit of generalizing to the interface, especially when the class avoids the need for using `nonNullProp` here?

src/setupEventStream.ts
```diff
@@ -0,0 +1,86 @@
...
+    eventStream.write = function checkWrite(msg) {
+        const msgError = rpc.StreamingMessage.verify(msg);
+        if (msgError) {
+            systemError(`Worker ${workerId} malformed message`, msgError);
+            throw new InternalException(msgError);
+        }
+        oldWrite.apply(eventStream, [msg]);
+    };
+
+    return channel;
```
hossam-nasr: Do we need this return value? If we do, I would prefer delegating the creation of the channel to `WorkerChannel.ts` then passing it to this function as an argument. `setupEventStream` for me doesn't imply that it's also creating the worker channel on my behalf and returning it to me

src/eventHandlers/invocationRequest.ts
ejizba: If I leave it as `InvocationRequest`, I get this error in `setupEventStream.ts`:

> Argument of type 'IInvocationRequest' is not assignable to parameter of type 'InvocationRequest'.
  Property 'toJSON' is missing in type 'IInvocationRequest' but required in type 'InvocationRequest'.

So this was really just a byproduct of my code in `setupEventStream` being more type-safe than the old code in `WorkerChannel.ts`. Essentially, there's no guarantee we'll get an instance of the class here which is why TypeScript threw the error

Also, `nonNullProp` is a little util my old teams and I used pretty regularly so I expect it to be more and more prevalent in our codebase and not much of a "downside"

src/setupEventStream.ts
ejizba: Addressed this

It's probably worth noting that I expect more refactoring work related to `WorkerChannel` moving forward. For example, I've removed so much from it that I'm not really sure it's a "Channel" anymore. One idea I have is to rename it to `WorkerContext`, basically the same thing as `InvocationContext` just at the worker scope. But that would be a separate PR

hossam-nasr (approved):

PR #545, opened 2022-02-25
Refactor unit tests to prioritize async/await over context.done

ejizba:
Our current invocation tests had zero "async/await" tests and only tested "context.done". I fixed it so that we're testing both cases and made a bunch of related improvements
- More explicitly validate each message sent over the rpc channel
- Test a few more edge cases (i.e. functions throwing an error)
- Move several context tests over to invocation tests (otherwise they rely too much on context.done)


src/converters/RpcHttpConverters.ts
```diff
@@ -62,7 +62,10 @@ export function toRpcHttp(inputMessage): rpc.ITypedData {
     const httpMessage: rpc.IRpcHttp = inputMessage;
     httpMessage.headers = toRpcHttpHeaders(inputMessage.headers);
     httpMessage.cookies = toRpcHttpCookieList(inputMessage.cookies || []);
-    const status = inputMessage.statusCode || inputMessage.status;
+    let status = inputMessage.statusCode;
```
ejizba, line 65: The is the only source code I changed in this PR. Basically, my new tests revealed that it was possible for us to stringify the `inputMessage.status` _method_ on `Response` getting this literal value: 

`"status(statusCode) {\n        this.statusCode = statusCode;\n        return this;\n    }"`

(It never affected production though, because somewhere along the way any non-expected status code gets defaulted to 200 instead)

hossam-nasr: Wow, good catch! I guess another argument against overloading the `status` property

hossam-nasr: But does this affect being able to set the response code using the below method?
```TS 
context.res = {
    status: 200
}
```

hossam-nasr: I just tested this.

```TS
context.res = {
    status: 404,
};
```

This would return a 404 status code with the old code, but return 200 with this change.

hossam-nasr: OOoops sorry ignore me I just noticed the rest of your change 🤦🏽‍♂️

hossam-nasr (approved):

PR #547, opened 2022-02-26
Refactor source code to use async/await instead of done

ejizba:
This is the second piece of https://github.com/Azure/azure-functions-nodejs-worker/pull/545, but focused on source code instead of test code. If we're going to encourage users to use "async/await" over "context.done", our own code should default to async/await behavior as well. I removed all logic from "context.done" and it's now a simple event emitter whose sole purpose is to maintain backwards compatibility.

There's very minor changes to the logs (as evidenced by my changes to the test files), but otherwise functionality should be the exact same.

### DIFF
--- a/src/Context.ts
+++ b/src/Context.ts
@@ -22,16 +22,12 @@ import {
 import { FunctionInfo } from './FunctionInfo';
 import { Request } from './http/Request';
 import { Response } from './http/Response';
+import EventEmitter = require('events');
 import LogLevel = rpc.RpcLog.Level;
-import LogCategory = rpc.RpcLog.RpcLogCategory;
 
-export function CreateContextAndInputs(
-    info: FunctionInfo,
-    request: rpc.IInvocationRequest,
-    logCallback: LogCallback,
-    callback: ResultCallback
-) {
-    const context = new InvocationContext(info, request, logCallback, callback);
+export function CreateContextAndInputs(info: FunctionInfo, request: rpc.IInvocationRequest, logCallback: LogCallback) {
+    const doneEmitter = new EventEmitter();
+    const context = new InvocationContext(info, request, logCallback, doneEmitter);
 
     const bindings: ContextBindings = {};
     const inputs: any[] = [];
@@ -76,6 +72,7 @@ export function CreateContextAndInputs(
     return {
         context: <Context>context,
         inputs: inputs,
+        doneEmitter,
     };
 }
 
@@ -95,7 +92,7 @@ class InvocationContext implements Context {
         info: FunctionInfo,
         request: rpc.IInvocationRequest,
         logCallback: LogCallback,
-        callback: ResultCallback
+        doneEmitter: EventEmitter
     ) {
         this.invocationId = <string>request.invocationId;
         this.traceContext = fromRpcTraceContext(request.traceContext);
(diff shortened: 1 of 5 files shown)


ejizba:
ping @alrod

src/Context.ts
```diff
@@ -107,89 +104,32 @@ class InvocationContext implements Context {
...
     bindings: ContextBindings;
 }
 
-export type DoneCallback = (err?: Error | string, result?: any) => void;
-
-export type LogCallback = (level: LogLevel, category: rpc.RpcLog.RpcLogCategory, ...args: any[]) => void;
+export type DoneCallback = (err?: unknown, result?: any) => void;
 
-export type ResultCallback = (err?: any, result?: InvocationResult) => void;
+export type LogCallback = (level: LogLevel, ...args: any[]) => void;
```
alrod: Don't we use `rpc.RpcLog.RpcLogCategory`?

ejizba: All worker/system logic was moved out of `context.done` into `invocationRequest.ts`. The only thing left needing a `LogCallback` inside this file is `context.log`, and that's always going to be user logs. 

I can update the name to be `UserLogCallback` to make clear its not for system logs

test/eventHandlers/invocationRequest.test.ts
```diff
@@ -169,7 +169,7 @@ namespace Msg {
             invocationId: '1',
             message: "Error: Choose either to return a promise or call 'done'.  Do not use both in your script.",
             level: LogLevel.Error,
-            logCategory: LogCategory.User,
+            logCategory: LogCategory.System,
```
alrod: Why are we changing the log category? I assume the errors will go to different places.

ejizba: This was not necessarily a planned change, but I would say my refactor uncovered a "bug" and my easiest course of action was to fix the bug by changing the category here.

To me, the distinction is:
- If the message came from the user, it's a "User" log
- If the message came from us, it's a "System" log

One key difference is I believe we don't track user logs in our own telemetry, just system logs. I think having these particular errors in our own telemetry would be very useful for tracking purposes, detectors, etc.

alrod: yes, it's useful for us as well as for the user - user also need a way to be notified about the issue. Can we add the error to both - user and system?

alrod:
> This is the second piece of #545, but focused on source code instead of test code. If we're going to encourage users to use "async/await" over "context.done", our own code should default to async/await behavior as well. I removed all logic from "context.done" and it's now a simple event emitter whose sole purpose is to maintain backwards compatibility.
> 
> There's very minor changes to the logs (as evidenced by my changes to the test files), but otherwise functionality should be the exact same.

What are the the logs changes? Only category change?

test/eventHandlers/invocationRequest.test.ts
ejizba: I think that would result in duplicate logs. In all my testing, system logs are also shown to the user. In fact, the vast majority of the logs in this repo (15 out of 17ish) were already categorized as `System` (for example, "incorrect Node.js version" [here](https://github.com/Azure/azure-functions-nodejs-worker/blob/v3.x/src/eventHandlers/workerInitRequest.ts#L27) and "Failed to load function" [here](https://github.com/Azure/azure-functions-nodejs-worker/blob/v3.x/src/eventHandlers/functionLoadRequest.ts#L24))

I honestly think the only reason we have the distinction is so that we don't track user logs in our telemetry to protect their privacy. I don't think it affects what gets shown

test/eventHandlers/invocationRequest.test.ts
```diff
@@ -438,8 +438,8 @@ describe('invocationRequest', () => {
         sendInvokeMessage([httpInputData]);
         await stream.assertCalledWith(
             Msg.receivedInvocLog(),
-            Msg.invocResponse([getHttpResponse()]),
-            Msg.asyncAndDoneLog
+            Msg.asyncAndDoneLog,
```
ejizba, line 441: > What are the the logs changes? Only category change?

The only other change has to do with ordering/timing of the logs (that's why these lines are switched in the test). **tl;dr** if the user is using context.done and executing synchronous code after they call context.done the logs will be in a different order. Didn't seem like a big deal because the same exact logs are still displayed and to get into this situation the user is doing a bunch of stuff we explicitly tell them not to do.

Here's a more in-depth analysis if you're curious:

If the user has _synchronous blocking_ code after they call `context.done`, the invocation response will now come after that code finishes. Here's an example:

```typescript
module.exports = function (context: Context, _req: HttpRequest): void {
    context.log('1');
    delaySync(500);
    context.log('2');
    context.done();
    delaySync(500);
    context.log('3');
};

function delaySync(ms: number) {
    const end = Date.now() + ms;
    while (Date.now() < end) {
        // wait
    }
}
```

### Previous Behavior

```
[2022-03-08T01:00:14.830Z] Executing 'Functions.delay' (Reason='This function was programmatically called via the host APIs.', Id=cb92b92f-aa3a-4154-bb0a-66f734a00ca1)
[2022-03-08T01:00:15.916Z] 1
[2022-03-08T01:00:15.916Z] 2
[2022-03-08T01:00:15.929Z] InvocationResponse received for invocation id: 'cb92b92f-aa3a-4154-bb0a-66f734a00ca1'
[2022-03-08T01:00:16.004Z] Executed 'Functions.delay' (Succeeded, Id=cb92b92f-aa3a-4154-bb0a-66f734a00ca1, Duration=1212ms)
[2022-03-08T01:00:16.031Z] Warning: Unexpected call to 'log' on the context object after function execution has completed. Please check for asynchronous calls that are not awaited or calls to 'done' made before function execution completes. Function name: delay. Invocation Id: cb92b92f-aa3a-4154-bb0a-66f734a00ca1. Learn more: https://go.microsoft.com/fwlink/?linkid=2097909 
```

### New Behavior

```
[2022-03-08T01:09:20.582Z] Executing 'Functions.delay' (Reason='This function was programmatically called via the host APIs.', Id=1237a61d-c126-4d0a-8c93-5821409fae84)
[2022-03-08T01:09:21.587Z] 1
[2022-03-08T01:09:21.587Z] 2
[2022-03-08T01:09:21.587Z] Warning: Unexpected call to 'log' on the context object after function execution has completed. Please check for asynchronous calls that are not awaited or calls to 'done' made before function execution completes. Function name: delay. Invocation Id: 1237a61d-c126-4d0a-8c93-5821409fae84. Learn more: https://go.microsoft.com/fwlink/?linkid=2097909 
[2022-03-08T01:09:21.587Z] 3
[2022-03-08T01:09:21.588Z] InvocationResponse received for invocation id: '1237a61d-c126-4d0a-8c93-5821409fae84'
[2022-03-08T01:09:21.588Z] Executed 'Functions.delay' (Succeeded, Id=1237a61d-c126-4d0a-8c93-5821409fae84, Duration=1006ms)
```

I think people doing that kind of sync blocking code after context.done are very rare. Any _non-blocking_ code (like a callback or async method) will have the exact same behavior as before, even if I use 1ms as the timeout like this:
```typescript
module.exports = function (context: Context, _req: HttpRequest): void {
    context.log('1');
    setTimeout(() => {
        context.log('2');
        context.done();
        setTimeout(() => {
            context.log('3');
        }, 1);
    }, 1);
};
```

test/eventHandlers/invocationRequest.test.ts
alrod: > for example, "incorrect Node.js version" [here](https://github.com/Azure/azure-functions-nodejs-worker/blob/v3.x/src/eventHandlers/workerInitRequest.ts#L27) and "Failed to load function" [here](https://github.com/Azure/azure-functions-nodejs-worker/blob/v3.x/src/eventHandlers/functionLoadRequest.ts#L24)

the logs are [console logs](https://github.com/Azure/azure-functions-nodejs-worker/blob/v3.x/src/utils/Logger.ts#L5) which "guarantee the messages reach the host despite potential performance impact."

> I think that would result in duplicate logs. In all my testing, system logs are also shown to the user.

I am wandering what is the difference between grpc channel logs "User" and "System" in app insights? Are the logs shown differently to the customer?

test/eventHandlers/invocationRequest.test.ts
alrod: it looks like a breaking change: `context.log('3');` is not executed in previous behavior, we exit from function code immediately. With new behavior we continue executing even after context.done(). Imagine if the customer has the code after `context.done`. It can break the logic.

```
if ()
{
  context.done();
}
// other code 
```

alrod: Thanks for the clarification. Also we emit log in new behavior after `context.done` but I think it's not a very big deal.

test/eventHandlers/invocationRequest.test.ts
ejizba: > the logs are [console logs](https://github.com/Azure/azure-functions-nodejs-worker/blob/v3.x/src/utils/Logger.ts#L5) which "guarantee the messages reach the host despite potential performance impact."

Sorry, the first link is a console log, but the second link for "function load failure" is a channel.log.

I'm not seeing much difference at all in app insights. For example here's some info I see in the portal in the app insights "Live metrics" blade:
|user|system|
|---|---|
|Screen Shot 2022-03-08 at 6 38 14 PM|Screen Shot 2022-03-08 at 6 38 27 PM|

The only other difference I can think of is how logs get filtered. [This repo](https://github.com/anthonychu/functions-log-suppression) has a good description on that. I believe "User" logs would translate to the "Function.<function_name>.User" category and our system logs would translate to the "Worker.rpcWorkerProcess" category. All of these can be individually filtered by the user so there's a chance they wouldn't see the log if they explicitly turned off all Worker logs, but by default all worker errors and warnings (aka the logs being discussed in this thread) are shown

ejizba: As discussed offline, we discovered some of these logs are currently broken _before_ my changes. I undid my behavior change to the LogCategory and we will fix the problem separately. Issue in the host is here: https://github.com/Azure/azure-functions-host/issues/8222

alrod (approved):

Issue #539, opened 2022-02-17
Provide startup/teardown hooks


ejizba:
Related to https://github.com/Azure/azure-functions-nodejs-worker/issues/522, which is focused on invocation hooks (needed for distributed tracing), there is benefit to startup & teardown hooks as well

A few differences between putting code at the top level of the startup file vs. a startup hook:
1. Startup hooks can use `await` (commonjs files don't support top level `await`, although es6 does)
2. Startup hooks can be passed a context object with extra information about the worker

hossam-nasr:
Startup hooks added in #577. Teardown hooks tracked by #606. Closing this issue

Issue #527, opened 2022-02-01
Switch to native JavaScript private fields


ejizba:
We currently use TypeScript private fields (`private _data: string`), but those are not enforced as soon as the code is compiled into JavaScript. We should use the relatively new JavaScript private fields (`#data: string`) supported by Node.js 12 and later:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_fields

This will prevent users in both TS and JS from using a property/method we do not intend for them to use.

PR #548, opened 2022-03-01
Add pre and post invocation hooks

### LINKED ISSUE #522: Provide function execution hooks
## Sample Usage

**This has been updated to reflect the final design**

```typescript
import { registerHook } from '@azure/functions-core';

registerHook('preInvocation', (context: PreInvocationContext) => {
    // can completely change the inputs passed to a user's function
    context.inputs = ['hello'];
    // can write to `context.hookData` and read the value in a postInvocationHook
    context.hookData.value = 1;
    // can write to the _invocation_ context passed to a user's function
    context.invocationContext.dataForInvoc = 2;
});

const disposable = registerHook('postInvocation', (context: PostInvocationContext) => {
    // can read data from `context.hookData` set in the preInvocationHook
    if (context.hookData.value === 1) {
    }
    // can modify the error thrown by the function
    if (context.error) {
        context.error = new Error(`Add a prefix to error text: ${context.error.message}`);
    }
    // can completely change the result of a function
    context.result = 'world';
});

disposable.dispose(); // Can stop the hook from executing in the future
```

NOTE: If you want to use hooks with esm, you have to import `registerHook` slightly differently:
```typescript
import { createRequire } from "module";
const require = createRequire(import.meta.url);
const { registerHook } = require('@azure/functions-core');

// The rest works the same as above
registerHook('preInvocation', async (context) => {

});
```

## Background

Basically, users should be able to register a hook that runs a piece of code at various points in time (i.e. pre invocation and post invocation) at both the function level and app level (meaning you register it once and it runs for all functions).

The main scenario is to simplify users integrating with the Application Insights sdk for Node.js. Users can already do it, but it's a bit complicated (see docs [here](https://github.com/microsoft/ApplicationInsights-node.js#azure-functions)). If we were to add pre and post invocation hooks, most of that logic could be shifted to the sdk itself instead of the user's code. Some work was done in [this PR](https://github.com/Azure/azure-functions-nodejs-worker/pull/339), but the hooks were never actually exposed or usable.

Also, users could leverage these hooks for any reason (other than app insights) so IMO we need to make sure this is an "official" api. The Python team did this by adding [worker extensions](https://docs.microsoft.com/en-us/azure/azure-functions/develop-python-worker-extensions?tabs=linux%2Cpypi) which we may or may not want to model after. NOTE: It would be much easier for us to support app-level hooks after the new [programming model changes](https://github.com/Azure/azure-functions-nodejs-worker/issues/480).

ejizba:
Fixes https://github.com/Azure/azure-functions-nodejs-worker/issues/522. Importantly, this PR does not address _all_ possible hook-related features, but it should allow for them in the future (hopefully near future). See that issue thread for more details on that discussion.

Sample usage:
```typescript
import { registerHook } from '@azure/functions-core';

registerHook('preInvocation', (context: PreInvocationContext) => {
    // can completely change the inputs passed to a user's function
    context.inputs = ['hello'];
    // can write to `context.hookData` and read the value in a postInvocationHook
    context.hookData.value = 1;
    // can write to the _invocation_ context passed to a user's function
    context.invocationContext.dataForInvoc = 2;
});

const disposable = registerHook('postInvocation', (context: PostInvocationContext) => {
    // can read data from `context.hookData` set in the preInvocationHook
    if (context.hookData.value === 1) {
    }
    // can modify the error thrown by the function
    if (context.error) {
        context.error = new Error(`Add a prefix to error text: ${context.error.message}`);
    }
    // can completely change the result of a function
    context.result = 'world';
});

disposable.dispose(); // Can stop the hook from executing in the future
```

See `workerTypes/index.d.ts` for more details/docs. See [here](https://github.com/ejizba/func-nodejs-prototype#split-up-the-worker) for why I picked the name I did and how I see this built-in module working in the new programming model. Ultimately, though, we could easily rename this module and do something completely different in the future.


src/WorkerChannel.ts
```diff
@@ -170,17 +153,49 @@ export class WorkerChannel {
...
-        for (const after of this._invocationRequestAfter) {
-            after(context);
+    private getHooks(hookName: string): HookCallback[] {
+        switch (hookName) {
+            case 'preInvocation':
+                return this._preInvocationHooks;
+            case 'postInvocation':
+                return this._postInvocationHooks;
+            default:
+                throw new RangeError(`Unrecognized hook "${hookName}"`);
```
hossam-nasr: It's always been a pet peeve of mine when essentially an enum argument is treated as a plain string. I think it makes writing code more error-prone, both from our side and the user's, plus it doesn't make it immediately obvious what are all the 'right' choices. 

Would it be possible to have separate functions for pre/post invocation hooks, or have them be different properties in some object, or different methods within a class? For example something like `registerPreInvocationHook(() => {})` or `registerHook.preInvocation(() => {})`.

At the very least, I would support a custom enum-style type for the argument to this function, rather than a string.

src/WorkerChannel.ts
```diff
@@ -12,12 +13,10 @@ import { IFunctionLoader } from './FunctionLoader';
 import { IEventStream } from './GrpcClient';
 import { InternalException } from './utils/InternalException';
 import { systemError } from './utils/Logger';
+import Module = require('module');
 import LogCategory = rpc.RpcLog.RpcLogCategory;
 import LogLevel = rpc.RpcLog.Level;
 
-type InvocationRequestBefore = (context: Context, userFn: Function) => Function;
```
hossam-nasr, line 14: Looks like before we were allowing pre-invocation hooks access to the user-defined function. Are we not allowing that anymore? It doesn't make sense to me why they would've needed access to it in the first place though

tsconfig.json
```diff
@@ -11,6 +11,9 @@
         "paths": {
             "@azure/functions": [
                 "types"
+            ],
+            "@azure/functions-worker": [
```
hossam-nasr: Honestly, this naming is kind of confusing to me. I read your notes on the naming choices, but if the idea is for the worker to be a stripped down version, essentially only used by us and not exposed to the users, then it doesn't sound to me like registering pre and post invocations hooks should belong here? It sounds to me like this belongs to the 'programming model' package instead.

src/setupWorkerModule.ts
```diff
@@ -0,0 +1,24 @@
...
+
+export function setupWorkerModule(channel: WorkerChannel): void {
+    const workerApi = {
+        registerHook: (hookName: string, callback: HookCallback) => channel.registerHook(hookName, callback),
+        Disposable,
+    };
+
+    Module.prototype.require = new Proxy(Module.prototype.require, {
+        apply(target, thisArg, argArray) {
+            if (argArray[0] === '@azure/functions-worker') {
```
hossam-nasr: Is there a way to make this import also just `@azure/functions`?

If not, I would favor something like `@azure/functions-sdk` or something. If this is public-facing, I think "worker" is internal lingo, and it might not mean too much to a node developer

workerTypes/index.d.ts
```diff
@@ -0,0 +1,83 @@
...
+        hookName: 'postInvocation',
+        callback: (context: PostInvocationContext) => void | Promise<void>
+    ): Disposable;
+    export function registerHook(hookName: string, callback: HookCallback): Disposable;
+
+    export type HookCallback = (context: {}) => void | Promise<void>;
+
+    /**
+     * Context on a function that is about to be executed
+     * This object will be passed to all pre and post invocation hooks and can be used to share information between hooks
```
hossam-nasr: update this comment to indicate this is only for pre-invocation hooks

workerTypes/index.d.ts
```diff
@@ -0,0 +1,83 @@
...
+
+        /**
+         * The input values for this specific invocation. Changes to this array _will_ affect the inputs passed to your function
+         */
+        inputs: any[];
+    }
+
+    /**
+     * Context on a function that has just executed
+     * This object will be passed to all pre and post invocation hooks and can be used to share information between hooks
```
hossam-nasr: Same here -- this object will only be passed to post invocation hooks

workerTypes/index.d.ts
```diff
@@ -0,0 +1,83 @@
...
+        hookName: 'preInvocation',
+        callback: (context: PreInvocationContext) => void | Promise<void>
+    ): Disposable;
+    export function registerHook(
+        hookName: 'postInvocation',
+        callback: (context: PostInvocationContext) => void | Promise<void>
+    ): Disposable;
+    export function registerHook(hookName: string, callback: HookCallback): Disposable;
+
+    export type HookCallback = (context: {}) => void | Promise<void>;
```
hossam-nasr: Similar to an earlier comment, it feels cleaner (and safer) to me to just have separate functions, with separate hook callback types specific to pre and post-invocation.

workerTypes/index.d.ts
```diff
@@ -0,0 +1,83 @@
...
+
+    /**
+     * Context on a function that is about to be executed
+     * This object will be passed to all pre and post invocation hooks and can be used to share information between hooks
+     */
+    export interface PreInvocationContext {
+        /**
+         * The context object passed to the function
+         */
+        invocationContext: Context;
```
hossam-nasr: Could possibly add a comment here that this is the context that can be used to share information between hooks

workerTypes/index.d.ts
```diff
@@ -0,0 +1,83 @@
...
+    export interface PostInvocationContext extends PreInvocationContext {
+        /**
+         * The result of the function, or null if there is no result. Changes to this value _will_ affect the overall result of the function
+         */
+        result: any;
+
+        /**
+         * The error for the function, or null if there is no error. Changes to this value _will_ affect the overall result of the function
+         */
+        error: any;
```
hossam-nasr: I would support explicitly adding a `context` property here, to make it clear that the `context` object can be used to communicate between post and pre invocation hooks.

Perhaps also add the `input` property, since it also ends up being passed to the post invocation context?

src/WorkerChannel.ts
```diff
@@ -1,25 +1,21 @@
...
-type InvocationRequestAfter = (context: Context) => void;
-
 export class WorkerChannel {
     public eventStream: IEventStream;
     public functionLoader: IFunctionLoader;
-    private _invocationRequestBefore: InvocationRequestBefore[];
-    private _invocationRequestAfter: InvocationRequestAfter[];
+    private _preInvocationHooks: HookCallback[] = [];
+    private _postInvocationHooks: HookCallback[] = [];
```
hossam-nasr: For the interest of type safety, I would support adding separate types for pre and post invocation hook callbacks, and reflecting that here.

src/WorkerChannel.ts
ejizba: The "before" code never worked and was contributed by someone on the app insights team who is no longer on that team. I would prefer we basically ignore the previous implementation and assume it never existed. I plan to get feedback from the app insights team before the hooks feature actually goes out

workerTypes/index.d.ts
ejizba: My comment in the code is accurate. This _is_ passed to both pre and post invocation hooks. `PostInvocationContext` is just `PreInvocationContext` with a little more information added on, but it's still the same instance of an object. That's why `PostInvocationContext extends PreInvocationContext` below

workerTypes/index.d.ts
ejizba: As mentioned, the pre invocation context _is_ passed to post invocation hooks, so they can add information anywhere on `PreInvocationContext` and it'll also be on `PostInvocationContext`. I would actually discourage hook-authors (e.g. the app insights team) from writing to `invocationContext`, because a function-author  (e.g. any random user) might be writing information on `invocationContext` and hook-authors wouldn't want to conflict with that

src/setupWorkerModule.ts
ejizba: There's a few reasons I don't want to use `@azure/functions`:
1. That name is currently used for our types-only package. I think users would be confused if they're using something other than a type (an actual function `registerHook`) from that package all of the sudden. It's also confusing because the new module is "built-in" and doesn't need to be listed in a user's package.json, but to get types they _do_ need to list it as a dev dependency
2. I want to stay away from that name until we have the new programming model. It's the "best" name for a package of ours and I find it likely we will repurpose the name in the new programming model. If we're going to repurpose the name, I'd prefer to only do that once.

I'm open to names other than "-worker", although I disagree that "worker" is "internal lingo". A few other names I considered:

`@azure/functions-worker-api`
`@azure/functions-api`

src/WorkerChannel.ts
ejizba: A lot to unpack in that comment, so I've split it up into a few comments in this file

src/WorkerChannel.ts
```diff
@@ -33,32 +29,32 @@ export class WorkerChannel {
...
-        return wrappedFunction;
     }
 
-    public runInvocationRequestAfter(context: Context) {
-        for (const after of this._invocationRequestAfter) {
-            after(context);
+    private getHooks(hookName: string): HookCallback[] {
+        switch (hookName) {
+            case 'preInvocation':
+                return this._preInvocationHooks;
```
ejizba: ## package split

This is a list of bullet points I see for the "@azure/functions-worker" package:
- shipped with the worker (aka built-in module. _not_ listed in user's package.json)
- slow to update
- forced on the user when it does update
- backwards compatability is paramount
- barebones API, very unopinionated

_If_ we add a new programming model package down the line, this would be its bullet points:
- shipped with the user's app (aka listed in the user's package.json dependencies)
- fast to update
- user decides when to take the update
- backwards compatability is nice, but we can easily do a new major version
- fancy api with lots of nice intellisense, can be opinionated

I do like your "separate function" suggestion, but I think that matches the second set of bullet points better. I actually already had it in my prototype [here](https://github.com/ejizba/func-nodejs-prototype/blob/main/types/index.d.ts#L57). For example, say we added a new hook "beforeDinner" at some point and a user tries to use that in their app. If they're on the wrong version of the host/worker in Azure for some reason, the existing code will throw an error "Unrecognized hook 'beforeDinner'", whereas the suggested code would throw a null ref exception because "registerBeforeDinnerHook" doesn't exist on the worker api. I think the first error is better "barebones"/"unopinionated" behavior and I think null ref errors should be avoided.

src/WorkerChannel.ts
```diff
@@ -33,32 +29,32 @@ export class WorkerChannel {
...
+            case 'postInvocation':
```
ejizba, line 65: ## enums

TypeScript has great support for [string literal types](https://www.typescriptlang.org/docs/handbook/literal-types.html) (`'a' | 'b'`) and I frankly never use enums. At least in TypeScript, enums are super clunky and can have weird behavior if the first value of the enum is `0` because users so often do `if (val) {`. I recommend you try out string literal types (if you haven't already) and see if you can cause a bug without TypeScript throwing an error first. But yes `string` by itself is not type safe and I agree with that

src/WorkerChannel.ts
```diff
@@ -33,32 +29,32 @@ export class WorkerChannel {
...
+                return this._postInvocationHooks;
+            default:
```
ejizba, line 67: ## this instance

For this particular instance, yes I'm using string which would not be great on its own. _However_, the TypeScript types for "@azure/functions-worker" _are_ using string literals which helps a lot with intellisense. Specifically, the types package has stuff like this:
```
export function registerHook(
    hookName: 'postInvocation',
    callback: (context: PostInvocationContext) => void | Promise<void>
): Disposable;
```
I also have many unit tests, so a random typo here is unlikely to cause a regression in the future.

The reason I chose `string` here is because users _can_ pass any string at runtime. There is no guarantee it'll just be the hook names I currently have listed.

tsconfig.json
ejizba: We _have_ to provide a worker package that exposes any functionality the worker provides to the user like `registerHook`. The only question is what to name it and how "fancy" it should be. My proposal, as discussed elsewhere, is for all the fancy stuff to be in the separate programming model package. However, the worker package still needs to be exposed and the programming model package would directly reference it. End users _could_ reference the worker package, but the targeted audience of the worker package is just programming model package authors.

ejizba (commented):
Thanks Hossam! Most of your comments are topics I've debated myself, so I'm glad to be ironing them out with someone else now 🙂

workerTypes/index.d.ts
hossam-nasr: Right, that makes a lot of sense now. I misunderstood what was happening. I think in that case, maybe renaming this one might make it clearer? Maybe just `InvocationContext`? To me at least, `preInvocationContext` implied it would *only* be passed to pre-invocation hooks, not both.

In terms of comments, I guess the only thing would be updating the post invocation comment.

workerTypes/index.d.ts
hossam-nasr: Ignore this, other comments have clarified my understanding and I now notice the `extends`

workerTypes/index.d.ts
hossam-nasr: Right, that's a lot clearer now, thank you :)

src/WorkerChannel.ts
hossam-nasr: Yeah totally agree with this. That's what I meant with a "custom enum-style type," similar to what we use for `HttpMethod` for example. I just didn't know the technical term was string literal types lol

tsconfig.json
hossam-nasr: Yeah that makes sense. So is the plan that later in the future, we would also expose these hooks in the "fancier" programming model package, perhaps adding some fancier methods and syntactic sugar on top of the ones here, and have that package use this API here? Because to me it feels like these pre and post invocation hooks _are_ targeted towards more than just the programming model package authors.

src/setupWorkerModule.ts
hossam-nasr: Those reasons make a lot of sense to me.

Do we have a name in mind for this "fancier" programming model package? I agree that the best name for it would be `@azure/functions`, but then we would still run into the first problem that you described. This might be a very premature discussion, but I think just having some temporary tentative name could help clear out some confusion here.

Let's say the programming model package becomes `@azure/functions` while the types package is `@types/azure-functions`. If the intended audience of `@azure/functions-worker` is mainly the authors of `@azure/functions`, then I'm on board with that name :) My only argument would be that we should also expose these hooks in `@azure/functions`.

tsconfig.json
ejizba: Yes that is correct. Importantly though, nothing would prevent us from adding the "fancier methods" to this worker package if we end up deciding we don't want a separate programming model package.

src/setupWorkerModule.ts
ejizba: My preferred name for the "programming model package" at the moment is `@azure/functions`.

> Let's say the programming model package becomes @azure/functions while the types package is @types/azure-functions

We wouldn't need a separate types package for `@azure/functions`. Most npm modules ship the types as a part of their main module and we would do the same. We only need a separate types package for the worker package because it's built-in and not shipped with the user's app.

> My only argument would be that we should also expose these hooks in @azure/functions.

Yes I view this as a requirement for both packages

Another thing I haven't mentioned is that it would be super easy to support _another_ name for the worker package here since it's just an `if` statement. If we decide the worker package needs to be `@azure/functions` down the line, all it would take is `if (argArray[0] === '@azure/functions-worker' || argArray[0] === '@azure/functions') {`

src/WorkerChannel.ts
hossam-nasr: So is there a benefit to having this line?

```TS
export function registerHook(hookName: string, callback: HookCallback): Disposable;
```

At least when using TS, leaving that part out would flag an error right away when using an unidentified hook, rather than on runtime.

src/setupWorkerModule.ts
hossam-nasr: Yeah my preferred name would be `@azure/functions` as well. This comment here

> 1. That name is currently used for our types-only package. I think users would be confused if they're using something other than a type (an actual function `registerHook`) from that package all of the sudden

implied to me that we want to reserve `@azure/functions` for types only. But my preference would be what you described anyway!

> Another thing I haven't mentioned is that it would be super easy to support _another_ name for the worker package here since it's just an `if` statement. If we decide the worker package needs to be `@azure/functions` down the line, all it would take is `if (argArray[0] === '@azure/functions-worker' || argArray[0] === '@azure/functions') {`

Great point :)

workerTypes/index.d.ts
ejizba: Refactored hook contexts to be more intuitive

workerTypes/index.d.ts
ejizba: Refactored hook contexts to be more intuitive

workerTypes/index.d.ts
ejizba: Added separate types for each callback

src/WorkerChannel.ts
ejizba: I added the separate types, but I ran into too many build/lint errors if I tried to reference them here. If you can come up with a simple/clean way to do that let me know, otherwise I don't think this is worth the effort

src/WorkerChannel.ts
ejizba: I can't think of a benefit if the TypeScript types and worker are always exactly in sync. _However_, I think the team's past history with typescript types shows how difficult it is to keep them in sync 😝 With a generic function like this, users can pass in a hookName that is supported by the worker even if the types package they're on doesn't have it yet.

I would describe this line as the "safe, flexible, easily backward compatible" option whereas the other lines are much more strict. There's benefits to both, but for the worker package I think it makes more sense to leans towards the former

src/Disposable.ts
```diff
@@ -0,0 +1,35 @@
...
+                    if (disposable && typeof disposable.dispose === 'function') {
+                        disposable.dispose();
+                    }
+                }
+                disposables = undefined;
+            }
+        });
+    }
+
+    #callOnDispose?: () => any;
```
alrod, line 23: I am curios why `#` prefix is used?

workerTypes/index.d.ts
```diff
@@ -0,0 +1,101 @@
...
+    export type HookData = { [key: string]: any };
+
+    /**
+     * Base interface for all hook context objects
+     */
+    export interface HookContext {
+        /**
+         * The recommended place to share data between hooks
+         */
+        hookData: HookData;
```
alrod: we do not want `invocationContext: Context;` in the `HookContext` despite children of the interface have it, right?

src/Disposable.ts
ejizba: The `#` prefix is used for native JavaScript private fields, a pretty recent feature (Node.js v12+). I actually want to switch the whole repo over to use that. Issue and more info here: https://github.com/Azure/azure-functions-nodejs-worker/issues/527

workerTypes/index.d.ts
ejizba: Correct, because eventually we will likely have hooks unrelated to invocation (like worker startup/teardown tracked by https://github.com/Azure/azure-functions-nodejs-worker/issues/539)

src/setupWorkerModule.ts
```diff
@@ -0,0 +1,24 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License.
+
+import { HookCallback } from '@azure/functions-worker';
+import { Disposable } from './Disposable';
+import { WorkerChannel } from './WorkerChannel';
+import Module = require('module');
+
+export function setupWorkerModule(channel: WorkerChannel): void {
```
alrod: Is this code needed to provide `registerHook` from build-in `@azure/functions-worker` without specifying `@azure/functions-worker` in the function app `package.json`?

when `require('@azure/functions-worker')` will be called? in sample usage there is:
`import { registerHook } from '@azure/functions-worker';`


Can you please add a comment to describe what the function is doing.

ejizba: > Is this code needed to provide registerHook from build-in @azure/functions-worker without specifying @azure/functions-worker in the function app package.json?

Correct

> when require('@azure/functions-worker') will be called? in sample usage there is:
import { registerHook } from '@azure/functions-worker';

The sample usage is in typescript. When you compile `import` to javascript, it gets converted to `require`

> Can you please add a comment to describe what the function is doing.

Done

ejizba:
Met with the app insights team today. Overall looks good to them, but we need to add the function callback to the hooks api somehow. Filed https://github.com/Azure/azure-functions-nodejs-worker/issues/553 to address separately

alrod (approved):

src/WorkerChannel.ts
hossam-nasr: Yeah I gave it a try but I couldn't get them to work either, without having to use different functions to retrieve/execute each hook. I guess it's another argument for having separate functions 😋

src/WorkerChannel.ts
hossam-nasr: This makes a lot of sense, thanks!

src/WorkerChannel.ts
hossam-nasr: Makes sense, thanks!

src/setupWorkerModule.ts
hossam-nasr: A follow-up question to this: does this mean `@azure/functions-worker` can't be used with ESM?

src/setupCoreModule.ts
```diff
@@ -0,0 +1,29 @@
...
+import { WorkerChannel } from './WorkerChannel';
+import Module = require('module');
+
+/**
+ * Intercepts the default "require" method so that we can provide our own "built-in" module
+ * This module is essentially the publicly accessible API for our worker
+ * This module is available to users only at runtime, not as an installable npm package
+ */
+export function setupCoreModule(channel: WorkerChannel): void {
+    const workerApi = {
```
hossam-nasr: Maybe rename this to `coreApi` too?

hossam-nasr (approved):

src/setupWorkerModule.ts
ejizba: It looks a little different, but yes it can be used with ESM

```
import { createRequire } from "module";
const require = createRequire(import.meta.url);
const { registerHook } = require('@azure/functions-core');

registerHook('preInvocation', async (context) => {
    context.invocationContext.log('HELLO FROM PRE INVOCATION mjs');
});
```

This is the standard pattern for using a `require` method within ESM

ejizba: You can't use a simple `import`, but frankly I don't think that problem is unique to us. When I originally looked into this, the closest I could find to a solution is a preview "loaders" feature which may or may not help us in the future:
- https://github.com/nodejs/loaders
- https://nodejs.org/api/esm.html#loaders

I think there's plenty of oddities with Node's switch to ESM that are still being ironed out, so I think people using ESM should be fine with a few extra lines of code as shown above.

hossam-nasr: Makes sense, thanks for looking into it!

PR #549, opened 2022-03-04
Add deprecation flag for context.done()

hossam-nasr:
Adds a `@deprecated` flag for `context.done()`, and adds comments to remove deprecation linting errors.

Question: is this considered a breaking change? If a customer was also using the deprecation linting flag, this would break previously working code.

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -76,6 +76,9 @@ declare module '@azure/functions' {
          * @param err A user-defined error to pass back to the runtime. If present, your function execution will fail.
          * @param result An object containing output binding data. `result` will be passed to JSON.stringify unless it is
          *  a string, Buffer, ArrayBufferView, or number.
+         *
+         * @deprecated Use of sync functions with `context.done()` is deprecated. Use async/await and pass the response as the return value instead.
+         * See the docs here for more information: https://aka.ms/functions-js-async-await
          */
         done(err?: Error | string | null, result?: any): void;
         /**
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -77,7 +77,7 @@ declare module '@azure/functions' {
          * @param result An object containing output binding data. `result` will be passed to JSON.stringify unless it is
          *  a string, Buffer, ArrayBufferView, or number.
          *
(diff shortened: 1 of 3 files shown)


src/eventHandlers/invocationRequest.ts
```diff
@@ -119,9 +119,11 @@ export function invocationRequest(channel: WorkerChannel, requestId: string, msg
         if (result && typeof result.then === 'function') {
             result
                 .then((result) => {
+                    //eslint-disable-next-line deprecation/deprecation
```
ejizba: Any chance we can get https://github.com/Azure/azure-functions-nodejs-worker/pull/547 merged first? This code is completely changing

types/index.d.ts
```diff
@@ -76,6 +76,9 @@ declare module '@azure/functions' {
          * @param err A user-defined error to pass back to the runtime. If present, your function execution will fail.
          * @param result An object containing output binding data. `result` will be passed to JSON.stringify unless it is
          *  a string, Buffer, ArrayBufferView, or number.
+         *
+         * @deprecated Use of sync functions with `context.done()` is not recommended. Use async/await and pass the response as the return value instead.
```
ejizba, line 80: > Question: is this considered a breaking change? If a customer was also using the deprecation linting flag, this would break previously working code.

No, the point of deprecating something is to give users a heads up _before_ you make a breaking change. We currently have no intention of removing `context.done`, but marking it as deprecated now gives users _plenty_ of time to change their code in case we ever decide to remove it.

Also, lint rules are purely optional and we don't consider them for breaking changes.

src/eventHandlers/invocationRequest.ts
hossam-nasr: Sure, I'm happy to wait on merging this PR until #547 gets merged. I tried to review #547 but honestly I don't super follow what's going on, so I'd rather defer until @alrod can take a look.

types/index.d.ts
hossam-nasr: Great, thank you!

ejizba (approved):

Issue #449, opened 2021-10-28
Node process is making calls to Azure Files for package.json during cold start


safihamid:
Cold Start profiles show calls directly to Azure files to try to read package.json file during specialization.
This happens even when content is deployed az zip and we are using zip cache.

Looks like node process looks at all these places to try to load the package.json file:

<HTML>
<BODY>
<TABLE><TR><TD>FileName</TD></TR><TR><TD>C:\Program Files (x86)\SiteExtensions\Functions\4.0.0\workers\node\dist\src\package.json</TD></TR><TR><TD>C:\Program Files (x86)\SiteExtensions\Functions\4.0.0\workers\node\dist\package.json</TD></TR><TR><TD>C:\Program Files (x86)\SiteExtensions\Functions\4.0.0\workers\node\package.json</TD></TR><TR><TD>C:\Program Files (x86)\SiteExtensions\Functions\4.0.0\workers\package.json</TD></TR><TR><TD>C:\Program Files (x86)\SiteExtensions\Functions\4.0.0\package.json</TD></TR><TR><TD>C:\Program Files (x86)\SiteExtensions\Functions\package.json</TD></TR><TR><TD>C:\Program Files (x86)\SiteExtensions\package.json</TD></TR><TR><TD>C:\Program Files (x86)\package.json</TD></TR><TR><TD>C:\package.json</TD></TR><TR><TD>\\somestorage.file.core.windows.net\someappname\site\package.json</TD></TR><TR><TD>\\somestorage.file.core.windows.net\someappname\package.json</TD></TR></TABLE>

</BODY>
</HTML>

alrod:
I am wondering if node tries to find package.json on the function code load:
https://github.com/Azure/azure-functions-nodejs-worker/blob/v2.x/src/FunctionLoader.ts#L47

anthonychu:
I think this is standard behavior for Node. It uses package.json to figure out things like module system type (commonjs or module) and it'll search all parents to find it. It's searching from that location because workers\node\dist\srcnodejsWorker.js is the entry point. Not sure if we can prevent this but can probably limit it to a single read if we include a package.json in workers\node\dist\src.

alrod:
To prove the theory I did not load function from file system but instead loaded hardcoded function from memory. It looks like we need some "empty" package.json added to the worker so it will be loaded from worker folder before going to AzureFiles:
[...]

PR #552, opened 2022-03-05
Respect "type" property of package.json file

### LINKED ISSUE #531: Respect package.json type: module for ESM Functions
The preview support for ESM (ES6) modules available (#104 for tracking) requires the use of `mjs` file extensions (https://github.com/Azure/azure-functions-nodejs-worker/blob/v3.x/src/FunctionLoader.ts#L31 for source reference) which is a bit of a pain when it comes to working with TypeScript, [you need to use the nightly compiler](https://www.typescriptlang.org/docs/handbook/esm-node.html) to output a `mjs` extension.

It would be good if the worker would respect the [`type` field in package.json](https://nodejs.org/docs/latest-v16.x/api/packages.html#type) and if that is set to `module` it loads them as ESM rather than commonjs.

hossam-nasr:
Addresses #531. 

1. Adds ability to read package.json file during `initWorkerRequest`
2. Adds logic to respect "type" property in package.json and adds support for `.cjs` extension in `functionLoadRequest`.


src/utils/getPackageJson.ts
```diff
@@ -0,0 +1,18 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License.
+
+import path = require('path');
+import { access } from 'fs';
+import { promisify } from 'util';
+
+const accessAsync = promisify(access);
+
+export async function getPackageJson(appDir: string): Promise<Object> {
```
ejizba: We should avoid `object` and `any` in almost all cases because it's not type safe. Please add an interface like `PackageJson` and return that instead

src/utils/getPackageJson.ts
```diff
@@ -0,0 +1,18 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License.
+
+import path = require('path');
+import { access } from 'fs';
+import { promisify } from 'util';
+
+const accessAsync = promisify(access);
+
```
ejizba: It's a constant battle to keep the "utils" folder from growing out of control because any code _could_ be labeled a util. For that reason, I have some strict rules about that folder. One is that code in the "utils" folder has to be used across several files. Since this is only used by `WorkerChannel`, please put it directly in that file

src/eventHandlers/workerInitRequest.ts
```diff
@@ -47,6 +47,9 @@ export function workerInitRequest(channel: WorkerChannel, requestId: string, _ms
     }
 
     logColdStartWarning(channel);
+    if (_msg.functionAppDirectory) {
```
ejizba: nit: Rename to `msg`. We only use `_` to prefix unused variables

src/WorkerChannel.ts
```diff
@@ -61,4 +66,11 @@ export class WorkerChannel {
             after(context);
         }
     }
+
+    public async initAppDir(dir: string) {
```
ejizba: I think this needs to be called from `functionEnvironmentReloadRequest` as well, since that can affect `functionAppDirectory`

src/WorkerChannel.ts
```diff
@@ -61,4 +66,11 @@ export class WorkerChannel {
             after(context);
         }
     }
+
+    public async initAppDir(dir: string) {
+        if (dir !== this.functionAppDir) {
```
ejizba: In the case of `functionEnvironmentReloadRequest`, I feel like we would want to re-read the package.json file regardless of if the dir changed, just in case the file contents changed

src/FunctionLoader.ts
```diff
@@ -20,13 +20,13 @@ export class FunctionLoader implements IFunctionLoader {
         };
     } = {};
 
-    async load(functionId: string, metadata: rpc.IRpcFunctionMetadata): Promise<void> {
+    async load(functionId: string, metadata: rpc.IRpcFunctionMetadata, packageJson: Object = {}): Promise<void> {
```
ejizba: I assume the only reason you made it optional was to make the test code easier, but test code should not affect src code like this. It would be a bug if we forgot to pass `packageJson` to `load` in src code, so the parameter should be required

src/FunctionLoader.ts
```diff
@@ -70,6 +70,19 @@ export class FunctionLoader implements IFunctionLoader {
...
+
+    isUsingMjs(filePath: string, packageJson: Object): Boolean {
+        if (filePath.endsWith('.mjs')) {
+            return true;
+        }
+        if (filePath.endsWith('.cjs')) {
+            return false;
+        }
+        if (packageJson['type'] && packageJson['type'] === 'module') {
+            return true;
```
ejizba: nit: How about we rename this to `isESModule`? This case returns true even though the file name isn't "mjs"

src/utils/getPackageJson.ts
```diff
@@ -0,0 +1,18 @@
...
+export async function getPackageJson(appDir: string): Promise<Object> {
+    const packageJsonPath = path.join(appDir, 'package.json');
+    try {
+        await accessAsync(packageJsonPath);
+        return require(packageJsonPath);
```
ejizba: I've always used [`fs-extra`](https://www.npmjs.com/package/fs-extra) for file system operations in my Node.js projects - it just makes a lot of things easier. This repo isn't using it yet, but I planned on adding it eventually. In this case, you would use `fse.readJson` instead of `access` and `require`

src/utils/getPackageJson.ts
hossam-nasr: I added that interface based on the package.json schema, but it might be a bit of an overkill now. Let me know what you think

src/utils/getPackageJson.ts
hossam-nasr: Addressed this. After using `fs-extra`, I think that simplified the code enough to the point where I don't think a helper function is even necessary

src/eventHandlers/workerInitRequest.ts
hossam-nasr: Done

src/WorkerChannel.ts
hossam-nasr: Done

src/WorkerChannel.ts
hossam-nasr: Done

src/FunctionLoader.ts
hossam-nasr: Busted haha, addressed this now

src/FunctionLoader.ts
hossam-nasr: Done

src/utils/getPackageJson.ts
hossam-nasr: Thanks! Looks like a pretty helpful library, better than promisifying everything :p

types/index.d.ts
```diff
@@ -526,4 +526,464 @@ declare module '@azure/functions' {
             lastUpdated: string;
         };
     }
+
+    /**
```
ejizba: Wooooaahhhh I just meant a simple internal interface haha. This file is considered "external" because users can access these type. Anything _not_ exported in this file is "internal"

And for our purposes, we only need
```
export interface PackageJson {
    type?: whatever
}
```

src/utils/getPackageJson.ts
ejizba: Wooooaahhhh yeah I just meant a simple internal interface haha (`types/index.d.ts` is considered "external" because users can access those types. Anything _not_ exported in `types/index.d.ts` is "internal". We don't export `WorkerChannel` or any types related to that)

And for our purposes, we just need the `type` property, not everything else

hossam-nasr: Ooops my bad haha, thanks! So is there a centralized place for the "internal" types/interfaces? Not sure where to export the PackageJson interface from since it's used in multiple files

ejizba: There is no centralized place - we have interfaces in many different files. The source of the object and the types for it should be in the same file. In this case that means put it in `WorkerChannel.ts` because that file parses and provides the object

hossam-nasr: Makes sense, thanks :)

package.json
```diff
@@ -7,26 +7,28 @@
...
+        "fs-extra": "^10.0.1",
         "long": "^4.0.0",
         "minimist": "^1.2.5",
+        "mock-fs": "^5.1.2",
         "uuid": "^8.3.0"
     },
     "devDependencies": {
-        "@types/mocha": "^2.2.48",
         "@types/chai": "^4.2.22",
         "@types/chai-as-promised": "^7.1.5",
```
ejizba, line 17: Please add "@types/fs-extra" as a dev dependency

package.json
```diff
@@ -7,26 +7,28 @@
     "dependencies": {
         "@grpc/grpc-js": "^1.2.7",
         "@grpc/proto-loader": "^0.6.4",
+        "fs-extra": "^10.0.1",
         "long": "^4.0.0",
         "minimist": "^1.2.5",
+        "mock-fs": "^5.1.2",
```
ejizba: This should be a dev dependency, not a dependency

src/FunctionLoader.ts
```diff
@@ -70,6 +71,19 @@ export class FunctionLoader implements IFunctionLoader {
             throw new InternalException(`Function code for '${functionId}' is not loaded and cannot be invoked.`);
         }
     }
+
+    isESModule(filePath: string, packageJson: PackageJson): Boolean {
```
ejizba: nit:

```suggestion
    isESModule(filePath: string, packageJson: PackageJson): boolean {
```

https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html

src/FunctionLoader.ts
```diff
@@ -70,6 +71,19 @@ export class FunctionLoader implements IFunctionLoader {
...
     }
+
+    isESModule(filePath: string, packageJson: PackageJson): Boolean {
+        if (filePath.endsWith('.mjs')) {
+            return true;
+        }
+        if (filePath.endsWith('.cjs')) {
+            return false;
+        }
+        if (packageJson['type'] && packageJson['type'] === 'module') {
```
ejizba: nit:

```suggestion
        if (packageJson.type === 'module') {
```

src/WorkerChannel.ts
```diff
@@ -2,22 +2,32 @@
...
 import { AzureFunctionsRpcMessages as rpc } from '../azure-functions-language-worker-protobuf/src/rpc';
 import { IFunctionLoader } from './FunctionLoader';
 import { IEventStream } from './GrpcClient';
+import path = require('path');
 
 type InvocationRequestBefore = (context: Context, userFn: Function) => Function;
 type InvocationRequestAfter = (context: Context) => void;
 
+export interface PackageJson {
+    type?: 'commonjs' | 'module';
```
ejizba:
```suggestion
    type?: string;
```

This interface needs to describe what a user _can_ do, not what they _should_ do. For example, does Node.js respect "MoDuLe"? If so, we probably should, too (aka `packageJson.type?.toLowerCase() === 'module'`)

src/WorkerChannel.ts
```diff
@@ -2,22 +2,32 @@
...
+}
+
 export class WorkerChannel {
     public eventStream: IEventStream;
     public functionLoader: IFunctionLoader;
+    public functionAppDir: string;
```
ejizba: `functionAppDir` is never used outside `initAppDir`. Please remove it as a property

src/WorkerChannel.ts
```diff
@@ -61,4 +71,13 @@ export class WorkerChannel {
             after(context);
         }
     }
+
+    public async initAppDir(dir: string) {
```
ejizba: How about `updatePackageJson`

To me "init" means it only happens once

test/eventHandlers/workerInitRequest.test.ts
```diff
@@ -84,4 +85,33 @@ describe('workerInitRequest', () => {
...
+        const appDir = 'appDir';
+        const expectedPackageJson = {
+            type: 'module',
+        };
+        mock({
+            [appDir]: {
+                'package.json': JSON.stringify(expectedPackageJson),
+            },
+        });
+        await channel.initAppDir(appDir);
```
ejizba: I prefer if this file does its tests with `stream.addTestMessage(initMessage);` instead of calling a method on `channel`. Feels more realistic

test/eventHandlers/workerInitRequest.test.ts
```diff
@@ -84,4 +85,33 @@ describe('workerInitRequest', () => {
...
+        const appDir = 'appDir';
+        mock({
+            [appDir]: {
+                'not-package-json': 'some content',
+            },
+        });
+        await channel.initAppDir(appDir);
+        expect(channel.functionAppDir).to.equal(appDir);
+        expect(channel.packageJson).to.be.empty;
+        mock.restore();
```
ejizba: Move this to `afterEach` above

test/eventHandlers/workerInitRequest.test.ts
```diff
@@ -84,4 +85,33 @@ describe('workerInitRequest', () => {
...
+                'package.json': JSON.stringify(expectedPackageJson),
+            },
+        });
+        await channel.initAppDir(appDir);
+        expect(channel.functionAppDir).to.equal(appDir);
+        expect(channel.packageJson).to.deep.equal(expectedPackageJson);
+        mock.restore();
+    });
+
+    it('loads empty package.json', async () => {
```
ejizba, line 117: Add another test with a poorly-formed package.json (aka not json-parseable) just to make sure it doesn't block the worker from initializing

src/FunctionLoader.ts
alrod: I think the method can be simplified to:
`return filePath.endsWith('.mjs') || packageJson.type === 'module';`

src/WorkerChannel.ts
```diff
@@ -61,4 +71,13 @@ export class WorkerChannel {
             after(context);
         }
     }
+
+    public async initAppDir(dir: string) {
+        this.functionAppDir = dir;
+        try {
+            this.packageJson = await readJson(path.join(this.functionAppDir, 'package.json'));
```
alrod: it looks like we will have an additional file system call during initialization and for consumption/EP plans we use azure storage file share which is pretty slow. I am wondering about cold start impact here.

src/FunctionLoader.ts
hossam-nasr: @alrod That wouldn't honor the case in which we want `.cjs` to override `packageJson.type === 'module'`.

src/WorkerChannel.ts
ejizba: I don't think its a huge concern here. For example in https://github.com/Azure/azure-functions-nodejs-worker/issues/449, my understanding is only the last two entries were problematic and this file read should be like the first 9, not the last 2. Specifically, I expect the path to be like `C:\home\site\wwwroot\package.json`

We already do a file read for every function script file [here](https://github.com/Azure/azure-functions-nodejs-worker/blob/v3.x/src/FunctionLoader.ts#L41), and since users can have many functions I certainly hope one file read isn't a big performance impact.

src/WorkerChannel.ts
aaronpowell: I don't think anything other than lowercase is _legal_ (see https://nodejs.org/docs/latest-v16.x/api/packages.html#type) but since it's JSON it's not exactly enforced.

src/eventHandlers/functionEnvironmentReloadRequest.ts
```diff
@@ -34,6 +34,7 @@ export function functionEnvironmentReloadRequest(
                 logCategory: LogCategory.System,
             });
             process.chdir(msg.functionAppDirectory);
+            await channel.updatePackageJson('.');
```
hossam-nasr: My tests revealed a bug in my earlier logic here. My ealier code was

```TS
await channel.updatePackageJson(msg.functionAppDirectory)
```

But because we're already `chdir`ing into the new `functionAppDirectory` in the line above, we would actually end up looking for `package.json` in `functionAppDirectory/functionAppDirectory/package.json`.

Need a second pair of eyes to validate my new logic here and if this makes sense.

hossam-nasr (commented):
Added tests to validate changing `package.json` in `functionEnvironmentReloadRequest`, which revealed a bug in my earlier code that I should have now fixed. 

Also added a `Msg` namespace in the `initWorkerRequest` tests to make tests easier, similar to `functionEnvironmentReloadRequest`.

test/eventHandlers/workerInitRequest.test.ts
hossam-nasr: Done

test/eventHandlers/workerInitRequest.test.ts
hossam-nasr: Done

test/eventHandlers/workerInitRequest.test.ts
hossam-nasr: I did that. I couldn't get it to work without also copying the response message from the first test into a call to `stream.assertCalledWith()` so I extracted that message into a `Msg` namespace, similar to the tests I saw for `functionEnvironmentReloadRequest`. I don't know if there's an easier way to write the tests without needing that message though.

src/WorkerChannel.ts
hossam-nasr: Done

src/WorkerChannel.ts
hossam-nasr: I figured it might become useful in the future, but I guess we could always add it if we need it. Removed it for now!

src/WorkerChannel.ts
hossam-nasr: I changed the interface to `string`. I kept the existing logic as-is though, not respecting anything other than `"module"` all lowercase.

src/FunctionLoader.ts
hossam-nasr: Thanks!

package.json
hossam-nasr: Good call lol thanks!

package.json
hossam-nasr: Done

src/eventHandlers/functionEnvironmentReloadRequest.ts
ejizba: The path passed as `functionAppDirectory` should always be an absolute path (`/Users/eric/appDir` as opposed to `./appDir`), so it should work regardless of the current process's directory. My preference is to revert to your previous logic so that the order of these lines doesn't matter.

I'm guessing it's just the tests that were doing a relative path

test/eventHandlers/workerInitRequest.test.ts
ejizba: Yes that is all by design and I like your changes. `stream.assertCalledWith` is supposed to validate the exact messages sent back and forth and so it's a good thing it failed without the response message

src/WorkerChannel.ts
```diff
@@ -61,4 +69,12 @@ export class WorkerChannel {
             after(context);
         }
     }
+
+    public async updatePackageJson(dir: string): Promise<void> {
+        try {
+            this.packageJson = await readJson(path.join(dir, 'package.json'));
+        } catch {
+            this.packageJson = {};
```
ejizba, line 95: Let's warn users when this happens. Similar to [this](https://github.com/Azure/azure-functions-nodejs-worker/blob/v3.x/src/eventHandlers/functionLoadRequest.ts#L22), except a warning

test/eventHandlers/workerInitRequest.test.ts
```diff
@@ -11,6 +12,44 @@ import { TestEventStream } from './TestEventStream';
 import LogCategory = rpc.RpcLog.RpcLogCategory;
 import LogLevel = rpc.RpcLog.Level;
 
+namespace Msg {
+    export function initMsg(functionAppDirectory?: string): rpc.IStreamingMessage {
```
ejizba: nit: You don't need "Msg" at the end of these names when they're in a "Msg" namespace

ejizba (commented):
Updated tests look great, thanks!

src/eventHandlers/functionEnvironmentReloadRequest.ts
hossam-nasr: Yeah makes sense. I returned this code to the original and changed the tests to use absolute paths too. Thanks!

src/WorkerChannel.ts
hossam-nasr: Done!

ejizba (approved):

PR #554, opened 2022-03-14
Bump .NET version for E2E tests on v2.x branch

ejizba:
Build agents don't have .NET 2.1 anymore (not surprising), so tests are currently failing.

cc @Francisco-Gamino this should fix the Node integration test failures


hossam-nasr (approved):
LGTM

PR #555, opened 2022-03-16
Use system log category for errors after done

ejizba:
Currently these logs are lost because they happen after execution completes and the host ignores "orphaned" user logs. Related to discussion in https://github.com/Azure/azure-functions-host/issues/8222

> User logs - Cx have access to these logs via App Insights - are super set for Function Execution logs + Host / Extension logs - Any logs coming from Host are considered as system logs and can be logged to Kusto. User logs cannot be logged to Kusto as they are considered PII

> If the error is coming from the worker itself, I would consider it a system log, especially if it doesn't include the user's actual function logs (which is the PII part).

### DIFF
--- a/src/eventHandlers/invocationRequest.ts
+++ b/src/eventHandlers/invocationRequest.ts
@@ -29,6 +29,7 @@ export async function invocationRequest(channel: WorkerChannel, requestId: strin
     let resultIsPromise = false;
 
     const info = channel.functionLoader.getInfo(nonNullProp(msg, 'functionId'));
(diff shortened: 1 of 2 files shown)


alrod:
@ejizba, did you have a chance to figure out why we emitted "LogCategory.User" at all? It looks like after the change we will not have "LogCategory.User" emitted from the worker code anymore.

ejizba:
We still mark any log coming from `context.log` as `LogCategory.User`. The user will reference `context.log` directly from their own code, which is why you don't see any references by us in this codebase

alrod (approved):

PR #563, opened 2022-03-28
Add worker version to core module api

ejizba:
I view `version` as one of the most basic pieces of information that should be on any api, including our built in module. I could imagine npm modules (like app insights or a new programming model package) using this to throw errors like "We only support worker version 3+".

Since this was yet another reference to our version that could easily get out of date, I also added a script to update/validate the various references to the version in this repo.


scripts/updateVersion.ts
```diff
@@ -0,0 +1,116 @@
...
+    if (!match) {
+        throw new Error(`Failed to find match for "${regex.source}".`);
+    }
+    return match[1];
+}
+
+function updateVersion(newVersion: string) {
+    updatePackageJsonVersion(repoRoot, newVersion);
+
+    if (newVersion.endsWith('.0')) {
```
hossam-nasr, line 88: What if `newVersion` is something like `4.0`?

hossam-nasr, line 88: What if `newVersion` is something like `4.0`?

ejizba: Well, `4.0` isn't valid semver https://semver.org/

hossam-nasr: Well, do we check that `newVersion` is a valid semver? 😛

Regardless, let's say, for whatever reason, we were upgrading from `3.0.0` to `3.1.1`, don't we still want to update the types package?

ejizba: I suppose we don't check locally, but `validateVersion` does make sure its valid semver, so the build would fail as soon as someone put up a PR.

Going from `3.0.0` to `3.1.1` is also an invalid semver pattern. Per the docs: "Patch version MUST be reset to 0 when minor version is incremented." If you have a simple suggestion to check that edge case I will add it, but otherwise I don't think it's worth the time for an engineering script. Hopefully if someone did that it would be caught in code review.

hossam-nasr: Fair enough. The only thing I had in mind was to get the old version first and compare with the new version, but I suppose that might not be worth it as you say :)

hossam-nasr (approved):

Issue #553, opened 2022-03-14
Add function callback to pre/post invocation hook context


ejizba:
Met with App Insights folks and the main thing missing from the new [hooks](https://github.com/Azure/azure-functions-nodejs-worker/pull/548) is the function callback. The callback is an argument to [`cls-hooked`](https://www.npmjs.com/package/cls-hooked) in their [wrapWithCorrelationContext](https://github.com/microsoft/ApplicationInsights-node.js#setting-up-auto-correlation-for-azure-functions) method.

Since we store function callbacks across executions, we should make sure the hooks for one execution can't affect another execution.

PR #564, opened 2022-03-29
Refactor FunctionLoader so that it returns a cloned function each time

ejizba:
Related to https://github.com/Azure/azure-functions-nodejs-worker/issues/553, people will be able to mess with the function more easily, so I want to make sure they can't affect the function across invocations.

Interestingly, we were already cloning the function in `getEntryPoint` as a part of setting the `this` arg. I just moved the logic so that it happens during `getFunc` instead of `load` and used the slightly cleaner `bind` instead of `apply`.

### DIFF
--- a/src/FunctionLoader.ts
+++ b/src/FunctionLoader.ts
@@ -18,6 +18,7 @@ export class FunctionLoader implements IFunctionLoader {
         [k: string]: {
             info: FunctionInfo;
             func: Function;
+            thisArg: unknown;
         };
     } = {};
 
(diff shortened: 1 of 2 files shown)


hossam-nasr (approved):
Seems like some magic to me with the `bind` business. How were we setting `this` appropriately before without using `bind`? Regardless, I don't want to block on this :)

ejizba:
### Before
```
return function () { return f.apply(obj, arguments); };
```

### After
```
return loadedFunction.func.bind(loadedFunction.thisArg);
```

If I were to use the old names, it would look like this:
```
return f.bind(obj);
```

The main benefit of `bind` is you can get rid of the wrapper `function() {}` stuff. Otherwise, I feel like `bind` and `apply` are both pretty magic-y so I don't think it makes much of a difference.

PR #570, opened 2022-04-06
Organize test code

ejizba:
The `test` folder structure should match the `src` folder structure and each test file should be named `<src file being tested>.test.ts`

I'm pretty sure this is a standard pattern - at a minimum my previous teams did this and I think the pattern makes more sense than what we currently have.


hossam-nasr (approved):
I like this pattern :)

PR #574, opened 2022-04-12
Add shared error logic for all event handlers

ejizba:
Currently each event handler does its own error handling and I worry errors will slip through the cracks and not get reported properly over the rpc channel. If I move the error logic out of the event handlers and into `setupEventStream`, it should be much safer and consistent.

Btw, definitely review this one with whitespace turned off


ejizba:
ping @hossam-nasr @alrod

Issue #569, opened 2022-04-05
Support defining functions in code instead of function.json


ejizba:
This work was unblocked in the host by https://github.com/Azure/azure-functions-host/pull/7606. For now, it requires users to include `EnableWorkerIndexing` in the `AzureWebJobsFeatureFlags` app setting.

This specific issue is to track basic support in the [core worker api](https://github.com/Azure/azure-functions-nodejs-worker/issues/567) for defining functions in code. I imagine this essentially just means users can copy/paste their current "function.json" content into a js/ts file. It should be a simple change to their codebase that will let them completely redefine their file structure. Anything fancier will be tracked by the new programming model package in https://github.com/Azure/azure-functions-nodejs-worker/issues/480

Prereqs:
- Provide api for the worker: https://github.com/Azure/azure-functions-nodejs-worker/issues/567
- Add entrypoint for app-level code: https://github.com/Azure/azure-functions-nodejs-worker/issues/537

ejizba:
I did a rough first commit on [this branch](https://github.com/Azure/azure-functions-nodejs-worker/tree/ej/indexing) that only works with "OptionA-Simple" in the prototype repo [here](https://github.com/ejizba/func-nodejs-prototype). Keep in mind you have to add this to the local.settings.json in the app you're testing:
`"AzureWebJobsFeatureFlags": "EnableWorkerIndexing"`

Plan as discussed today was to have @hossam-nasr take it from here, which includes:
1. Get the other prototype options to "work" - for the purposes of a user study and to validate the core api is on the right track
2. Make the rough commit an actual PR for merging

PR #576, opened 2022-04-15
Add support for app-level code

### LINKED ISSUE #537: Add entrypoint for app-level code
UPDATE: Per the final design, you can specify a single file or a glob pattern for multiple files as the "main" field in your "package.json" file.

---

Basically, provide a way for users to run app-level code when the worker is initialized. This would be directly beneficial for the new hooks feature (#522), but could realistically be used for any arbitrary code. Current proposal is to respect the [main](https://docs.npmjs.com/cli/v8/configuring-npm/package-json#main) field in a user's package.json

NOTE: I already added the function app dir to the worker init request in https://github.com/Azure/azure-functions-host/pull/8144, necessary to read the user's package.json

ejizba:
Users can specify an entry point file as the [main](https://docs.npmjs.com/cli/v8/configuring-npm/package-json#main) field in their package.json and we will load the file during worker initialization. The file can be used to run any arbitrary app-level code (as opposed to function-level code), but I expect the most common scenario is to register [hooks](https://github.com/Azure/azure-functions-nodejs-worker/issues/522) and eventually [functions](https://github.com/Azure/azure-functions-nodejs-worker/issues/569).

A few notes:
- I handle commonjs vs es6 the exact same way we already do it for function-level files
- The [npm docs](https://docs.npmjs.com/cli/v8/configuring-npm/package-json#main) say "If main is not set it defaults to index.js in the package's root folder.", but I worry users may have a random "index.js" field in the root of their app already because it's such a common name. Instead, I decided to make the feature opt-in only, meaning users have to explicitly add the main field if they want the entry point file to be loaded 

Fixes #537


src/eventHandlers/WorkerInitHandler.ts
```diff
@@ -25,6 +28,12 @@ export class WorkerInitHandler extends EventHandler<'workerInitRequest', 'worker
     async handleEvent(channel: WorkerChannel, msg: rpc.IWorkerInitRequest): Promise<rpc.IWorkerInitResponse> {
         const response = this.getDefaultResponse(msg);
 
+        channel.log({
+            message: 'Received WorkerInitRequest',
+            level: LogLevel.Debug,
+            logCategory: LogCategory.System,
+        });
```
hossam-nasr, line 35: Did we not log this before?

src/parsers/parsePackageJson.ts
```diff
@@ -0,0 +1,43 @@
...
+                delete data[field];
+            }
+        }
+        return data;
+    } catch (err) {
+        const error: Error = ensureErrorType(err);
+        if (error.name === 'SyntaxError') {
+            error.message = `file content is not valid JSON: ${error.message}`;
+        }
+        throw error;
```
hossam-nasr, line 41: Should we also set the `isAzureFunctionsInternalException` to true here? That property didn't exist when I wrote some of this code afaik

src/loadScriptFile.ts
```diff
@@ -0,0 +1,32 @@
...
+
+export function isESModule(filePath: string, packageJson: PackageJson): boolean {
+    if (filePath.endsWith('.mjs')) {
+        return true;
+    }
+    if (filePath.endsWith('.cjs')) {
+        return false;
+    }
+    return packageJson.type === 'module';
+}
```
hossam-nasr, line 32: Should this file be moved to utils? Per this discussion: https://github.com/Azure/azure-functions-nodejs-worker/pull/552#discussion_r821098511, it is being used in more than one files. Are there other rules for the utils folder?

src/eventHandlers/WorkerInitHandler.ts
```diff
@@ -54,8 +63,34 @@ export class WorkerInitHandler extends EventHandler<'workerInitRequest', 'worker
...
+                level: LogLevel.Debug,
+                logCategory: LogCategory.System,
+            });
+            try {
+                const entryPointFullPath = path.join(functionAppDirectory, entryPointFile);
+                if (!(await pathExists(entryPointFullPath))) {
+                    throw new Error(`file does not exist`);
+                }
+
+                await loadScriptFile(entryPointFullPath, channel.packageJson);
```
hossam-nasr, line 82: So my understanding here is that this code simply loads this entrypoint file, but it doesn't yet execute any code there?

src/parsers/parsePackageJson.ts
```diff
@@ -0,0 +1,43 @@
...
+            throw new Error('file content is not an object');
+        }
+
+        const stringFields = ['main', 'type'];
+        for (const field of stringFields) {
+            if (field in data && typeof data[field] !== 'string') {
+                // ignore fields with an unexpected type
+                delete data[field];
+            }
+        }
```
hossam-nasr, line 34: Is there a way to generalize this to instead ensure that all fields specified in the interface match their specified type? I'm worried if in the future, we add more properties to the interface but we fail to also add checks for them here

test/eventHandlers/WorkerInitHandler.test.ts
```diff
@@ -150,4 +211,72 @@ describe('WorkerInitHandler', () => {
         );
         expect(channel.packageJson).to.be.empty;
     });
+
+    for (const extension of ['.js', '.mjs']) {
```
hossam-nasr: Is there a reason `.cjs` isn't also there?

hossam-nasr (commented):
Looks good! Just left a few comments/questions

src/eventHandlers/WorkerInitHandler.ts
ejizba: We did not. I think that was fine before because the init logic was pretty simple, but now we're running user code so logs are more important because who knows what could happen

src/loadScriptFile.ts
ejizba: The comment you referenced is about how I have strict rules for keeping stuff _out_ of utils because that folder can easily become a mess (the rules are not used to put stuff _in_ utils). That means if a file already has a different place where it fits in the repo it should go there first. In this case, I think `loadScriptFile` is an important enough file that it can be in the root `src` folder. It has code that is very integral to how we run a user's code, the main point of the js worker.

src/parsers/parsePackageJson.ts
ejizba: That property has actually been there since before we both joined. I haven't fully figured out the nuances of when we should or shouldn't use it, but at some point in the future it would be nice to go through the repo and make sure we're using it consistently.

It seems like the property mostly affects if or how an error gets logged as a system log. You already added code in `channel.updatePackageJson` that logs this error as a warning, so I believe `isAzureFunctionsInternalException` would have no effect in this particular case:
```
this.log({
    message: `Worker failed to load package.json: ${error.message}`,
    level: LogLevel.Warning,
});
```

test/eventHandlers/WorkerInitHandler.test.ts
ejizba: Nope, I can add it

src/eventHandlers/WorkerInitHandler.ts
ejizba: Loading a file in javascript means you're executing all the code at the top level of the file. That usually means importing modules, defining a class/function/const, and in our case registering a hook. As an example, 'hello' would get printed but 'world' would not:
```
console.log('hello');

export function world() {
    console.log('world');
}
```

That being said, users should put any complex code in a startup hook (which is the issue you're about to work on) instead of at the top level of their code. There's a few reasons for this that I already listed in https://github.com/Azure/azure-functions-nodejs-worker/issues/539

src/parsers/parsePackageJson.ts
ejizba: I'm open to suggestions, but I'm guessing it'll be more complicated than it's worth. Mostly because I don't think we will use very many more properties from package.json at all, but slightly because I would hope someone adding a property to the *parse*PackageJson file would know they might need to add logic to _parse_ the property

src/eventHandlers/WorkerInitHandler.ts
hossam-nasr: Makes sense, thanks :)

src/eventHandlers/WorkerInitHandler.ts
hossam-nasr: Gotcha, that makes more sense now, thanks! 

Is "loading" standard terminology to refer to import/require? If not, renaming this to something like `executeScriptFile` would make it clearer to me what it's doing, but maybe that's just me :P

ejizba: Yeah I think loading is pretty standard. Its frowned upon to put complex code at the top level and complex code is the stuff you generally think of as "executing". If you take my examples from above, you could rephrase it to "loading" modules, "loading" classes/functions/consts and "loading" hooks. So while people _can_ execute complex stuff, they should ideally just be loading things

src/parsers/parsePackageJson.ts
hossam-nasr: I think you're right. I was hoping there was an easy way to do this but apparently not :p

hossam-nasr (approved):

PR #577, opened 2022-04-25
Add app-level startup hooks

hossam-nasr:
Partially implements #539. This PR builds on the work to add invocation hooks (#548) and to support app-level code (#576), to provide the ability to register app startup hooks, which are run in `workerInitHandler` (or `functionEnvironmentReloadHandler` in the specialization scenario).

In addition, some additions/decisions in this PR, which may require more discussion:

* An app teardown hook is not provided for now, since the Host does not yet support the `worker_terminate` signal for elegant shutdown of the worker. This issue is currently being worker on in the host (https://github.com/Azure/azure-functions-host/issues/2308 and https://github.com/Azure/azure-functions-host/pull/8385). Once this issue is implemented, a handler for the `worker_terminate` gRPC message can be added to add app teardown hooks.
* Added a `logger` property to `HookContext`, which allows all hooks, including invocation and app startup hooks, to emit user logs to app insights.
    - Pros:
         - Provides a clear and recommended way to emit logs to app insights from hooks, including app startup ones
         - Allows users to choose the log level, which will be honored in app insights
         - Includes extra metadata in the logs (such as function invocation ID for invocation hooks)
    - Cons:
         - Right now, for invocation hooks, users can already access the same functionality using `hookContext.invocationContext.log`. Duplicating properties is bad design.
         - If we include this property in the worker core API, it will be harder to change/remove in the future
         - Right now, logs emitted from app startup hooks will be ignored by the host, as they will be treated as "orphaned" logs, without a function invocation ID attached to them. Including this property before the Host can support logging from outside a function invocation, which will essentially do nothing, can cause confusion.
* The `AppStartupContext` provided to app startup hooks also includes a `hookData` property. That `hookData` property is _copied_ to invocation hooks' contexts' `hookData` property, but cannot be directly modified by each invocation.
    - Pros:
        - This `hookData` property can in the future be shared between app-level hooks, such as app teardown hook.
        - Allows sharing of data between app-level and invocation-level hooks
    - Cons:
        - If the data is merely copied to each invocation hook, maybe this isn't necessary, and users can just use global variables instead.


src/WorkerChannel.ts
```diff
@@ -74,18 +82,60 @@ export class WorkerChannel {
...
+            verbose: (...args: any[]) => log(LogLevel.Trace, ...args),
+        });
+
+        return {
+            hookData: this.#hookData,
+            logger,
+        };
+    }
+
+    #userLog(level: LogLevel, functionInvocationId?: string, msgCategory?: string, ...args: any[]): void {
```
ejizba: > Added a logger property to HookContext, which allows all hooks, including invocation and app startup hooks, to emit user logs to app insights.

To me, the cons outweigh the pros on this one. I think https://github.com/Azure/azure-functions-host/issues/8222 is a blocker and I want to keep the core api as simple as possible until we make more progress on the programming model. We can have a discussion on this and perhaps you can convince me otherwise, but I would prefer if we do that in a separate PR as I think this would be a feature on its own

src/eventHandlers/FunctionEnvironmentReloadHandler.ts
```diff
@@ -43,7 +44,7 @@ export class FunctionEnvironmentReloadHandler extends EventHandler<
                 logCategory: LogCategory.System,
             });
             process.chdir(msg.functionAppDirectory);
-            await channel.updatePackageJson(msg.functionAppDirectory);
+            await channel.initalizeApp(msg.functionAppDirectory);
```
ejizba: It looks like `initalizeApp` is the method called from both event handlers so I would create a `initalizeApp.ts` file and put any function it calls from `WorkerChannel.ts` (i.e. `loadEntryPointFile`) also into that file. Then you can create a test file `initalizeApp.test.ts` to test the feature. 

nit: check the spelling of `initalizeApp` and also thoughts on `initialize` vs `startup`? Wondering if we can merge terminology somehow

src/WorkerChannel.ts
hossam-nasr: Yeah makes sense. I've removed it for now

src/eventHandlers/FunctionEnvironmentReloadHandler.ts
hossam-nasr: I refactored to a new file "appStartup"

src/eventHandlers/WorkerInitHandler.ts
```diff
@@ -34,35 +33,12 @@ export class WorkerInitHandler extends EventHandler<'workerInitRequest', 'worker
 
         logColdStartWarning(channel);
         const functionAppDirectory = nonNullProp(msg, 'functionAppDirectory');
```
ejizba: Were you able to test specialization locally? What ended up being the value of `functionAppDirectory` when starting in placeholder mode? `nonNullProp` will throw an error and so I was thinking we might want this to be `if (msg.functionAppDirectory) { ...` instead of `nonNullProp` just to be safe

types-core/index.d.ts
```diff
@@ -18,11 +18,13 @@ declare module '@azure/functions-core' {
      */
     export function registerHook(hookName: 'preInvocation', callback: PreInvocationCallback): Disposable;
     export function registerHook(hookName: 'postInvocation', callback: PostInvocationCallback): Disposable;
+    export function registerHook(hookName: 'appStartup', callback: AppStartupCallback): Disposable;
```
ejizba: Let's be extra careful on the name 'appStartup' because it'll be hard to change moving forward. cc @lilyjma

@hossam-nasr can you look for other examples to base this name off of? For example I think I got 'preInvocation' and 'postInvocation' from the python team and 'invocation' was pretty widely used for Functions already. Unfortunately, I'm not sure "startup" is common in functions and I don't think the python team has this hook. Fwiw, "startup" originally came from @anthonychu here: https://github.com/Azure/azure-functions-nodejs-worker/issues/522#issuecomment-1034609167. You could look at [fastify](https://www.fastify.io/docs/latest/Reference/Hooks/) or other frameworks for examples. Consider both the name "startup" and if "app" should be a prefix

src/appStartup.ts
```diff
@@ -0,0 +1,52 @@
...
+import { pathExists } from 'fs-extra';
+import { AzureFunctionsRpcMessages as rpc } from '../azure-functions-language-worker-protobuf/src/rpc';
+import { loadScriptFile } from './loadScriptFile';
+import { ensureErrorType } from './utils/ensureErrorType';
+import { WorkerChannel } from './WorkerChannel';
+import path = require('path');
+import LogLevel = rpc.RpcLog.Level;
+import LogCategory = rpc.RpcLog.RpcLogCategory;
+
+export async function appStartup(functionAppDirectory: string, channel: WorkerChannel): Promise<void> {
```
ejizba: I think we need to document the worker specialization stuff somewhere because it was pretty confusing. How about something like this attached to this function
```
/**
 * App startup can happen in two places, depending on if the worker was specialized or not
 * 1. The worker can start in "normal" mode, meaning `workerInitRequest` will reference the user's app
 * 2. The worker can start in "placeholder" mode, meaning `workerInitRequest` will reference a dummy app to "warm up" the worker and `functionEnvironmentReloadRequest` will be sent with the user's actual app.
 *    This process is called worker specialization and it helps with cold start times.
 *    The dummy app should never have actual startup code, so it should be safe to call `appStartup` twice in this case
 */
```

src/eventHandlers/InvocationHandler.ts
```diff
@@ -90,16 +90,21 @@ export class InvocationHandler extends EventHandler<'invocationRequest', 'invoca
                 });
             });
 
-            const hookData: HookData = {};
             let userFunction = channel.functionLoader.getFunc(functionId);
+
+            // create a copy of the hook data in the worker context (set by app hooks)
```
ejizba: I like the idea of providing the app-level data to invocation-level hooks, but I'm not so sure about copying like this. For example, the `...` operator is just a shallow copy (let me know if you need me to explain "shallow copy"), which may lead to unexpected behavior.

What if we add a property `appHookData` to `PreInvocationContext` and `PostInvocationContext`. Aka we don't do any copying and it's just a separate object. We could warn users that changes to the data will be app-wide and it's up to them if they want to create a copy

hossam-nasr: Oh I didn't know `...` only did a shallow copy, thank you for pointing that out. We could do a deep copy to get around that issue.

Alternatively, I do like the idea of a separate `appHookData` for invocation contexts. It also has the added benefit of allowing different invocations to speak to each other, which is dangerous but could be useful. In that case, would it still be the `hookData` property for `appStartupContext`? Or do we still want to keep some data property that is only shared by app-level hooks and not accessible by individual invocations?

src/eventHandlers/WorkerInitHandler.ts
hossam-nasr: I did not test specialization locally, but judging from host code [here](https://github.com/Azure/azure-functions-host/blob/1db1782f1bda1718215ce7337e89ebde32515325/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs#L232) and @pragnagopa 's comments on the thread we had about worker specialization: "In PlaceholderMode,  Host starts with dummy app content , script path points to a local directory that has a .csx function," I doubt that this value should be `undefined`/`null` in normal operations

src/eventHandlers/InvocationHandler.ts
ejizba: I think appStartupContext should have an `appHookData` property that is the one shared with invocation hook contexts. I'm just not sure if appStartupContext should _also_ have a `hookData` property for just app-level hooks. It might be a little confusing, but I guess it can't hurt to have both

types-core/index.d.ts
hossam-nasr: I look at fastify, and it looks like their hooks generally follow the `onEvent` pattern, which we don't do for the invocation hooks. It looks like their most equivalent hook is called `onReady`. I asked the Python team and you're right that they don't have this hook. They have an app-level level, but that is essentially an invocation hook that is by default run for _all_ functions, as opposed to their function-level hooks, which are specific to a particular function.

Other potential nouns we could use instead of "startup":
- start
- load
- init/initialize
- ready

Could definitely use suggestions from you or @lilyjma though.

I don't have a strong preference for keeping the "app" prefix by name, but I would like to have some way of telling that this hook will be run only once the function _app_ starts, not the function (i.e. a pre-invocation hook).

src/appStartup.ts
hossam-nasr: Done, and I also added a note about worker specialization happening only once, so we don't need about cleaning up resources

src/eventHandlers/InvocationHandler.ts
hossam-nasr: Yeah I agree it can be very confusing if we have both. We might need to think about how to name them to make sure the distinction is clear.

hossam-nasr: Thinking about it, what if we make the `hookData` property the one that is shared between all hooks, including app startup and different invocations (implied more by the name, and avoids different behavior for the same-named property in app vs invocation hooks), and we add a new property `invocationHookData` that is tied to one invocation?

src/eventHandlers/WorkerInitHandler.ts
ejizba: I still think it's better to be safe than sorry here. There's no downside to doing `if (msg.functionAppDirectory) {` instead of `nonNullProp` and it will give me peace of mind since we can't directly test a lot of worker specialization until it rolls out to Azure.

Also - as discussed offline, I think we need to test windows specialization locally. I'm okay if we merge the PR without testing, but I want at least one of us to test before releasing the worker with this change.

src/eventHandlers/InvocationHandler.ts
ejizba: Discussed this offline. `hookData` should be used only for hooks at the same scope (app vs invocation) and `appHookData` should be used to share hook data across scopes for all hooks. We can change these names in the programming model package once that's ready

types-core/index.d.ts
ejizba: I vote for `appStart` or `appInit`. Some reasoning:

### don't like
- load: This isn't broad enough - to me we also "execute" code and "load" doesn't include that
- initialize: word just seems too long
- ready: It just doesn't feel clear to me when an app is considered "ready"
- startup: I don't like words where you're not sure if it's one word or two words (startup vs startUp vs start-up) and IMO english has too many nuances to the words "up" vs "down" which make it confusing for non-native speakers

### do like
- app prefix: I agree it's nice to have a way to clarify this is for the app not the function
- init: This matches the rpc terminology (WorkerInitRequest) and feels accurate. Only downside is I'm not a huge fan of abbreviations by default, but I think most people know "init"
- start: feels simple, short, and obvious in the meaning. Downside is I don't have an example to point to saying "we're matching this"

src/appStartup.ts
```diff
@@ -0,0 +1,61 @@
...
+
+/**
+ * App startup can happen in two places, depending on if the worker was specialized or not
+ * 1. The worker can start in "normal" mode, meaning `workerInitRequest` will reference the user's app
+ * 2. The worker can start in "placeholder" mode, meaning `workerInitRequest` will reference a dummy app to "warm up" the worker and `functionEnvironmentReloadRequest` will be sent with the user's actual app.
+ *    This process is called worker specialization and it helps with cold start times.
+ *    The dummy app should never have actual startup code, so it should be safe to call `appStartup` twice in this case
+ *    Worker specialization happens only once, so we don't need to worry about cleaning up resources from previous `functionEnvironmentReloadRequest`s.
+ */
+export async function appStartup(functionAppDirectory: string, channel: WorkerChannel): Promise<void> {
```
ejizba: nit: It's common naming convention that a function in TypeScript should always start with a verb. Meaning this should be `startupApp`, `startApp`, or `initApp`.

test/appStartup.test.ts
```diff
@@ -0,0 +1,160 @@
...
+            Msg.executingHooksLog(1, 'appStartup'),
+            Msg.executedHooksLog('appStartup'),
+            EnvReloadMsg.reloadSuccess
+        );
+
+        expect(startupFunc.callCount).to.be.equal(1);
+        expect(startupFunc.args[0][0]).to.deep.equal(expectedStartupContext);
+    });
+
+    it('persists hookData changes from app startup hooks in worker channel', async () => {
```
ejizba: worker channel is an internal detail, and there's no reason we have to persist hook data on it long term. If you want to include this test it should check that appHookData is persisted for an invocation hook, which we do need to support long term

test/appStartup.test.ts
```diff
@@ -0,0 +1,160 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License.
+
+import * as coreTypes from '@azure/functions-core';
```
ejizba: Overall tests look good! 🎉 It's a bit confusing which tests go in appStartup.test.ts vs workerInitHandler.test.ts vs functionEnvironmentReloadHandler.test.ts, but I don't have a better suggestion. Perhaps something we improve in the future

ejizba (commented):
I think we already discussed all of this offline, but just documenting what I think is left for the PR

src/eventHandlers/WorkerInitHandler.ts
hossam-nasr: Should we at least log some kind of warning in the case where `msg.functionAppDirectory` is undefined? Because that means our app startup logic will not run on a `workerInitRequest` and that might lead to some unexpected behavior, and perhaps help with servicing too

types-core/index.d.ts
hossam-nasr: I agree with your reasoning. I ended up going with `appStart`. No specific _strong_ reason, but I didn't want `appInit` to be confused with the `workerInitRequest`, since these hooks will be called in also the `functionEnvironmentReloadRequest`.

src/appStartup.ts
hossam-nasr: Renamed to `startApp`

src/eventHandlers/WorkerInitHandler.ts
ejizba: I would honestly prefer if `msg.functionAppDirectory` was `undefined` instead of pointing to dummy app content, because I don't necessarily trust the dummy app content to stay the same forever. So no, I would not call it "unexpected behavior" when it's my preferred behavior (for the case of placeholders).

Keep in mind that too many logs or misleading logs can make servicing more difficult, so more logs does not always equal better. If you feel strongly that this needs a log, I would be okay with a compromise where you add a log at the "Information" level, which is easier to filter and does not construe this as a problem

src/eventHandlers/InvocationHandler.ts
hossam-nasr: Done

test/appStartup.test.ts
hossam-nasr: My purpose of this test was to ensure that the app-level `hookData` was persisted in the app scope. A better test would have been to ensure that a hypothetical app teardown hook sees changes that app start hooks make, but that hypothetical hook doesn't exist yet. I rewrote the title of this test though. Does it make more sense now?

hossam-nasr:
Addressed most comments, should be ready for another pass :)

test/startApp.test.ts
```diff
@@ -0,0 +1,161 @@
...
+        await stream.assertCalledWith(
+            WorkerInitMsg.receivedInitLog,
+            WorkerInitMsg.warning('Worker failed to load package.json: file does not exist'),
+            Msg.executingHooksLog(1, 'appStart'),
+            Msg.executedHooksLog('appStart'),
+            WorkerInitMsg.response
+        );
+
+        expect(startFunc.callCount).to.be.equal(1);
+        expect(channel.appLevelOnlyHookData).to.deep.equal(expectedHookData);
```
ejizba: > My purpose of this test was to ensure that the app-level hookData was persisted in the app scope. A better test would have been to ensure that a hypothetical app teardown hook sees changes that app start hooks make, but that hypothetical hook doesn't exist yet. I rewrote the title of this test though. Does it make more sense now?

I don't want you to rename the test, I want you to change the `expect` calls to reflect feature behavior that cannot be changed in the future or remove the test since teardown hooks don't exist yet. One of the main goals of unit tests is to prevent accidental regressions. Therefore, if someone changes WorkerChannel in the future and your test starts failing, they would be reasonable to assume their change caused a regression and they need to revert their change (even though that's not true). That's why unit tests should avoid testing internal behavior like this.

src/eventHandlers/InvocationHandler.ts
```diff
@@ -117,7 +120,8 @@ export class InvocationHandler extends EventHandler<'invocationRequest', 'invoca
             }
 
             const postInvocContext: PostInvocationContext = {
-                hookData,
+                hookData: preInvocContext.hookData,
```
ejizba: I don't think you meant to change behavior here, but you actually did. I would suggest changing this back to this:
```suggestion
                hookData: invocationHookData,
```

The behavior in question is "Should we let a hook change `hookData` to a different object, essentially erasing any existing state?" Aka in this example code should it print 'world' at the end or not:
```
// pre invoc hook 1
preInvocContext.hookData.hello = 'world';

// pre invoc hook 2
preInvocContext.hookData = {};

// post invoc hook 1
console.log(postInvocContext.hookData.hello);
```

I think the answer should be no, mainly because I expect these hooks can come from multiple sources that have no idea about each other and we should do our best to prevent one source from messing up another (e.g. one source is the user's app, another source is the "applicationinsights" npm package).

We are a bit inconsistent in how we handle this situation before your PR. We let users overwrite hookData within the same hook type, but not across hook types. To fix that, I would suggest we modify `WorkerChannel.executeHooks` to have the following code. You can either include it in your PR or let me do it as a separate PR. At a minimum, your PR should maintain existing behavior
```typescript
for (const callback of callbacks) {
    const oldHookData = context.hookData;
    const oldAppHookData = context.appHookData;
    await callback(context);
    if (context.hookData !== oldHookData || context.appHookData !== oldAppHookData) {
        context.hookData = oldHookData;
        context.appHookData = oldAppHookData;
        this.log({
            message:
                'WARNING: Ignoring changes to hook data. You may modify the hook data object, but cannot replace it.',
            level: LogLevel.Warning,
            invocationId,
            category: msgCategory,
        });
    }
}
```

src/eventHandlers/InvocationHandler.ts
```diff
@@ -140,6 +144,7 @@ export class InvocationHandler extends EventHandler<'invocationRequest', 'invoca
                 throw postInvocContext.error;
             }
             const result = postInvocContext.result;
+            channel.appHookData = postInvocContext.appHookData;
```
ejizba: Remove this line, for the same reason as the above comment.

src/eventHandlers/WorkerInitHandler.ts
```diff
@@ -30,9 +31,19 @@ export class WorkerInitHandler extends EventHandler<'workerInitRequest', 'worker
...
+        if (msg.hostVersion) {
+            channel.hostVersion = msg.hostVersion;
+        }
+
+        if (msg.functionAppDirectory) {
+            await startApp(msg.functionAppDirectory, channel);
+        } else {
+            channel.log({
+                message: 'functionAppDirectory was undefined in workerInitRequest message. Skipping startup logic.',
+                level: LogLevel.Warning,
```
ejizba: As mentioned, this needs to be `LogLevel.Information` or `LogLevel.Debug` if you're going to include a message about this. Also, please make the message more user-friendly with less internal variable names. Aka:

> Skipping app start logic because no function app directory was provided during worker init

Also if you're going to do this, you might as well do it in functionEnvironmentReload as well to be consistent.

src/startApp.ts
```diff
@@ -0,0 +1,63 @@
...
+    await channel.updatePackageJson(functionAppDirectory);
+    await loadEntryPointFile(functionAppDirectory, channel);
+    const appStartContext: AppStartContext = {
+        hookData: channel.appLevelOnlyHookData,
+        appHookData: channel.appHookData,
+        functionAppDirectory,
+        hostVersion: channel.hostVersion,
+    };
+    await channel.executeHooks('appStart', appStartContext);
+    channel.appHookData = appStartContext.appHookData;
```
ejizba: Remove line 32 and 33 - same reason as other comment

types-core/index.d.ts
```diff
@@ -83,6 +89,21 @@ declare module '@azure/functions-core' {
...
+     */
+    export interface AppStartContext extends HookContext {
+        /**
+         * Absolute directory of the function app
+         */
+        functionAppDirectory: string;
+        /**
+         * The version of the host running the function app
+         */
+        hostVersion?: string;
```
ejizba: Any idea why this would ever be undefined? It would be more convenient for users if it was not optional.

src/eventHandlers/WorkerInitHandler.ts
hossam-nasr: No that makes sense, I'm okay with not having the log. It is more unexpected behavior when starting the worker in "normal" rather than "placeholder" mode. If anything, I think it would be more unexpected for the `functionAppDirectory` to be `undefined` in the `functionEnvironmentReloadRequest` when specializing, and I see that there, we don't have a log if it is `undefined`, so since there's precedent there, I'm okay with that.

test/startApp.test.ts
hossam-nasr: So if there isn't a teardown hook yet, should we just remove this test entirely until there is one?

src/eventHandlers/InvocationHandler.ts
hossam-nasr: This actually was an intentional change, as I thought the original behavior was an oversight. In my opinion, I thought that setting the `hookData` state through something like:

```TS
preInvocContext.hookData = {test: 'hello'};
```

is an intuitive way of setting this `hookData` object (see the unit tests I wrote for example), and I would be confused why setting it that way, I'm not able to see the new state in my post-invocation hook. I do see your point though about not wanting different sources to influence each other, but I think that is somewhat inevitable and should be expected from these hooks, since if we _don't_ allow overriding, then user hooks will be forced to see state from app insights, for example. What if they both try to set a value of the same name and they don't get the behavior they expect? Point is, some interference between the two is inevitable, regardless of which way we go.

Another suggestion would be, we don't allow `preInvocationContext.hookData = {}` to _remove_ existing state, but we do allow it to _add_ new state. For example:

```TS
// pre invoc hook 1
preInvocContext.hookData.hello = 'world';

// pre invoc hook 2
preInvocContext.hookData = { test: 'hello' };

// post invoc hook 1
console.log(postInvocContext.hookData);
```
the value printed would be essentially the same behavior as `Object.assign`: 

```TS
{
    hello: 'world',
    test: 'hello'
}
```

That being said, regardless of which behavior we take, I think we should have consistent behavior between `appHookData` and `hookData`, and also consistent behavior for `hookData` for the different scopes. For example, if `preInvocationContext.hookData = {}` is not honored, then `preInvocationContext.appHookData = {}` should also not be honored.

types-core/index.d.ts
hossam-nasr: Looking at host code, it shouldn't ever be `undefined`, so I think we should be able to use `nonNullProp` here, but the same could be said about `functionAppDirectory`. Ideally we should treat them similarly?

ejizba: The host version is a static string that has never and will never change during the lifecycle of a worker. We already know one scenario where the functionAppDirectory changes values during the worker's lifecycle (specialization) and I could see more scenarios popping up in the future. The difference is that if `hostVersion` is undefined, that's a bug in the host and we should fix that. If `functionAppDirectory` is undefined, that's reasonable behavior

src/eventHandlers/InvocationHandler.ts
ejizba: > In my opinion, I thought that setting the hookData state through something like is an intuitive way of setting this hookData object (see the unit tests I wrote for example)

I disagree that it's intuitive because it inherently only works for one hook. Ignoring multiple sources, if you have multiple hooks in your own app this would overwrite your other hook's data every time. `hookData` is like a shared photo album with your family. People can add and remove photos, but they can't create a new album each time because that would destroy everyone else's contributions.

> some interference between the two is inevitable, regardless of which way we go.

Yes and that's by design. In order to share data, we have to let users interfere with each other to some extent. I would recommend the app insights team put all of their data in a nested object with a very unique key (`microsoft-applicationinsights`?) to prevent accidental modifications. The programming model package can also help make it easier to generate unique keys, but I don't want that kind of logic in the core module.

> That being said, regardless of which behavior we take, I think we should have consistent behavior between appHookData and hookData, and also consistent behavior for hookData for the different scopes.

I agree with this - but that's not the goal of this PR. This has already taken longer than I would like for any PR, meaning it was fixing too many things at once (not that the code itself is bad). Keep the scope small, maintain existing behavior if it's unrelated to the title of the PR, and we should be able to get every PR merged within a week or two IMO.

types-core/index.d.ts
hossam-nasr: Ok, I changed it to `nonNullProp`

test/startApp.test.ts
ejizba: Well I'm trying to find ways to speed this PR up, so yes the easiest answer is to just remove this test. I would say the most correct answer would be to add a second 'appStart' hook and you could at least ensure app level hookData is shared across multiple hooks

src/eventHandlers/InvocationHandler.ts
hossam-nasr: > `hookData` is like a shared photo album with your family. People can add and remove photos, but they can't create a new album each time because that would destroy everyone else's contributions.

This makes sense. I think my preference would be the `Object.assign`-like behavior, it feels like a good compromise and makes intuitive sense to me.

> I agree with this - but that's not the goal of this PR. This has already taken longer than I would like for any PR, meaning it was fixing too many things at once (not that the code itself is bad). Keep the scope small, maintain existing behavior if it's unrelated to the title of the PR, and we should be able to get every PR merged within a week or two IMO.

For the purposes of speeding up this PR, I have created a new issue here for discussion and tracking: #601, and will just revert to old behavior for both hooks for this PR

test/startApp.test.ts
hossam-nasr: I went with the second option :P

src/eventHandlers/WorkerInitHandler.ts
```diff
@@ -30,9 +32,15 @@ export class WorkerInitHandler extends EventHandler<'workerInitRequest', 'worker
         });
 
         logColdStartWarning(channel);
-        const functionAppDirectory = msg.functionAppDirectory;
-        if (functionAppDirectory) {
-            await channel.updatePackageJson(functionAppDirectory);
+
+        channel.hostVersion = nonNullProp(msg, 'hostVersion');
+
+        if (msg.hostVersion) {
```
ejizba: nit remove these lines - they're duplicate of line 36

src/WorkerChannel.ts
```diff
@@ -15,8 +15,14 @@ export class WorkerChannel {
     eventStream: IEventStream;
     functionLoader: IFunctionLoader;
     packageJson: PackageJson;
+    hostVersion = '';
+    // this hook data will be passed to (and set by) all hooks in all scopes
```
ejizba:
```suggestion
    /**
     * this hook data will be passed to (and set by) all hooks in all scopes
     */
```

If you do comments like this, they'll show up in Intellisense. I don't know the details on why, but it's just always been that way 🤷‍♂️
Screen Shot 2022-06-27 at 6 18 06 PM

src/WorkerChannel.ts
```diff
@@ -15,8 +15,14 @@ export class WorkerChannel {
     eventStream: IEventStream;
     functionLoader: IFunctionLoader;
     packageJson: PackageJson;
+    hostVersion = '';
```
ejizba:
```suggestion
    /**
     * This will only be set after worker init request is received
     */
    hostVersion?: string;
```

Making it optional with `?` is more tedious (you'll have to use `nonNullProp` more), but more correct. Timing issues are prime candidates for `undefined`/`nonNullProp` and this fits the bill because someone might try to access `hostVersion` before we get the worker init request

ejizba (approved):

PR #579, opened 2022-05-06
Fix "Cannot find module" e2e test failure

ejizba:
A Node.js test in the Core tools e2e [integration build](https://azfunc.visualstudio.com/Azure%20Functions/_build/results?buildId=67913&view=ms.vss-test-web.build-test-results-tab) is currently failing with an error like this:

> Cannot find module '/home/vsts/work/1/s/test/end-to-end/testFunctionApp/QueueTriggerAndOutput/index.js'

The bug only happens for the webpacked version of the worker and it came about because the webpack config hard-coded a file name and I moved some code out of that file in https://github.com/Azure/azure-functions-nodejs-worker/pull/576. The code in question does a 'require' on the user's scriptFilePath which is a dynamic value that webpack can't/shouldn't handle. I improved the webpack config so that we can use a simple comment in the code to tell webpack to ignore that 'require'. This should prevent similar bugs from happening in the future regardless of where we put this code.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -49,7 +49,6 @@
                 "semver": "^7.3.5",
                 "shx": "^0.3.3",
                 "sinon": "^7.0.0",
-                "string-replace-webpack-plugin": "^0.1.3",
                 "ts-node": "^3.3.0",
                 "typescript": "^4.5.5",
                 "typescript3": "npm:typescript@~3.7.0",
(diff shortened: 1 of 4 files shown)


hossam-nasr:
It looks like some E2E tests are still failing?

ejizba:
The existing errors are "Condition not reached within timeout." and timeout errors are usually transient

alrod (approved):

PR #582, opened 2022-05-10
Detect blocking sync operations during function executions #558

### LINKED ISSUE #558: Detect blocking sync operations during function executions
On blocking sync operation the nodejs worker becomes unresponsive:
1. New invocations hang as nodejs worker is not able to read them from grpc and can reach timeouts/long executions.
2. The logs from nodejs worker can not be written to grpc.
3. It's hard to understand the function code stack trace for blocking operation.

Related [CRI](https://portal.microsofticm.com/imp/v3/incidents/details/279999271/home)

We need to investigate how to detect sync blocking for a customer.
Possible solution [https://www.npmjs.com/package/blocked-at](https://www.npmjs.com/package/blocked-at)

alrod:
Fixes https://github.com/Azure/azure-functions-nodejs-worker/issues/558


package-lock.json
```diff
@@ -1,4751 +1,8 @@
 {
     "name": "azure-functions-nodejs-worker",
     "version": "3.4.0",
-    "lockfileVersion": 2,
+    "lockfileVersion": 1,
```
ejizba: Can you update your version of `npm` (`npm install -g npm`) and run `npm install` again? We want this to stay at lockfileVersion 2 and I think you need to be on at least v7 for that to happen

package.json
```diff
@@ -7,6 +7,7 @@
     "dependencies": {
         "@grpc/grpc-js": "^1.2.7",
         "@grpc/proto-loader": "^0.6.4",
+        "blocked-at": "^1.2.0",
```
ejizba, line 10: Add `@types/blocked-at` as a dev dependency so we get the types, too

src/Worker.ts
```diff
@@ -69,4 +70,8 @@ export function startNodeWorker(args) {
     process.on('exit', (code) => {
         systemLog(`Worker ${workerId} exited with code ${code}`);
     });
+
+    if (process.env.AZURE_FUNCTIONS_NODE_BLOCK_LOG) {
```
ejizba: Here's a little helper function I've used in the past to check if an env var is set:

```typescript
export function isEnvironmentVariableSet(val: string | boolean | number | undefined | null): boolean {
    return !/^(false|0)?$/i.test(val === undefined || val === null ? '' : String(val));
}
```

Basically if the env var  is 0, false, or empty - it counts as `false`. Everything else is `true`

src/utils/blockedMonitor.ts
```diff
@@ -0,0 +1,32 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License.
+
+import { AzureFunctionsRpcMessages as rpc } from './../../azure-functions-language-worker-protobuf/src/rpc';
+import LogCategory = rpc.RpcLog.RpcLogCategory;
+import LogLevel = rpc.RpcLog.Level;
+import blockedAt = require('blocked-at');
+
+export function startBlockedMonitor(channel: { log: (log: rpc.IRpcLog) => void }, threshold = 100): NodeJS.Timer {
+    let blockedHistory: { time: string; duration: number; stack: string; resource: string }[] = [];
```
ejizba: From the "blocked-at" docs:

> The stack trace is pointing to a start of a function called asynchronously, so in most cases the first stack frame pointing to your code is where you need to start analyzing all synchronous operations to find the slow one.
> 
> In some cases your code is not directly called and tracking it down will still be difficult.

This concerns me a bit. I would find it very helpful if it had the sync call stacks, but just the async ones feels like it may mislead some users and still not have the important info. The other piece that concerns me is that by nature the warning message can't be sent until _after_ the sync code finishes, meaning if your code hangs indefinitely you won't see these logs.

Have you considered these drawbacks and do you think this is still worth it?

Just FYI @liliankasem is in the early stages of addressing this same situation with a "worker heartbeat" feature. It may or may not work in the case of Node.js, but it is another avenue to explore: https://dev.azure.com/msazure/One/_git/AAPT-Antares-Docs/pullrequest/6035812
(If you have any CRIs where a user had problems with blocking sync code please forward those along to Lilian)

alrod: Yes, I considered these drawbacks. The actual call stacks were not very useful but it's better then nothing, also it includes "resource" of the blocking. Anyway I would not document the feature and only tell about it during CRIs investigations as the feature enabling can impact the perf. There is another lib which does not impact the perf but does not have the stack traces: https://www.npmjs.com/package/blocked. This feature can be constantly enabled. Do we think it will work better?

Regarding @liliankasem's doc we scheduled a meeting. actually we already have "WrkerStatusRequest" feature which

test/blockMonitorTest.ts
```diff
@@ -0,0 +1,32 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License.
+
+import { assert } from 'console';
```
ejizba: All of the following modules export their own version of `assert` and unfortunately the one from `console` will not fail the test (it just writes a log to the console):
Screen Shot 2022-05-25 at 10 30 42 AM

Some of those would work, but our other tests actually use `expect` not `assert`:
```typescript
import { expect } from 'chai';

expect(log.level).to.equal(LogLevel.Warning);
expect(message).to.equal('test');
```

src/utils/blockedMonitor.ts
ejizba: I agree the call stacks are better than nothing _if you know they only have async calls_, but otherwise I worry they might confuse users. As long as we add a bit of explanation in the logs I think that would suffice for me

package.json
ejizba: Still waiting on the above comment

src/utils/blockedMonitor.ts
```diff
@@ -0,0 +1,32 @@
...
+
+    //threshold - minimum miliseconds of blockage to report.
+    //other parameters are default, more details on https://github.com/naugtur/blocked-at.
+    blockedAt(
+        (ms, stack, resource) => {
+            blockedHistory.push({ time: new Date().toUTCString(), duration: ms, stack: stack, resource: resource });
```
ejizba: A threshold of 100 ms feels too small to me. I tested and consistently got false positives lasting around 200-300 ms. I doubt users would notice and file a CRI for a blockage that small anyways. How about 5000 ms?

If we have a larger threshold, we can get rid of the `setInterval` and just log directly from inside `blockedAt`. We also would not need to print the `time` since the time associated with the log would be accurate in that case.

src/utils/blockedMonitor.ts
```diff
@@ -0,0 +1,32 @@
...
+import LogLevel = rpc.RpcLog.Level;
+import blockedAt = require('blocked-at');
+
+export function startBlockedMonitor(channel: { log: (log: rpc.IRpcLog) => void }, threshold = 100): NodeJS.Timer {
+    let blockedHistory: { time: string; duration: number; stack: string; resource: string }[] = [];
+
+    //threshold - minimum miliseconds of blockage to report.
+    //other parameters are default, more details on https://github.com/naugtur/blocked-at.
+    blockedAt(
+        (ms, stack, resource) => {
```
ejizba: What information is `resource` supposed to contain? When I test I just see `{"type":"TickObject"}` or `{"type":"PROMISE"}` which doesn't mean much to me.

Looking at the docs, maybe we have to set `resourcesCap` for it to be useful?

src/utils/blockedMonitor.ts
```diff
@@ -0,0 +1,32 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License.
+
+import { AzureFunctionsRpcMessages as rpc } from './../../azure-functions-language-worker-protobuf/src/rpc';
+import LogCategory = rpc.RpcLog.RpcLogCategory;
+import LogLevel = rpc.RpcLog.Level;
+import blockedAt = require('blocked-at');
+
+export function startBlockedMonitor(channel: { log: (log: rpc.IRpcLog) => void }, threshold = 100): NodeJS.Timer {
```
ejizba: Since this will have an impact on performance, can we add a warning at the beginning so that users don't forget to turn it off

src/utils/blockedMonitor.ts
```diff
@@ -0,0 +1,32 @@
...
+    //other parameters are default, more details on https://github.com/naugtur/blocked-at.
+    blockedAt(
+        (ms, stack, resource) => {
+            blockedHistory.push({ time: new Date().toUTCString(), duration: ms, stack: stack, resource: resource });
+        },
+        { threshold: threshold }
+    );
+
+    // Log blockedHistory if it's not empty each 10 seconds
+    return setInterval(() => {
```
ejizba, line 40: Here is what the function would look like if you combine all my comments (please read those first).

IMO the main unknowns are (1) if you're okay doing a threshold of 5000 as that impacts a lot of the code structure and (2) if you still want to print `resource` depending on the answer to my question about it

```typescript
export function startBlockedMonitor(channel: { log: (log: rpc.IRpcLog) => void }, threshold = 5000): void {
    function systemWarn(message: string) {
        channel.log({
            message,
            level: LogLevel.Warning,
        });
    }

    systemWarn(
        `Warning: Monitoring for blocking code is turned on, with a threshold of ${threshold} ms. This will have a negative impact on performance. Adjust "AZURE_FUNCTIONS_NODE_BLOCK_LOG" to turn it off.`
    );

    //threshold - minimum milliseconds of blockage to report.
    //other parameters are default, more details on https://github.com/naugtur/blocked-at.
    blockedAt(
        (ms, stack, _resource) => {
            systemWarn(`Warning: Event loop blocked for ${Math.round(ms)} ms.`);
            systemWarn(
                `IMPORTANT NOTE: The following stack trace is only an approximation and you should analyze all synchronous operations from that point onward:`
            );
            systemWarn(stack.join(EOL));
        },
        { threshold: threshold }
    );
}
```

alrod: 5 secs threshold looks long for me - it would report only delays > 5 secs.
My idea was to detect delays < 100ms, accumulate history and report each 10 seconds so the grpc channel would be overwhelmed with the logs. In case we are using `systemWarn`(console logs) I am not sure about perf impact if we have to many logs.

Regarding `resource` - if the signature provides the parameter why do not log it even it can be not very useful?

src/utils/blockedMonitor.ts
alrod: Unfortunately the documentation does not explain the `resource` in details, if the signature provides the parameter why do not log it even it can be not very useful?

`resourceCap` was left as default as it needs experimenting for different scarious.

package.json
```diff
@@ -25,6 +25,7 @@
         "@types/sinon": "^7.0.0",
         "@typescript-eslint/eslint-plugin": "^5.12.1",
         "@typescript-eslint/parser": "^5.12.1",
+        "blocked-at": "^1.2.0",
```
ejizba: Still not right 😝

`blocked-at` should be in `dependencies`
`@types/blocked-at` should be in `devDependencies`

They're two separate packages. The first gives us the working javascript and needs to ship as a prod dependency, the second gives us the typescript types used to check for any build errors and is only needed as a dev dependency.

ejizba (approved):

PR #604, opened 2022-06-28
Fix environment variables casing bug

### LINKED ISSUE #600: Maintain environment variable case insensitivity for Windows after worker specialization
This issue was prompted by [this CRI](https://portal.microsofticm.com/imp/v3/incidents/details/312919342/home): a customer was facing frequent crashes on some Windows workers with an error that a particular environment variable is not set. After investigation, it turned out that the environment variable in question was being set in app settings and referenced from `process.env` using different casing, and this issue was only occurring when the worker had undergone specialization (after being sent a `functionEnvironmentReloadRequest`).

Normally, in Windows, setting/getting environment variables is expected to be case-insensitive, and `process.env` in node treats environment variables case insensitively on Windows. Example:

```bash
PS C:\Users\hossamnasr> $Env:test="hello"
PS C:\Users\hossamnasr> $Env:test
hello
PS C:\Users\hossamnasr> $Env:TeSt
hello
PS C:\Users\hossamnasr> node
Welcome to Node.js v16.13.0.
Type ".help" for more information.
> process.env['test']
'hello'
> process.env['TeSt']
'hello'
```

However, that behavior is _not_ maintained after worker specialization. This is because in the `functionEnvironmentReloadHandler` [here](https://github.com/Azure/azure-functions-nodejs-worker/blob/bbfc4abb264e653a22e1fde1518d15b8299db421/src/eventHandlers/FunctionEnvironmentReloadHandler.ts#L37), we are _overriding_ the `process.env` object with a new object we get from the Host:

```JS
process.env = Object.assign({}, msg.environmentVariables);
```

So now, environment variables behave case sensitively:

```node
> process.env = Object.assign({}, {'test': 'hello'})
{ test: 'hello' }
> process.env['test']
'hello'
> process.env['TeSt']
undefined
```

Instead, we should be setting each environment variable we get from the Host one by one, to ensure we maintain case insensitivity and not override the entire `process.env` object:

```node
> process.env['test'] = 'hello'
'hello'
> process.env['test']
'hello'
> process.env['TeSt']
'hello'
```

#### Repro steps

1. Create a function app with any trigger on Windows
2. Set some environment variable `foo` in app settings
3. Reference this environment variable in your function _using different casing_ (e.g. `fOo`):

```TS
const timerTrigger: AzureFunction = async function (context: Context, myTimer: any): Promise<void> {
    if (process.env['fOo'] === undefined) {
        throw new Error("Env variable undefined :(")
    }
};
```

#### Expected behavior

App works normally without crashes.

#### Actual behavior

Occasionally, the app will crash with error "Env variable undefined :(" if a worker is specialized

#### Known workarounds

Treat app settings case sensitively for Windows

hossam-nasr:
Resolves #600. Case insensitive behavior should only exist on Windows. Simply deleting and resetting keys on the `process.env` object allows it to maintain its default behavior, allowing case insensitivity on Windows and disallowing it on Mac/Linux

### DIFF
--- a/src/eventHandlers/FunctionEnvironmentReloadHandler.ts
+++ b/src/eventHandlers/FunctionEnvironmentReloadHandler.ts
@@ -34,7 +34,11 @@ export class FunctionEnvironmentReloadHandler extends EventHandler<
             logCategory: LogCategory.System,
         });
 
-        process.env = Object.assign({}, msg.environmentVariables);
+        // reset existing env vars
+        Object.keys(process.env).map((key) => delete process.env[key]);
+        // set new env vars
+        Object.assign(process.env, msg.environmentVariables);
+
         // Change current working directory
         if (msg.functionAppDirectory) {
             channel.log({
--- a/test/eventHandlers/FunctionEnvironmentReloadHandler.test.ts
+++ b/test/eventHandlers/FunctionEnvironmentReloadHandler.test.ts
@@ -103,28 +103,31 @@ describe('FunctionEnvironmentReloadHandler', () => {
         expect(process.env.PlaceholderVariable).to.be.undefined;
     });
 
-    if (process.platform === 'win32') {
-        it('preserves case insensitivity of environment variables on Windows', async () => {
-            process.env.PlaceholderVariable = 'TRUE';
-            stream.addTestMessage({
-                requestId: 'id',
-                functionEnvironmentReloadRequest: {
-                    environmentVariables: {
-                        hello: 'world',
-                        SystemDrive: 'Q:',
-                    },
-                    functionAppDirectory: null,
+    it('preserves OS-specific casing behavior of environment variables', async () => {
+        process.env.PlaceholderVariable = 'TRUE';
+        stream.addTestMessage({
+            requestId: 'id',
+            functionEnvironmentReloadRequest: {
+                environmentVariables: {
+                    hello: 'world',
+                    SystemDrive: 'Q:',
                 },
-            });
-            await stream.assertCalledWith(Msg.reloadEnvVarsLog(2), Msg.reloadSuccess);
-            expect(process.env.hello).to.equal('world');
-            expect(process.env.HeLlO).to.equal('world');
-            expect(process.env.SystemDrive).to.equal('Q:');
-            expect(process.env.systemdrive).to.equal('Q:');
-            expect(process.env.PlaceholderVariable).to.be.undefined;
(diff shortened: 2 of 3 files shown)


src/eventHandlers/FunctionEnvironmentReloadHandler.ts
```diff
@@ -34,7 +34,11 @@ export class FunctionEnvironmentReloadHandler extends EventHandler<
             logCategory: LogCategory.System,
         });
 
-        process.env = Object.assign({}, msg.environmentVariables);
+        // reset existing env vars
+        Object.keys(process.env).map((key) => delete process.env[key]);
```
ejizba, line 39: We "reset" existing env vars before your change, right? Since we completely overwrote process.env with a new object? By itself, this line feels dangerous - but assuming we're matching previous behavior I think it makes sense

test/eventHandlers/FunctionEnvironmentReloadHandler.test.ts
```diff
@@ -103,6 +103,29 @@ describe('FunctionEnvironmentReloadHandler', () => {
         expect(process.env.PlaceholderVariable).to.be.undefined;
     });
 
+    if (process.platform === 'win32') {
+        it('preserves case insensitivity of environment variables on Windows', async () => {
```
ejizba: We should test all OS's here to make sure we don't run into the _opposite_ bug on mac/linux.

Aka change the name to something like this:
> respects OS-specific casing behavior of environment variables

And move the `if windows` check into the test itself (with a comment to make it more obvious):
```typescript
// On Windows operating systems, environment variables are case-insensitive.
// https://nodejs.org/docs/api/process.html#processenv
if (process.platform === 'win32') {
    expect(process.env.HeLlO).to.equal('world');
    // etc.
} else {
    expect(process.env.HeLlO).to.be.undefined;
    // etc.
}
```

src/eventHandlers/FunctionEnvironmentReloadHandler.ts
hossam-nasr: Yes, and the current unit tests ensure that behavior, e.g. https://github.com/Azure/azure-functions-nodejs-worker/blob/ad99f5d4f5613e42c7630b39d9f24a849116d9c6/test/eventHandlers/FunctionEnvironmentReloadHandler.test.ts#L111

ejizba (approved):

Issue #601, opened 2022-06-27
How should hooks share data?


hossam-nasr:
Right now, we have the `hookData` property in all hook contexts, which allows different hooks _of the same cope_ to share data and communicate with each other. There is also the `appHookData`, which allows communication across scopes. The behavior of the `hookData` property is currently inconsistent in terms of allowing overriding it with a new object. Currently, we allow overwriting the `hookData` property within the same hook type, but not across hook types. We should have consistent behavior between `hookData` and `appHookData`, and also consistent behavior for `hookData` for different hook types/scopes. This issue is to discuss the different options for this behavior. Specifically, the question is should we allow hooks to overwrite `hookData` with a new object?

### Option 1 : No Overwriting

```TS
// pre invoc hook 1
preInvocContext.hookData.hello = 'world';

// pre invoc hook 2
preInvocContext.hookData = { foo: 'bar' };

// post invoc hook 1
console.log(postInvocContext.hookData);
```

This would print:

```TS
{
    hello: 'world'
}
```

Effectively ignoring the changes in pre invoc hook 2, perhaps with also a warning along the lines of "WARNING: Ignoring changes to hook data. You may modify the hook data object, but cannot replace it."

### Option 2: Allow Overwriting

```TS
// pre invoc hook 1
preInvocContext.hookData.hello = 'world';

// pre invoc hook 2
preInvocContext.hookData = { foo: 'bar' };

// post invoc hook 1
console.log(postInvocContext.hookData);
```

This would print:

```TS
{
    foo: 'bar'
}
```

Effectively erasing the state set by pre invoc hook 1, perhaps also with a warning message.

### Option 3: `Object.assign()` behaviour:

```TS
// pre invoc hook 1
preInvocContext.hookData.hello = 'world';

// pre invoc hook 2
preInvocContext.hookData = { foo: 'bar' };

// post invoc hook 1
console.log(postInvocContext.hookData);
```

This would print:

```TS
{
    hello: 'world',
    foo: 'bar'
}
```

ejizba:
### Option 4: readonly behaviour:

We can mark `hookData` readonly (we might have to refactor a tiny bit under the covers to make this work) to make it more clear that you can't overwrite it.

```TS
// pre invoc hook 1
preInvocContext.hookData.hello = 'world';

// pre invoc hook 2
preInvocContext.hookData = { foo: 'bar' };

// post invoc hook 1
console.log(postInvocContext.hookData);
```

[...]

Issue #586, opened 2022-05-24
context.log({ object: 42}) does not stringify object before logging


sedot42:
Hi! I really like Azure Functions for Node.js, but recently encountered an issue that causes imo unnecessary work. Please excuse me if this is the wrong repo to file this issue.

### Repro steps

```js
module.exports = async function (context) {
    context.log({foo: "bar", 1: 42});
    context.done();
}
```

### Expected behavior

Logging an object should write a stringified JSON to the logs, because otherwise parsing this information in e.g. Application Insights requires transforming the log message to proper JSON. 

It should look like this in the console:

`[2022-05-24T13:37:52.421Z] {"1":42,"foo":"bar"}`

And the `message` property in ApplicationInsights should be `'{"1":42,"foo":"bar"}'`

### Actual behavior

Console log: `[2022-05-24T13:35:48.326Z] { '1': 42, foo: 'bar' }`

`message` property in ApplicationInsights: `"{ '1': 42, foo: 'bar' }"`

### Known workarounds

There are two workarounds:

1. Send proper JSON to ApplicationInsights by logging objects via `JSON.stringify()`.
2. Transform the logged string with something like KQL `replace_regex(replace_regex(message, @"'", @'"'), @'([\w]+):', @'"\1":')`

While you can argue that developers should do 1., I argue that it is more intuitive to JSON.stringify() objects by default, thus not letting devs step into this trap.

ejizba:
It looks like we use [util.format](https://nodejs.org/api/util.html#utilformatformat-args) to turn the arguments into a string, which is the same behavior you would see using [console.log](https://nodejs.org/api/console.html#consolelogdata-args) (with or without Azure Functions). My gut says to leave this as-is to maintain consistency and because ultimately I don't see a clear right or wrong answer.

That being said, we recently added [a feature](https://github.com/Azure/azure-functions-nodejs-worker/issues/522) called invocation hooks, which should make it easier to pick your preference and adjust the behavior for your whole app in one go. Specifically, you could do something like this:
```typescript
import { registerHook } from '@azure/functions-core';

registerHook('preInvocation', (context) => {
    function stringifyArgs(...args: any[]): any[] {
        return args.map((a) => (typeof a === 'object' ? JSON.stringify(a) : a));
    }

    const oldLogger = context.invocationContext.log;
    context.invocationContext.log = Object.assign((...args: any[]) => oldLogger(...stringifyArgs(...args)), {
[...]

PR #608, opened 2022-07-15
Split programming model out of worker

### LINKED ISSUE #568: Move existing programming model into separate npm package
I propose we split the worker up into two pieces, one that ships as-is with the host and one that ships as a separate node module (from here on out, I will reference these as the "core" and "programming model" packages respectively). This specific issue is to track moving the existing programming model out of the worker and into a separate package. The first iteration of this package should exactly replicate what the worker does today. The second iteration should be an improved programming model as tracked by https://github.com/Azure/azure-functions-nodejs-worker/issues/480.

When I say "programming model", this includes anything that makes our user's lives easier like parsing/converting/handling data from the host. Another way to think about it is if there is currently a type for something in our types package (`Context`, `HttpRequest`, etc.), the underlying code for that should probably be in the programming model package. The only thing left in the worker will likely be setting up and managing the gRPC channel used to communicate with the host

## The Split

|Core worker package ("@azure/functions-core")|Programming model package ("@azure/functions" - name TBD)|
|---|---|
|Tracked by https://github.com/Azure/azure-functions-nodejs-worker/issues/567|First version tracked by this issue|
|Required - we _have_ to create an API in some form|Optional - we do not _need_ this package. Instead, we could put everything in the core api|
|built-in module shipped with the worker - _not_ listed in user's package.json|shipped with the user's app - listed in their package.json dependencies|
|slow to update because it relies on the host release cycle|fast to update - just publishing an npm package|
|forced on the user when it does update|user decides when to take the update|
|backwards compatability is paramount|backwards compatability is nice, but we can easily do a new major version|
|barebones API, very unopinionated|fancy api with lots of nice intellisense, can be opinionated|

## Benefits

- When we make breaking changes (like [this](https://github.com/Azure/azure-functions-nodejs-worker/issues/388)), users will be able to choose when they integrate those changes as opposed to us forcing it on them in Azure.
- If we ever _want_ to force a breaking change on the user, all we have to do is change the "default" programming model package used in Azure, but users can still reference the old version in their own app.
- We can easily support multiple different programming models in their own packages (whether they're created by us or other people) at the same time. [This PR](https://github.com/Azure/azure-functions-nodejs-worker/pull/529) is one example where we run into difficulties today because our various models are combined into the worker.
- Since both JS and TS users will need the npm package to run their app, they will both get the same experience of [intellisense in VS Code](https://code.visualstudio.com/docs/nodejs/working-with-javascript#_intellisense)

ejizba:
The goal is essentially to split our code into two repos, while maintaining existing behavior:
- **azure-functions-nodejs-library**: This is a new repo I created based on offline discussion about the name. It will have the same git history as the worker repo up until this point, but will diverge moving forward. It will ship the `@azure/functions` npm package, which will include the entire programming model moving forward (not _just_ the types, although it'll still have that). See PR here: https://github.com/Azure/azure-functions-nodejs-library/pull/1
- **azure-functions-nodejs-worker**: Our existing repo. NOTE: I won't be able to merge this PR until the new npm package exists.

Most important files to review:
- `types-core/index.d.ts`: This defines the API between the two pieces
- `src/eventHandlers/InvocationHandler.ts`: This is the main file that uses the new api

Fixes https://github.com/Azure/azure-functions-nodejs-worker/issues/568


package.json
```diff
@@ -5,13 +5,12 @@
     "description": "Microsoft Azure Functions NodeJS Worker",
     "license": "(MIT OR Apache-2.0)",
     "dependencies": {
+        "@azure/functions": "file:../js-framework/azure-functions-3.4.0.tgz",
```
hossam-nasr: Shouldn't we import this as the npm package?

package.json
```diff
@@ -64,13 +61,13 @@
     },
     "scripts": {
         "clean": "rimraf dist && rimraf azure-functions-language-worker-protobuf/src/rpc*",
-        "build": "rimraf dist && npm run gen && shx mkdir -p dist/azure-functions-language-worker-protobuf/src && shx cp azure-functions-language-worker-protobuf/src/rpc.* dist/azure-functions-language-worker-protobuf/src/. && node ./node_modules/typescript/bin/tsc",
+        "build": "rimraf dist && npm run gen && shx mkdir -p dist/azure-functions-language-worker-protobuf/src && shx cp azure-functions-language-worker-protobuf/src/rpc.* dist/azure-functions-language-worker-protobuf/src/. && tsc",
```
hossam-nasr, line 64: What's the difference between calling `node ./node_modules/typescript/bin/tsc` and calling `tsc` directly?

src/FunctionLoader.ts
```diff
@@ -1,63 +1,60 @@
...
-        };
-    } = {};
+interface LoadedFunction {
+    metadata: rpc.IRpcFunctionMetadata;
+    callback: FunctionCallback;
+    thisArg: unknown;
+}
+
+export class FunctionLoader<TContext = unknown> {
+    #loadedFunctions: { [k: string]: LoadedFunction | undefined } = {};
```
hossam-nasr, line 24: why allow a loaded function to be `undefined` here?

src/FunctionLoader.ts
```diff
@@ -1,63 +1,60 @@
...
         const script: any = await loadScriptFile(nonNullProp(metadata, 'scriptFile'), packageJson);
         const entryPoint = <string>(metadata && metadata.entryPoint);
-        const [userFunction, thisArg] = getEntryPoint(script, entryPoint);
-        this.#loadedFunctions[functionId] = {
-            info: new FunctionInfo(metadata),
-            func: userFunction,
-            thisArg,
-        };
+        const [callback, thisArg] = getEntryPoint(script, entryPoint);
+        this.#loadedFunctions[functionId] = { metadata, callback, thisArg };
```
hossam-nasr, line 33: looks like we're replacing the `FunctionInfo` class with just the raw `IRpcFunctionMetadata`. Is `FunctionInfo` not used anymore? I would argue that the converters and processing done by `FunctionInfo` are helpful abstractions that should be part of the core worker not the programming model package

types-core/index.d.ts
```diff
@@ -1,37 +1,43 @@
...
-    export type HookCallback = (context: HookContext) => void | Promise<void>;
-    export type PreInvocationCallback = (context: PreInvocationContext) => void | Promise<void>;
-    export type PostInvocationCallback = (context: PostInvocationContext) => void | Promise<void>;
-    export type AppStartCallback = (context: AppStartContext) => void | Promise<void>;
+    type HookCallback = (context: HookContext) => void | Promise<void>;
+    type PreInvocationCallback<TContext = unknown> = (context: PreInvocationContext<TContext>) => void | Promise<void>;
+    type PostInvocationCallback<TContext = unknown> = (
+        context: PostInvocationContext<TContext>
+    ) => void | Promise<void>;
+    type AppStartCallback = (context: AppStartContext) => void | Promise<void>;
```
hossam-nasr, line 25: What's the difference between adding `export` and excluding it?

types-core/index.d.ts
```diff
@@ -129,4 +135,386 @@ declare module '@azure/functions-core' {
...
+         */
+        context: TContext;
+
+        /**
+         * The remaining arguments passed to a function callback, generally describing the trigger/input bindings
+         */
+        inputs: unknown[];
+    }
+
+    type FunctionCallback<TContext = unknown> = (context: TContext, ...inputs: unknown[]) => unknown;
```
hossam-nasr: Why are we parameterizing `TContext`? At the end of the day, regardless of the programming model, the actual function callback will still receive the same type for the first argument, correct?

types-core/index.d.ts
```diff
@@ -129,4 +135,386 @@ declare module '@azure/functions-core' {
...
+
+        /**
+         * Describes the current state of invocation, or undefined if between states
+         */
+        state?: InvocationState;
+
+        /**
+         * The recommended way to log information
+         */
+        log(level: RpcLog.Level, category: RpcLog.RpcLogCategory, message: string): void;
```
hossam-nasr, line 192: At a high level, should a programming model package worry or know about rpc and rpc-level information? If not, then should the core worker provide abstractions on top of, for example, `RpcFunctionMetadata`, `RpcLog.Level`, etc.?

types-core/index.d.ts
```diff
@@ -129,4 +135,386 @@ declare module '@azure/functions-core' {
...
+
+        /**
+         * A version for this programming model, generally only used for tracking purposes
+         */
+        version: string;
+
+        /**
+         * Returns a new instance of the invocation model for each invocation
+         */
+        getInvocationModel(coreContext: CoreInvocationContext): InvocationModel<TContext>;
```
hossam-nasr: So far, it looks like a programming model doesn't also include indexing the functions?

types-core/index.d.ts
hossam-nasr, line 192: Who is now responsible for logging, the worker package or the programming model package? Or both?

types-core/index.d.ts
```diff
@@ -129,4 +135,386 @@ declare module '@azure/functions-core' {
...
+    interface InvocationArguments<TContext> {
+        /**
+         * This is always the first argument passed to a function callback
+         */
+        context: TContext;
+
+        /**
+         * The remaining arguments passed to a function callback, generally describing the trigger/input bindings
+         */
+        inputs: unknown[];
```
hossam-nasr, line 234: Should this be optional? Not all functions have more than one argument

types-core/index.d.ts
```diff
@@ -129,4 +135,386 @@ declare module '@azure/functions-core' {
...
+        getArguments(): Promise<InvocationArguments<TContext>>;
+
+        /**
+         * The main method that executes the user's function callback
+         * This is run between preInvocation and postInvocation hooks
+         * @param context The context object returned in `getArguments`, potentially modified by preInvocation hooks
+         * @param inputs The input array returned in `getArguments`, potentially modified by preInvocation hooks
+         * @param callback The function callback to be executed
+         */
+        invokeFunction(context: TContext, inputs: unknown[], callback: FunctionCallback<TContext>): Promise<unknown>;
```
hossam-nasr: Since it is the worker module that still controls pre and post invocation hooks, and overriding the context and input arguments, it feels weird for a programming model to get to define the `invokeFunction` function. What extra logic could be reasonably implemented by a programming model here besides just invoking the function with the provided arguments?

Also, as it stands, `context` doesn't actually have to be the first argument to the function callback; the programming model can pass whatever it wants to the actual callback.

types-core/index.d.ts
```diff
@@ -129,4 +135,386 @@ declare module '@azure/functions-core' {
...
+         * This is run after postInvocation hooks
+         * @param context The context object created in `getArguments`
+         * @param result The result of the function callback, potentially modified by postInvocation hooks
+         */
+        getResponse(context: TContext, result: unknown): Promise<RpcInvocationResponse>;
+    }
+
+    interface InvocationArguments<TContext> {
+        /**
+         * This is always the first argument passed to a function callback
```
hossam-nasr: This isn't actually enforced

src/eventHandlers/InvocationHandler.ts
```diff
@@ -1,18 +1,23 @@
...
+import {
+    HookData,
+    InvocationState,
+    PostInvocationContext,
+    PreInvocationContext,
+    ProgrammingModel,
+    RpcFunctionMetadata,
+    RpcInvocationRequest,
+} from '@azure/functions-core';
 import { AzureFunctionsRpcMessages as rpc } from '../../azure-functions-language-worker-protobuf/src/rpc';
```
hossam-nasr, line 14: We have two sources of `rpc` types now?

src/eventHandlers/InvocationHandler.ts
```diff
@@ -25,187 +30,100 @@ export class InvocationHandler extends EventHandler<'invocationRequest', 'invoca
...
-        function onDone(): void {
-            if (isDone) {
-                const message = resultIsPromise
-                    ? `Error: Choose either to return a promise or call 'done'. Do not use both in your script. Learn more: ${asyncDoneLearnMoreLink}`
-                    : "Error: 'done' has already been called. Please check your script for extraneous calls to 'done'.";
-                systemLog(LogLevel.Error, message);
-            }
-            isDone = true;
+        inputs = preInvocContext.inputs;
+        callback = preInvocContext.functionCallback;
```
hossam-nasr, line 64: Not also overriding the context here?

hossam-nasr (commented):
Overall, I think I follow what the code is doing, and no major issues there, but I'm not sure what goal this achieves?

As far as I understand, this so far only includes modifying function invocation, not function loading and indexing, which was the major difference and advantage of having new programming models. Also, I'm not sure what benefit allowing programming models to define `invokeFunction()` and `getResponse()` achieves? For example, what's an example of a programming model that's meaningfully different from the default one that would leverage those methods?

package.json
ejizba: This was necessary when we had multiple version of typescript listed in the devDependencies above, because you weren't sure which version of tsc you were using unless you specified the explicit path. Since there's only one version now, we can just use the shortened `tsc`.

ejizba:
This work is an important stepping stone on the way to the new indexing stuff. At its core, it is purely engineering work with no functional difference to the user. Perhaps that's why the "goal" isn't clear, but I did it that way because I think that makes it a lot easier to review. I try to separate "engineering PRs that affect a lot of files" from "feature PRs". The indexing stuff is tracked by https://github.com/Azure/azure-functions-nodejs-worker/issues/569 and it will be a separate PR - that PR will be a lot smaller (comparatively 😅), but it will also be a lot more complex because it's introducing new features.

Please read through https://github.com/Azure/azure-functions-nodejs-worker/issues/568 if you haven't already. I think it answers a lot of your questions. I think the biggest piece you're missing is the _type_ of code we want in each place. We want simple code with no chance of bugs in the worker because it takes so long to ship. We want more complex code (meaning more helpful code, but also code with a higher chance of bugs) in the npm package. Also keep in mind that people can build off of our programming model npm package. So even though I'm moving FunctionInfo (and other classes) into the npm package, people can still share that code and build off of it. In other words, a "new" programming model doesn't have to start from scratch - it can just wrap the one we provide in the npm package.

ejizba:
Sent a meeting invite to discuss offline - I think that'll be easiest

types-core/index.d.ts
ejizba: Adding `export` does nothing here which is why I removed it

src/FunctionLoader.ts
ejizba: Because depending on the key it _will_ be undefined. TypeScript actually has a strict rule to help with this exact scenario, but I haven't turned it on in the repo yet. Once I turn the rule on, I could remove `| undefined` because it would be assumed to be there.

More info:
https://www.typescriptlang.org/tsconfig#noUncheckedIndexedAccess
https://github.com/microsoft/TypeScript/issues/13778

types-core/index.d.ts
ejizba: I don't have a strong preference. I figured they can just pass an empty array if there are no arguments

src/eventHandlers/InvocationHandler.ts
ejizba: That is correct. Similar to the debate we had for hooksData (https://github.com/Azure/azure-functions-nodejs-worker/issues/601), I don't want people overriding context, just modifying it. It is a class we as the worker created, and it has too much important information shared across our worker, the library package, and the user's app to be overwritten.

By contrast, the inputs and callback were created directly by the user and the programming model so I think it makes sense to let them overwrite it. Also they are just much simpler than context

src/FunctionLoader.ts
ejizba: Discussed offline. This is one of the files that moved to the library

types-core/index.d.ts
ejizba: Discussed offline. I want to separate engineering changes from feature changes

types-core/index.d.ts
ejizba: Discussed offline. I want the worker to simply pass along all rpc information and let the library handle any manipulation

ejizba: They both do logging stuff. However, the core api expects an already formatted string here (hence why message is just type `string`) and the programming model package is in charge of formatting the string. That's because of bugs like this where people want different formatting behavior that conflict with each other, so it would be a breaking change if we ever modified this: https://github.com/Azure/azure-functions-nodejs-worker/issues/586

types-core/index.d.ts
ejizba: Discussed offline. One example of different ways to invoke the function is `context.done` vs async/await

src/eventHandlers/InvocationHandler.ts
ejizba: Discussed offline. That is somewhat correct - the `...-protobuf/src/rpc` types are still the main source. The types-core stuff is just a copy of the minimum rpc types that are relevant to the core api and programming model package (<500 lines copied from >5,000 total).

types-core/index.d.ts
ejizba: You are correct - I will update the comment

types-core/index.d.ts
ejizba: I can get rid of the parameterization. It's probably more annoying than it's worth

src/eventHandlers/InvocationHandler.ts
hossam-nasr: What do you mean? I'm talking about the plain `context`, which is created by the invocation model, not the `coreCtx`. This is one is created and managed by the programming model, so it makes sense if it wants to override it? How is it used by the worker?

hossam-nasr (approved):

src/eventHandlers/InvocationHandler.ts
ejizba: Yes you are correct it is the plain `context` not `coreCtx` so the worker does not create or use it directly. However, my larger point is still accurate in the sense that it has too much important information and should not be overwritten. The hooks are _not_ a part of the programming model, so it would not be a case of the programming model overriding its own context.

PR #609, opened 2022-07-15
[DoNotMerge] Split programming model out of worker (Example framework PR)

ejizba:
See description in this PR first: https://github.com/Azure/azure-functions-nodejs-worker/pull/608

This PR represents what will be in the new `Azure/azure-functions-nodejs-framework` repo, but I wanted to get your approval on the name first. Once that repo is created, I will close this PR and re-open it there.

**Reasoning on the name**:
First, I was inspired by the Python repo [`Azure/azure-functions-python-library`](https://github.com/Azure/azure-functions-python-library). Then, I researched the differences between "sdk", "library", and "framework" because I've heard those used interchangeably. I landed on "framework" based on blogs like [this](https://www.interviewbit.com/blog/framework-vs-library):

> By using a library, you control the flow of the program. The library can be invoked whenever and wherever you like. Contrary to this, when you use a framework, the flow is controlled by the framework. The framework instructs you where to put your code, but it will call your code as required. Simply put, our code calls the library’s code, but in a framework, it’s the framework’s code that calls our code.

Since Azure Functions controls when the user's code is executed, I think "framework" is the most accurate.

Related to https://github.com/Azure/azure-functions-nodejs-worker/issues/568


ejizba:
Per offline discussion we landed on `azure-functions-nodejs-library`, primarily to be consistent with existing [python](https://github.com/Azure/azure-functions-python-library) and [java](https://github.com/Azure/azure-functions-java-library) repos

Here is the new, real PR: https://github.com/Azure/azure-functions-nodejs-library/pull/1

PR #610, opened 2022-07-18
Azure Function Node.JS github hosted website.

EIonv:
https://ae-001-1001.github.io/azure-functions-nodejs-worker/
#This is a github hosted website, though it looks functionless, You could theoretically add more. 
**A**.Maybe this is completely pointless, but for small business anything works. 
**B**.Even free long url, but completely customizable website for a developer to tinker with html,elements,css,etc.
**C**.The root problem for most/many websites they leave important code that someone can simply at some point decompile and find data that could possibly compromise said website.
using http.client requests using python can find certain things like server,content-length,The encoding it is in, the name of that .css file, (F12 Inspect Element/Network/Files shown below.)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,11 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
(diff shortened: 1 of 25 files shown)


EIonv (commented):
Please consider looking at the above, I don't know how to delete this. I am sorry for double.

PR #621, opened 2022-08-22
Make `hookData` and `appHookData` properties readonly

### LINKED ISSUE #601: How should hooks share data?

hossam-nasr:
Resolves #601 . See discussion there for more details and context

### DIFF
--- a/src/utils/ReadOnlyError.ts
+++ b/src/utils/ReadOnlyError.ts
@@ -1,10 +1,9 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License.
 
-import { InternalException } from './InternalException';
-
-export class ReadOnlyError extends InternalException implements TypeError {
+export class ReadOnlyError extends TypeError {
+    isAzureFunctionsInternalException = true;
     constructor(propertyName: string) {
-        super(`Cannot assign to read-only property '${propertyName}'`);
+        super(`Cannot assign to read only property '${propertyName}'`);
     }
 }
--- a/src/utils/ReadOnlyError.ts
+++ b/src/utils/ReadOnlyError.ts
@@ -5,6 +5,6 @@ import { InternalException } from './InternalException';
 
 export class ReadOnlyError extends InternalException implements TypeError {
     constructor(propertyName: string) {
-        super(`Attempting to set readonly property ${propertyName}`);
+        super(`Cannot assign to read-only property '${propertyName}'`);
     }
 }
--- a/src/utils/ReadOnlyError.ts
+++ b/src/utils/ReadOnlyError.ts
@@ -3,7 +3,7 @@
 
 import { InternalException } from './InternalException';
 
-export class ReadOnlyException extends InternalException {
+export class ReadOnlyError extends InternalException implements TypeError {
     constructor(propertyName: string) {
         super(`Attempting to set readonly property ${propertyName}`);
     }
--- a/src/utils/ReadOnlyException.ts
+++ b/src/utils/ReadOnlyException.ts
@@ -1,0 +1,10 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License.
+
+import { InternalException } from './InternalException';
+
+export class ReadOnlyException extends InternalException {
+    constructor(propertyName: string) {
+        super(`Attempting to set readonly property ${propertyName}`);
+    }
+}
--- a/test/eventHandlers/InvocationHandler.test.ts
+++ b/test/eventHandlers/InvocationHandler.test.ts
@@ -781,13 +781,13 @@ describe('InvocationHandler', () => {
                     context.hookData = {
                         foo: 'bar',
                     };
-                }).to.throw(`Cannot assign to read-only property 'hookData'`);
+                }).to.throw(`Cannot assign to read only property 'hookData'`);
                 expect(() => {
                     // @ts-expect-error: setting readonly property
                     context.appHookData = {
                         foo: 'bar',
                     };
-                }).to.throw(`Cannot assign to read-only property 'appHookData'`);
+                }).to.throw(`Cannot assign to read only property 'appHookData'`);
                 hookData += 'pre';
             })
         );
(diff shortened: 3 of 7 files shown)


src/utils/ReadOnlyException.ts
```diff
@@ -0,0 +1,10 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License.
+
+import { InternalException } from './InternalException';
+
+export class ReadOnlyException extends InternalException {
```
ejizba: `Exception` is more of a C# word AFAIK and `InternalException` just hasn't been renamed properly yet. Let's use `Error` moving forward

src/utils/ReadOnlyException.ts
```diff
@@ -0,0 +1,10 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License.
+
+import { InternalException } from './InternalException';
```
ejizba: Let's try to use (or at least mimic) native javascript readonly fields. For example, I did some digging and you can make an existing property on an object readonly like this:
```
Object.defineProperty(appStartContext, 'appHookData', { writable: false });
```

The error is of type `TypeError` and the message is something like this:

> Cannot assign to read only property 'hookData'...

src/utils/ReadOnlyException.ts
```diff
@@ -0,0 +1,10 @@
...
+
+export class ReadOnlyException extends InternalException {
+    constructor(propertyName: string) {
+        super(`Attempting to set readonly property ${propertyName}`);
```
ejizba: I noticed you do this in the other PR and I don't know where you got this pattern of starting errors with a present tense "ing" verb. I generally see "Can't do this" or "This is not allowed" instead of "Trying to do this"

src/utils/ReadOnlyException.ts
hossam-nasr: Were you able to reproduce this error message? In my testing, I saw that setting a property to readonly will only cause it to ignore set statements, but I didn't see an error being thrown. Anyway, I can use that message in my error

src/utils/ReadOnlyException.ts
hossam-nasr: I can't remember off the top of my head tbh, but I feel like I've seen that "ing" pattern before too. Anyway, I don't feel strongly about it, I renamed the error message to the one you had above

src/utils/ReadOnlyException.ts
hossam-nasr: Renamed to `Error`

src/utils/ReadOnlyError.ts
```diff
@@ -0,0 +1,10 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License.
+
+import { InternalException } from './InternalException';
+
+export class ReadOnlyError extends InternalException implements TypeError {
```
ejizba: This should be `extends TypeError` so that if people do `err instanceof TypeError` it'll return `true`. It's a built-in JavaScript/Node.js class to be used "when attempting to modify a value that cannot be changed" per [these docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypeError)

Since a class can only extend one other class you might have to copy the `isAzureFunctionsInternalException` property, but I think that's fine if it means we use standard Node.js classes.

src/utils/ReadOnlyException.ts
ejizba: Yes I was able to reproduce. I believe this error is only thrown if you're running in `strict` mode, but I thought TypeScript output `"use strict";` in all our js files by default

Anyways I don't have a strong preference as long as we're exactly replicating node.js behavior

hossam-nasr: Ah that makes sense. I was only testing on the node CLI

src/utils/ReadOnlyError.ts
hossam-nasr: Yeah I was trying to `extend` first but because I can only extend one class I did the `implements` instead. Changed to your suggestion

test/eventHandlers/InvocationHandler.test.ts
```diff
@@ -769,6 +769,60 @@ describe('InvocationHandler', () => {
...
+                    context.hookData = {
+                        foo: 'bar',
+                    };
+                }).to.throw(`Cannot assign to read-only property 'hookData'`);
+                expect(() => {
+                    // @ts-expect-error: setting readonly property
+                    context.appHookData = {
+                        foo: 'bar',
+                    };
+                }).to.throw(`Cannot assign to read-only property 'appHookData'`);
```
ejizba: Update these messages to fix the tests

ejizba (approved):

PR #623, opened 2022-08-23
Add app termination hooks

### LINKED ISSUE #606: Add app teardown execution hooks
After the host supports the "worker_terminate" gRPC message (https://github.com/Azure/azure-functions-host/issues/2308 and https://github.com/Azure/azure-functions-host/pull/8385), we should implement a handler for the worker_terminate message, allow app teardown executions hooks, and gracefully shut down the worker

hossam-nasr:
Resolves #606 now that https://github.com/Azure/azure-functions-host/pull/8385 is merged.

This PR adds a handler for the `worker_terminate` gRPC message to gracefully shut down the language worker, and introduces the new `appTerminate` hook. The name is modeled after `appStart` hook and the `workerTerminate` message, but ofc other ideas and suggestions are welcome.


test/eventHandlers/terminateWorker.test.ts
```diff
@@ -0,0 +1,190 @@
...
+        expect(streamEndStub.callCount).to.be.equal(1);
+    });
+
+    it('shuts down worker process', async () => {
+        stream.addTestMessage(Msg.workerTerminate());
+        await stream.assertCalledWith(Msg.receivedWorkerTerminateLog);
+        expect(processExitStub.calledWith(0)).to.be.true;
+    });
+
+    it('runs app termiante hooks', async () => {
```
ejizba:
```suggestion
    it('runs app terminate hooks', async () => {
```

test/eventHandlers/terminateWorker.test.ts
```diff
@@ -0,0 +1,190 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License.
+
+import * as coreTypes from '@azure/functions-core';
```
ejizba, line 4: The amount of unit test code you wrote was about 4-5x the amount of feature code 😅 I'm glad to see tests, but this ratio feels off. I won't block for this PR, but if the pattern continues I will block future PRs.

Some general guidelines
- The time spent should be proportional to the complexity and importance of the feature itself
- Most features deserve at least 1 or 2 basic smoke tests, but not all
- For edge case tests, try to test patterns unique to that feature, as opposed to small variations on the same pattern already covered by tests for other features
- If it is worth testing small variations, set up parameterized tests instead of copy/pasting tests

src/parsers/secondsFromDuration.ts
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License.
+
+import { google } from '../../azure-functions-language-worker-protobuf/src/rpc';
```
ejizba: I would prefer if we remove the gracePeriod property for now. You can still keep the type if you want, though:
```
type AppTerminateContext = HookContext;
```

A few reasons:
1. I think people will make the most of whatever time they are given, regardless of what that time is
2. I don't think the value will be consistent or accurate. The time period can be affected by operations we perform as the worker, other processes on the VM, slow calls over the internet, etc. Plus the host could always just terminate early
3. I think `secondsFromDuration` is the most complex (or at least unique) code in your PR and the most susceptible to bugs. I'll refrain from adding comments on it until we decide what to do with `gracePeriod`, though

test/eventHandlers/terminateWorker.test.ts
hossam-nasr: Sounds good, thanks for the guidelines

test/eventHandlers/terminateWorker.test.ts
hossam-nasr: Fixed, thanks

src/parsers/secondsFromDuration.ts
hossam-nasr: That's fair. Although the same arguments could be made for the host passing down that property to the worker in the first place, but it still does. This was the reason I included it in the context. To your point about `secondsFromDuration`, I did have some unit tests for it, but I could also just pass down the `Duration` object as-is if that feels better

ejizba: I don't think the host should pass it down to the worker either haha

hossam-nasr: Ok I removed it

ejizba (approved):

PR #624, opened 2022-08-25
Clean up error naming

ejizba:
- Use `Error` instead of `Exception` because it's more common in Node.js
- Use `System` instead of `Internal` to match the "user"/"system" names that the host uses to differentiate which logs we can or can't track in our own telemetry
- Fix a few straggling errors that should be system errors


src/errors.ts
```diff
@@ -0,0 +1,54 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License.
+
+export interface AzFuncError {
+    /**
+     * System errors can be tracked in our telemetry
+     * User errors cannot be tracked in our telemetry because they could have user information (users can still track it themselves in their app insights resource)
+     */
+    isAzureFunctionsSystemError: boolean;
```
hossam-nasr, line 9: Now that this property is renamed, is this property used by the host at all or just in our codebase? I don't think it is and I couldn't find `isAzureFunctionsInternalException` searching in the host repo, but just making sure

src/errors.ts
```diff
@@ -0,0 +1,54 @@
...
+    }
+}
+
+/**
+ * This is mostly for callbacks where `null` or `undefined` indicates there is no error
+ * By contrast, anything thrown/caught is assumed to be an error regardless of what it is
+ */
+export function isError(err: unknown): boolean {
+    return err !== null && err !== undefined;
+}
```
hossam-nasr, line 54: Thoughts on taking this file out of `utils/` and putting it in root `src/`? Feels like it should still belong in `src/utils/` to me

src/errors.ts
ejizba: This is just used in our codebase, not the host. This is the main place it's used:

https://github.com/Azure/azure-functions-nodejs-worker/blob/v3.x/src/setupEventStream.ts#L97

src/errors.ts
ejizba: I feel like this file is more similar to `constants.ts` (which we put at the root) than stuff in `utils`. Also I just don't like the utils folder and try to avoid it when something else might make sense

hossam-nasr (approved):

PR #631, opened 2022-09-28
Treat failure to load entry point as warning instead of error

### LINKED ISSUE #630: Worker was unable to load entry point
With the rollout of host version 4.12 users can see this error:

> Worker was unable to load entry point "index.js": Found zero files matching the supplied pattern

NOTE: The filename in the message may be different, but "index.js" is by far the most common.

### Mitigation

Update: The change was rolled back in host v4.12.2.19454. No further mitigation should be necessary.

### Background

We introduced a new feature in https://github.com/Azure/azure-functions-nodejs-worker/issues/537 that lets users run app-level entrypoint code. By default we would load the file specified in a user's package.json "main" field and throw an error if we have any issues loading the file specified. Unfortunately, many existing users appear to have "main" set to "index.js" but that file does not exist in their app. When we started rolling out this feature, that means their app would fail to start.

ejizba:
Fixes https://github.com/Azure/azure-functions-nodejs-worker/issues/630

Most of the apps we saw with an error had "main" set to index.js but that file didn't exist. We don't don't want to break those existing users, so we will convert the error to a warning. ~~However, we still want to throw an error in the new programming model (which obviously doesn't have any existing users) - so we will use the existence of "function.json" files to determine which model is being used and act accordingly.~~ EDIT: Changed our mind on this.

### DIFF
--- a/src/startApp.ts
+++ b/src/startApp.ts
@@ -64,11 +64,45 @@ async function loadEntryPointFile(functionAppDirectory: string, channel: WorkerC
             }
         } catch (err) {
             const error = ensureErrorType(err);
-            channel.log({
-                message: `Worker was unable to load entry point "${entryPointPattern}": ${error.message}`,
-                level: LogLevel.Warning,
-                logCategory: LogCategory.System,
-            });
+            const message = `Worker was unable to load entry point "${entryPointPattern}": ${error.message}`;
+
+            // If this is an old existing app, we can't throw an error about the entrypoint for backwards compat reasons
+            // More info here: https://github.com/Azure/azure-functions-nodejs-worker/issues/630
+            // However, if we determine this is the new programming model, we will do the "proper" thing and throw an error
+            if (await isNewProgrammingModel(functionAppDirectory)) {
+                channel.log({
+                    message:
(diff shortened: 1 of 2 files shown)


AnatoliB (approved):

src/startApp.ts
```diff
@@ -64,9 +64,45 @@ async function loadEntryPointFile(functionAppDirectory: string, channel: WorkerC
...
     }
 }
+
+/**
+ * Best effort to determine if this is the new programming model based on the existence of "function.json" files
+ * Defaults to false if we can't figure it out
+ */
+async function isNewProgrammingModel(functionAppDirectory: string): Promise<boolean> {
+    try {
+        const files = await globby('*/function.json', {
```
alrod: Are we scanning all the folders? I am wondering about coldstart impact if there are a lot functions. I would check this with Hamid

ejizba:
Changing back so that all errors are treated as warnings for the entry point scenario. It's simpler and much less risky - discussed in further detail offline.

hossam-nasr (approved):

src/startApp.ts
ejizba: Should not have had a big impact, but regardless discussed offline and I think a simpler solution is better anyways.

PR #634, opened 2022-10-05
Handle multiple workers

ejizba:


hossam-nasr (commented):
This makes sense, but don't we need to update the default version of `@azure/functions` shipped in Azure to be one where a valid `functionId` is being sent to the worker before we can merge this with the host?

ejizba:
This only applies to v4 of the library, which is not the default

hossam-nasr (approved):

hossam-nasr:
Created an issue here: https://github.com/Azure/azure-functions-nodejs-library/issues/28 to track this for v4 to make sure we don't forget and also approved :)

ejizba:
Closing in favor of https://github.com/Azure/azure-functions-nodejs-worker/pull/641

Issue #638, opened 2022-10-11
Support multiple worker scenario in new framework


ejizba:
Users can get into this situation by explicitly setting the [FUNCTIONS_WORKER_PROCESS_COUNT](https://learn.microsoft.com/azure/azure-functions/functions-app-settings#functions_worker_process_count) app setting to 2 (or more).

It can also happen as a part of the intelligent worker concurrency feature that is currently rolling out. This feature is meant to help with performance and because there is no strict rule on _when_ it will introduce new workers, it's a major blocker for the new framework in azure because we don't want random failures.

If you try [the new framework](https://github.com/Azure/azure-functions-nodejs-library/wiki/Azure-Functions-Node.js-Framework-v4), it just does not work with multiple workers. A few problems:
1. The host expects the same metadata for each worker. Meaning if the second worker has a different function id or binding id, our worker fails to find that function or binding. Proposed fix is to make these ids deterministic
2. The host sends requests to the second worker in a weird order. Tracking here: https://github.com/Azure/azure-functions-host/issues/8825. Hopefully host can fix this

## Workaround

As a workaround, users can explicitly set `FUNCTIONS_WORKER_PROCESS_COUNT` to `1` in their app settings

ejizba:
Addressed in https://github.com/Azure/azure-functions-nodejs-worker/pull/641 and https://github.com/Azure/azure-functions-host/pull/8900

PR #641, opened 2022-10-20
Support multiple workers with new indexing

ejizba:
Related to https://github.com/Azure/azure-functions-nodejs-worker/issues/638

This PR does a few things:
1. Always include worker id when logging that we received a request. This makes debugging the "multiple worker" scenario so much easier
2. Use function name as function id instead of a random guid. This ensures each worker returns the same thing
3. Allow the library package to specify both the name and id. This just gives us more flexibility to fix things on the library-side instead of worker-side if necessary down the line
4. Adjust the worker indexing booleans on `WorkerChannel` for two reasons:
    a. To account for the fact that only the 1st worker gets the FunctionsMetadataRequest
    b. To more explicitly separate the worker-indexing and legacy scenarios and avoid any mingling

### DIFF
--- a/src/eventHandlers/FunctionLoadHandler.ts
+++ b/src/eventHandlers/FunctionLoadHandler.ts
@@ -6,6 +6,8 @@ import { ensureErrorType } from '../errors';
 import { nonNullProp } from '../utils/nonNull';
 import { WorkerChannel } from '../WorkerChannel';
 import { EventHandler } from './EventHandler';
+import LogCategory = rpc.RpcLog.RpcLogCategory;
+import LogLevel = rpc.RpcLog.Level;
 
 /**
  * Worker responds after loading required metadata to load function with the load result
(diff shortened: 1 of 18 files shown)


src/eventHandlers/FunctionLoadHandler.ts
```diff
@@ -18,11 +20,19 @@ export class FunctionLoadHandler extends EventHandler<'functionLoadRequest', 'fu
...
-        const functionId = nonNullProp(msg, 'functionId');
-        const metadata = nonNullProp(msg, 'metadata');
-        if (!channel.functions[functionId]) {
+        channel.log({
+            message: `Worker ${channel.workerId} received FunctionLoadRequest`,
+            level: LogLevel.Debug,
+            logCategory: LogCategory.System,
+        });
+
+        if (!channel.isUsingWorkerIndexing) {
```
hossam-nasr, line 33: I'm slightly worried about this if condition here. Is there a possible scenario in which a user is using the old indexing, but still `registerFunction()` gets called somehow, setting `channel.isUsingWorkerIndexing` to true? It feels safer to keep the old check `!channel.functions[functionId]`

ejizba: The user would have to explicitly call `registerFunction()` either from the core api or v4 of the library, which I can't see happening on accident. Our general position has been that we don't support mixing and matching the new & old models and my change takes a stricter stance to prevent that. It was very helpful as I was debugging the multi-worker scenario and should be helpful moving forward for customers as well. For example, it results in clearer error messages catered to each model when things aren't working quite right.

hossam-nasr (approved):

PR #657, opened 2023-02-17
Remove workerId from error logs

ejizba:
I've noticed a few situations when troubleshooting an app where it can be difficult to get the count of a specific error because we have the worker id in the message. You'll end up with something like this:

Screenshot 2023-02-17 at 3 26 32 PM

The worker id is already tracked in the "Source" field in telemetry, so we don't really need it in the message.

I could've removed the workerId from debug and info logs, too, but I think they're fine as-is. You never really need to collate those messages like you would error logs.

### DIFF
--- a/src/Worker.ts
+++ b/src/Worker.ts
@@ -57,9 +57,9 @@ export function startNodeWorker(args) {
         const error = ensureErrorType(err);
         let errorMessage: string;
         if (error.isAzureFunctionsSystemError) {
-            errorMessage = `Worker ${workerId} uncaught exception: ${error.stack || err}`;
+            errorMessage = `Worker uncaught exception: ${error.stack || err}`;
         } else {
-            errorMessage = `Worker ${workerId} uncaught exception (learn more: https://go.microsoft.com/fwlink/?linkid=2097909 ): ${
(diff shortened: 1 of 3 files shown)


hossam-nasr:
Are we losing important information here? I imagine being able to narrow down the errors coming from a specific worker instance could still be useful in some scenarios, especially given the new language worker concurrency feature. On the other hand, the scenario you're describing can still be achieved by a simple `| parse` line, albeit more verbosely. Is this worker id information still preserved elsewhere in the logs? I did some minimal testing and couldn't find it.

ejizba:
> Are we losing important information here?

There's a "Source" field in telemetry that has the worker id, so no we're not losing anything in Azure. The source field isn't displayed when running locally (AFAIK, there might be a way), but I don't think the multiple worker scenario is important enough when running locally to matter.

Screenshot 2023-03-02 at 10 10 08 AM

> On the other hand, the scenario you're describing can still be achieved by a simple | parse line, albeit more verbosely

The goal of this is to help with queries that display lists of all types of errors, which inherently means you don't want to special-case one specific error message. For example, I do this to see if there's a spike in any particular message when a new worker is rolling out. Also, the built-in detectors on applense don't use a "parse"

hossam-nasr (approved):
Sounds good :)

PR #659, opened 2023-02-18
Preemptively unblock Node 20

ejizba:
The sooner the better!

v20 will be based off of v19, so I think it's good enough to use v19 for our CI testing for now

Related to https://github.com/Azure/azure-functions-nodejs-library/issues/170

### DIFF
--- a/test/eventHandlers/WorkerInitHandler.test.ts
+++ b/test/eventHandlers/WorkerInitHandler.test.ts
@@ -6,6 +6,7 @@ import * as escapeStringRegexp from 'escape-string-regexp';
 import 'mocha';
 import { ITestCallbackContext } from 'mocha';
 import * as mockFs from 'mock-fs';
+import * as semver from 'semver';
 import { AzureFunctionsRpcMessages as rpc } from '../../azure-functions-language-worker-protobuf/src/rpc';
 import { logColdStartWarning } from '../../src/eventHandlers/WorkerInitHandler';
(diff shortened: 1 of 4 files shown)


test/eventHandlers/WorkerInitHandler.test.ts
```diff
@@ -218,14 +219,18 @@ describe('WorkerInitHandler', () => {
             },
         });
 
+        const jsonError = semver.gte(process.versions.node, '19.0.0')
```
hossam-nasr, line 222: I support the idea of this PR to preemptively support node 20 as much as possible. But I worry, are we adding additional unnecessary burden here by using node 19 as a proxy for node 20? Are we forcing ourselves to support features/updates in node 19 that may not end up making it to node 20? This change is an example. Are we sure that node 20 will still have this different behavior?

ejizba: Well...it took me all of 10 minutes to fix this unit test. Can you expand on what you mean by burden?

For the record, this PR doesn't mean we actually support anything. Users have to wait for official public preview/GA announcements.

hossam-nasr: Well, I'm just saying in general. I'm not sure what other changes in node 19 that might require us to update/modify tests that won't be necessary in node 20. If you don't think this is a significant risk, I'm cool merging this and won't block on it

hossam-nasr (approved):

PR #660, opened 2023-02-22
Warn users with long load times

ejizba:
and tell them to use WEBSITE_RUN_FROM_PACKAGE

This is one of the most common problems I see on incidents. As an example, for one incident we decreased the load time from 96 seconds to 90 milliseconds after adding WEBSITE_RUN_FROM_PACKAGE.

### DIFF
--- a/src/loadScriptFile.ts
+++ b/src/loadScriptFile.ts
@@ -1,24 +1,56 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License.
 
+import * as path from 'path';
 import * as url from 'url';
+import { AzureFunctionsRpcMessages as rpc } from '../azure-functions-language-worker-protobuf/src/rpc';
 import { AzFuncSystemError } from './errors';
 import { PackageJson } from './parsers/parsePackageJson';
+import { WorkerChannel } from './WorkerChannel';
+import LogCategory = rpc.RpcLog.RpcLogCategory;
+import LogLevel = rpc.RpcLog.Level;
 
-export async function loadScriptFile(filePath: string, packageJson: PackageJson): Promise<unknown> {
-    let script: unknown;
-    if (isESModule(filePath, packageJson)) {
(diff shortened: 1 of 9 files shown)


src/loadScriptFile.ts
```diff
@@ -1,24 +1,56 @@
...
+    } finally {
+        warnIfLongLoadTime(channel, filePath, start);
+    }
+}
+
+function warnIfLongLoadTime(channel: WorkerChannel, filePath: string, start: number): void {
+    const timeElapsed = Date.now() - start;
+    const rfpName = 'WEBSITE_RUN_FROM_PACKAGE';
+    const rfpValue = process.env[rfpName];
+    if (timeElapsed > 1000 && (rfpValue === undefined || rfpValue === '0')) {
```
hossam-nasr, line 42: Should we always give this warning regardless of how much time it takes? For example, there could be a scenario where the app has tons of files, each file takes less than 1 second, but collectively they still add up to a significant latency?

ejizba: The main downside of rfp is that it makes your app read-only and you can't edit it on the portal. If people are editing on the portal, they are likely to be newer users with smaller apps who would not have long load times. There's something to be said for preventing noise and only showing warnings that we know are a problem.

> For example, there could be a scenario where the app has tons of files, each file takes less than 1 second, but collectively they still add up to a significant latency?

We don't call `loadScriptFile` on every file in the app, just the entrypoint/function files. Any sub-imports are included in the load time of the original files. It's common to have A LOT of sub-imports (through npm modules or whatever), but I can't see people having enough entrypoint/function files to be a problem.

ejizba: I would prefer to stick to 1 second for now, but depending how successful/annoying this warning is I would consider reducing the threshold down the line

hossam-nasr: Okay, I won't block on this. Your reasoning sounds good to me

hossam-nasr (approved):

Issue #670, opened 2023-03-21
Reset all app data on WorkerChannel during env reload


ejizba:
For placeholders, we're currently relying on the fact that the default dummy app is not a Node.js app (I think it's a csx app). Theoretically if the dummy app was a Node.js app (especially a v4 app), some of its data (like hooks, registered functions, etc.) could bleed over into the real customer's app after a func env reload.

We should consider a clean reset of all app-related data just to be safe. Also if we reset all app-related data, we don't have to worry about forgetting to reset a random boolean like in #669.

PR #671, opened 2023-03-21
Fix registerFunction for placeholders when indexing is on by default

### LINKED ISSUE #669: Register function fails on placeholders when indexing is on by default
Here's what happens:
1. Host starts worker in placeholder mode (aka with a dummy app)
2. Host asks worker to index. Worker finds nothing to index in the dummy app and sets an internal boolean basically saying "indexing done"
3. Host sends reload request with customer's actual app
4. Worker reloads, but fails to load the user's app with error "A function can only be registered during app startup." because our internal boolean said we already indexed

NOTE: This is only a problem if indexing is on by default because step 2 would be skipped if it's off by default.

ejizba:
Fixes https://github.com/Azure/azure-functions-nodejs-worker/issues/669

I honestly spent more time on the test than the fix in `src`, but I feel it's important to have a test to repro an impactful bug like this. Here's proof that the test will fail without my fix in `src`:
Screenshot 2023-03-21 at 10 59 29 AM

Also, I wanted to keep this change scoped as small as possible to get it out asap and keep the risk minimal. I feel like there's more we could do here, but left that work for a separate issue: https://github.com/Azure/azure-functions-nodejs-worker/issues/670

### DIFF
--- a/src/WorkerChannel.ts
+++ b/src/WorkerChannel.ts
@@ -59,6 +59,13 @@ export class WorkerChannel {
         this.packageJson = {};
     }
 
+    reset(): void {
+        // Ideally this resets all app-related data
+        // That worked is tracked by https://github.com/Azure/azure-functions-nodejs-worker/issues/670
+        this.workerIndexingLocked = false;
+        this.isUsingWorkerIndexing = false;
+    }
+
     /**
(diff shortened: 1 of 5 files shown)


hossam-nasr (approved):
LGTM

PR #679, opened 2023-04-14
Make workerBitness consistent

### LINKED ISSUE #633: Fix worker metadata format
We'd like to be more strict about format for the analytics team.

We currently see this in the metrics table:

```
"RuntimeName": node,
"RuntimeVersion": v16.16.0,
"WorkerVersion": 3.5.0,
"WorkerBitness": ia32
```

Is it possible to remove the version prefix ("v") from runtimeVersion?
We also noticed the bitness looks wrong: "ia32" which should be "x86"

ejizba:
Fixes https://github.com/Azure/azure-functions-nodejs-worker/issues/633

### DIFF
--- a/test/eventHandlers/msg.ts
+++ b/test/eventHandlers/msg.ts
@@ -21,7 +21,7 @@ function stackTraceRegExpProps(responseName: string, message: string): RegExpPro
 function workerMetadataRegExps(responseName: string) {
     return {
         [`${responseName}.workerMetadata.runtimeVersion`]: /^[0-9]+\.[0-9]+\.[0-9]+$/,
-        [`${responseName}.workerMetadata.workerBitness`]: /^(x64|ia32|arm64)$/,
+        [`${responseName}.workerMetadata.workerBitness`]: /^(x64|x86|arm64)$/,
         [`${responseName}.workerMetadata.workerVersion`]: /^3\.[0-9]+\.[0-9]+$/,
         [`${responseName}.workerMetadata.customProperties.modelVersion`]: /^3\.[0-9]+\.[0-9]+$/,
     };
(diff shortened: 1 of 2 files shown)


test/eventHandlers/msg.ts
```diff
@@ -21,7 +21,7 @@ function stackTraceRegExpProps(responseName: string, message: string): RegExpPro
 function workerMetadataRegExps(responseName: string) {
     return {
         [`${responseName}.workerMetadata.runtimeVersion`]: /^[0-9]+\.[0-9]+\.[0-9]+$/,
-        [`${responseName}.workerMetadata.workerBitness`]: /^(x64|ia32|arm64)$/,
+        [`${responseName}.workerMetadata.workerBitness`]: /^(x64|x86|arm64)$/,
```
hossam-nasr, line 24: I saw the link here that you posted earlier: https://nodejs.org/api/process.html#processarch and it seems like there are more possible values than just `x64`, `x86` and `arm64`. Are those the only ones that we expect Functions to be run on? Relatedly, is `ia32` the only one that needs to be "normalized"?

hossam-nasr (approved):
I left one question for my own curiosity but I trust you know what you're doing here so I won't block on this :)

test/eventHandlers/msg.ts
ejizba: ia32 is the only one they asked about in the github issue, so I figured that's the only one I'd do instead of trying to predict what they want. Ultimately I think we only care about the standard arch's used in Azure, x86 and x64. I included arm64 in the unit tests just in case someone with an M1 mac clones our repo and wants a successful test run, but I don't think it's used in azure.

hossam-nasr: Makes sense to me!

PR #680, opened 2023-04-14
Reset all app-level data during environment reload

### LINKED ISSUE #670: Reset all app data on WorkerChannel during env reload

ejizba:
Fixes https://github.com/Azure/azure-functions-nodejs-worker/issues/670

### DIFF
--- a/src/WorkerChannel.ts
+++ b/src/WorkerChannel.ts
@@ -1,26 +1,22 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License.
 
-import { FunctionCallback, HookCallback, HookContext, HookData, ProgrammingModel } from '@azure/functions-core';
+import { HookCallback, HookContext, ProgrammingModel } from '@azure/functions-core';
 import { AzureFunctionsRpcMessages as rpc } from '../azure-functions-language-worker-protobuf/src/rpc';
+import { AppContext } from './AppContext';
 import { Disposable } from './Disposable';
 import { IEventStream } from './GrpcClient';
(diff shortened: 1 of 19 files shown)


src/WorkerChannel.ts
```diff
@@ -34,36 +30,14 @@ export class WorkerChannel {
...
     }
 
-    reset(): void {
-        // Ideally this resets all app-related data
-        // That worked is tracked by https://github.com/Azure/azure-functions-nodejs-worker/issues/670
-        this.workerIndexingLocked = false;
-        this.isUsingWorkerIndexing = false;
+    resetApp(): void {
+        this.app = new AppContext();
+        this.app.programmingModel = this.defaultProgrammingModel;
```
hossam-nasr, line 40: What if we make the `AppContext` constructor accept a `ProgrammingModel` argument and this line can be moved there? Makes logical sense to me to have such logic in a constructor.

hossam-nasr (approved):
Left one non-blocking suggestion, but lgtm:)

src/WorkerChannel.ts
ejizba: That would work for this case (environment reload), but for the other case (worker init) we don't have the defaultProgrammingModel yet when creating the AppContext

PR #687, opened 2023-05-18
Move code out of WorkerChannel and setupCoreModule

ejizba:
No functional changes, just moving code

### DIFF
--- a/src/setupCoreModule.ts
+++ b/src/setupCoreModule.ts
@@ -1,15 +1,13 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License.
 
-import { FunctionCallback, FunctionMetadata, HookCallback, ProgrammingModel } from '@azure/functions-core';
-import { AzureFunctionsRpcMessages as rpc } from '../azure-functions-language-worker-protobuf/src/rpc';
 import { Disposable } from './Disposable';
 import { channel } from './WorkerChannel';
 import { version } from './constants';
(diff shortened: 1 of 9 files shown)


src/setupCoreModule.ts
```diff
@@ -22,26 +20,12 @@ export function setupCoreModule(): void {
...
-        },
+        registerHook,
+        setProgrammingModel,
         getProgrammingModel: () => {
             return channel.app.programmingModel;
         },
-        registerFunction: (metadata: FunctionMetadata, callback: FunctionCallback) => {
-            return registerFunction(metadata, callback);
-        },
+        registerFunction,
```
hossam-nasr, line 28: Was there a reason this second layer of indirection added here? Maybe something about calling `this`?

hossam-nasr (approved):
Left one small comment, but lgtm! I guess this is another benefit of making channel global

hossam-nasr (approved):
Left one small comment, but lgtm! I guess this is another benefit of making channel global

src/setupCoreModule.ts
ejizba: I had the indirection because I had to pass `channel` in, which was fixed when I made `channel` global: https://github.com/Azure/azure-functions-nodejs-worker/pull/686/files#diff-13b181030e6532174ea055646adc264a6263de35e2336dbeb25767e3ee59eabbL43

You can run into problems related to `this`, but it should not apply here.

PR #689, opened 2023-05-23
Switch to new end-to-end tests

### LINKED ISSUE #76: Add end-to-end tests for v4
Here's the tests we have for the v3 model: https://github.com/Azure/azure-functions-nodejs-worker/tree/v3.x/test/end-to-end

We need at least the same thing for the v4 model, although wouldn't hurt to upgrade them to something better. A few things I feel like we could improve:
1. The tests take forever (over an hour)
2. The tests fail intermittently if you have multiple builds at the same time
3. The tests are written in C#. This makes it easier to share tests across workers, but discourages us Node.js folks from improving our own e2e tests

ejizba:
Tests were redesigned and moved [here](https://github.com/azure/azure-functions-nodejs-e2e-tests). You can see an example test run [here](https://azfunc.visualstudio.com/Azure%20Functions/_build/results?buildId=131570&view=results). I will update Francisco's integration tests pipeline (which schedules a test run daily) to reference the new tests once this PR is approved.

Highlights:
1. Tests v4 model for the first time
2. Takes ~15 minutes instead of ~65 minutes
3. Written in Node.js instead of C#
4. Much easier to run locally

Fixes https://github.com/Azure/azure-functions-nodejs-library/issues/76


azure-pipelines/test.yml
```diff
@@ -76,81 +76,3 @@ jobs:
       testResultsFiles: 'test/unit-test-results.xml'
       testRunTitle: '$(Agent.JobName)'
     condition: succeededOrFailed()
-
-- job: E2ETests
```
hossam-nasr, line 80: Where is this being replaced by the new tests?

ejizba: https://github.com/Azure/azure-functions-nodejs-e2e-tests/blob/main/azure-pipelines/test.yml

The caveat is that it's not run as often, which might be what you're asking about. I'm not running it for PRs - just on a daily schedule. I think e2e tests are excessive for PRs

hossam-nasr (approved):

PR #691, opened 2023-05-27
Make entry point failures blocking for v4

### LINKED ISSUE #85: Make v4 entry point failures more obvious
UPDATE by @ejizba: This change has been included in host 4.25. It is a *breaking change* for model v4 as some errors that were previously ignored will now block your app as an error.

### Workaround to get old behavior

You can set the app setting `FUNCTIONS_NODE_BLOCK_ON_ENTRY_POINT_ERROR` to `false` if you want the old behavior. However, we actually recommend setting this to `true` for all v4 apps. For more information, see the [App Settings reference](https://learn.microsoft.com/en-us/azure/azure-functions/functions-app-settings#functions_node_block_on_entry_point_error)
________

- Timestamp: 2023-04-17T20:18:05.2026993Z
- Function App name: (private)
- Function name(s) (as appropriate): (all)
- ~Invocation~ Host Instance ID: 16717bc2-cc29-4b43-815b-db2781f9ef3f
- Region: northcentralus

#### Repro steps
1. Write a simple node v4 functions app which throws an Error in the entry point.
2. Run the app locally; see that the error is show in the console, although the runtime ignores the error and proceeds.
3. Deploy the app to cloud Azure Functions with Application Insights; see that the error is unable to be found.

#### Expected behavior
The error message should have been exposed in Application Insights. Additionally, it would be helpful to expose the full exception object, not just the message.
As a potential configuration option, startup should fail completely if an entry point fails to execute.

#### Actual behavior
When running functions locally, any exception which is thrown while executing entry point files is exposed in the console. E.g. see the first line below:
```bash
Azure Functions Core Tools
Core Tools Version:       4.0.5095 Commit hash: N/A  (64-bit)
Function Runtime Version: 4.16.5.20396
[2023-04-17T20:02:47.709Z] Worker was unable to load entry point "dist/entryPoints/**/*.js": <THROWN EXCEPTION MESSAGE IS SHOWN HERE>
[2023-04-17T20:02:50.722Z] Workers Directory set to: C:\Program Files\Microsoft\Azure Functions Core Tools\workers
...
[2023-04-17T20:02:51.511Z] 0 functions loaded
...
[2023-04-17T20:02:51.682Z] No job functions found. Try making your job classes and methods public. If you're using binding extensions (e.g. Azure Storage, ServiceBus, Timers, etc.) make sure you've called the registration method for the extension(s) in your startup code (e.g. builder.AddAzureStorage(), builder.AddServiceBus(), builder.AddTimers(), etc.).
```

When deploying this same code to cloud Azure Functions with Application Insights configured, the first trace which is emitted is:
```
4/17/2023, 1:18:05 PM TRACE
Workers Directory set to: /azure-functions-host/workers
Severity level: Verbose
```

Doing a search for `Worker was unable to load entry point` in Application Insights returns no results.

#### Known workarounds
None

### LINKED ISSUE #697: Entry point errors should block app from running
Starting with Node.js 20, we are introducing a breaking change where entry point errors will block your function app from running. The most common error is this:

> Worker was unable to load entry point "index.js": File does not exist

## Workaround

> NOTE: If you're on Node.js <20, you can technically ignore the error and your app will work just fine, but we still recommend following the workaround to get rid of the error.

The entrypoint is specified by the "main" field in your `package.json`. For the most common error mentioned above, you can workaround it by removing the "main" field entirely from your "package.json", assuming you don't actually have an "index.js" file you want to use.

Most other errors will have a fix specific to your own code, which will need to be addressed for your app to properly run.

## Background

We introduced support for an entry point file in [Fall of 2022](https://github.com/Azure/azure-functions-nodejs-worker/releases/tag/v3.5.0), however at the time we made it a non-blocking warning to maintain backwards compatibility for apps that "worked" even though they had invalid entry point files. An entry point file is optional for programming model v3, but required for programming model v4 (Read more about the different models [here](https://learn.microsoft.com/azure/azure-functions/functions-node-upgrade-v4)). In the vast majority of cases, a v4 model app will not work if you have an entry point error, which is why we wanted to make these errors very obvious/blocking starting with Node 20.

ejizba:
And make this behavior the default starting with Node v20. For Node <v20, we'll provide an app setting `FUNCTIONS_NODE_BLOCK_ON_ENTRY_POINT_ERROR` to configure this.

There's still an edge case where you get an error before you've registered v4 as the programming model. Not much we can do about that for Node <20 because of [#630](https://github.com/Azure/azure-functions-nodejs-worker/issues/630), though

Fixes https://github.com/Azure/azure-functions-nodejs-library/issues/85
Fixes https://github.com/Azure/azure-functions-nodejs-worker/issues/697


hossam-nasr (dismissed):

src/startApp.ts
```diff
@@ -63,7 +65,16 @@ async function loadEntryPointFile(functionAppDirectory: string): Promise<void> {
...
             const files = await globby(entryPointPattern, { cwd: functionAppDirectory });
             if (files.length === 0) {
-                throw new AzFuncSystemError(`Found zero files matching the supplied pattern`);
+                let message: string = globby.hasMagic(entryPointPattern, { cwd: functionAppDirectory })
+                    ? 'Found zero files matching the supplied pattern'
+                    : 'File does not exist';
+
+                if (entryPointPattern === 'index.js') {
+                    // This is by far the most common error and typically happens by accident, so we'll give these folks a little more help
+                    message += '. Learn more here: https://aka.ms/AAla7et';
```
hossam-nasr, line 74: Can we have a more descriptive aka.ms link?

src/startApp.ts
```diff
@@ -88,13 +99,40 @@ async function loadEntryPointFile(functionAppDirectory: string): Promise<void> {
...
+
+function shouldBlockOnEntryPointError(): boolean {
+    if (isNode20Plus()) {
+        // Starting with Node 20, this will always be blocking
+        // https://github.com/Azure/azure-functions-nodejs-worker/issues/697
+        return true;
+    } else {
+        const key = 'FUNCTIONS_NODE_BLOCK_ON_ENTRY_POINT_ERROR';
+        if (isDefined(process.env[key])) {
+            return isEnvironmentVariableSet(process.env[key]);
```
hossam-nasr, line 135: I'm not super experienced with regex patterns, so what does `isEnvironmentVariableSet` test exactly? Can this be used in place of the `isDefined(process.env[key])` line entirely?

src/startApp.ts
```diff
@@ -88,13 +99,40 @@ async function loadEntryPointFile(functionAppDirectory: string): Promise<void> {
...
+        } else {
+            // We think this should be a blocking error by default, but v3 can't do that for backwards compatibility reasons
+            // https://github.com/Azure/azure-functions-nodejs-worker/issues/630
+            const model = nonNullProp(worker.app, 'programmingModel');
+            return !(model.name === '@azure/functions' && model.version.startsWith('3.'));
```
hossam-nasr, line 140: What happens in the case the model isn't set yet?

test/eventHandlers/WorkerInitHandler.test.ts
```diff
@@ -117,25 +117,125 @@ describe('WorkerInitHandler', () => {
...
+
+            it('init fails (v4)', async () => {
+                worker.app.programmingModel = <any>{ name: '@azure/functions', version: '4.0.0' };
+                await verifyInitFails();
+            });
+
+            it('init fails (v3) (app setting=1)', async () => {
+                process.env.FUNCTIONS_NODE_BLOCK_ON_ENTRY_POINT_ERROR = '1';
+                await verifyInitFails();
+            });
```
hossam-nasr: Is there a reason we're not testing the cases with v3 app setting = 0, and v4 app setting =1?

test/eventHandlers/WorkerInitHandler.test.ts
```diff
@@ -117,25 +117,125 @@ describe('WorkerInitHandler', () => {
...
+            );
+        }
 
-    it('Fails for invalid entry point', async function (this: ITestCallbackContext) {
-        const fileSubpath = await setTestAppMainField('throwError.js');
+        describe('Node >=v20', () => {
+            before(function (this: IHookCallbackContext) {
+                if (!isNode20Plus()) {
+                    this.skip();
+                }
```
hossam-nasr: Because it can get really messy/confusing otherwise, I'd be in favor of adding a line here to clear environment variables, and set the desired app setting explicitly in each test case

ejizba: env vars get cleared after each test here: https://github.com/Azure/azure-functions-nodejs-worker/blob/v3.x/test/eventHandlers/TestEventStream.ts#L83

test/eventHandlers/WorkerInitHandler.test.ts
ejizba: Because the app setting wouldn't change the behavior in that case

src/startApp.ts
ejizba: That _should_ be impossible - the default programming model is set before worker init

src/startApp.ts
ejizba: `isDefined` checks if it is `null` or `undefined`

`isEnvironmentVariableSet` handles string values as well. It will return `false` for strings like `0`, `false`, `FaLse`, and return `true` for basically anything else including `1`, `true`, `tRue`, `randomstuff`, `99`

They are not interchangeable. In this case, `isDefined` says

> Did the user add `FUNCTIONS_NODE_BLOCK_ON_ENTRY_POINT_ERROR` to their app settings?

If the answer is yes, we use `isEnvironmentVariableSet` to check

> Does the value of the app setting indicate true or false?

src/startApp.ts
ejizba: The reason I use aka.ms links is for their ability to change the underlying url at any time (without needing to ship the worker), and because it's way shorter than the real url. I've got about 40 urls I've created over the years. Ain't nobody got time to come up with cute names for all of them, plus you wanna save the good names for when you actually need them (like big announcements or social media or whatnot)

src/startApp.ts
hossam-nasr: I thought one of the arguments for having the app setting was for cases where the error happens before the default programming model is set?

test/eventHandlers/WorkerInitHandler.test.ts
hossam-nasr: I don't get what you're saying. It _shouldn't_ change the behavior, but don't we need to check it actually _doesn't_? You could also say v3/v4/app setting=0/1 shouldn't change the behavior if it's node >=20 anyway

src/startApp.ts
ejizba: The default is v3. I think you're thinking of the case where an error happens before the new v4 model is set

hossam-nasr: Oh makes sense now

test/eventHandlers/WorkerInitHandler.test.ts
ejizba: I try not to overthink unit tests in general - a bunch of this is super arbitrary. That's why my response was short, but I can explain more

First of all, we don't _need_ to test any of this. A lot of times the tests I choose to add are because I use [test-driven-development](https://en.wikipedia.org/wiki/Test-driven_development). If you're not familiar, that means I'll add tests before I write the source code, and ideally they fail until I actually implement the feature so that I know my code made a difference. My previous comment was alluding to the fact that your suggested test would've succeeded before I implemented any of the app setting logic, not really helping with my test-driven-development.

Second, I guess I figure your suggestion would be testing `isEnvironmentVariableSet` itself, which should theoretically be covered by its own unit tests instead of these unit tests.

Lastly, I copied all the node <= 20 tests to use for node >= 20, so I didn't really analyze each one individually. I somewhat consider unit tests as a spec for a feature, and when it comes to breaking changes, I like the idea of copying because you're essentially taking the previous spec and saying "okay here's what it does now"

ejizba:
A few updates after going back and forth with Brett on the host team. I was able to get an "end-to-end" scenario tested in Azure and I feel more comfortable with the overall experience now. Two main changes:

1. Log the error using `console.log`, which will forward the most important worker errors to a user's app insights
2. Don't throw the error during worker init or function environment reload, because the host interprets these failures as "system" errors (aka problem with the worker itself) and gets into a pretty broken state. Instead, throw the error in functionMetadata or functionLoad, which the host interprets as "user" errors (aka a problem with the user's code) and it handles it much better. A few example problems if we throw the error during worker init request:
    1. The portal won't show app keys - it will just load indefinitely
    2. The host will retry loading the worker, and end up sending a bunch of unhelpful errors to the user's app insights like "Error: 14 UNAVAILABLE"
    3. If you try to execute a function in the old model (with the "block" app setting turned on), the host will return a generic 404 error and nothing helpful in the `exceptions` table in app insights. If we throw the app start error during function load, the host will know to associate that with the execution and surface it in the `exceptions` table for app insights. (This only applies to the old model, because in that case the host loaded the functions before the app start error occured. In the new model, the app start error occurs before the worker can load the functions.)

src/eventHandlers/FunctionsMetadataHandler.ts
```diff
@@ -12,7 +13,7 @@ export class FunctionsMetadataHandler extends EventHandler<'functionsMetadataReq
 
     getDefaultResponse(_msg: rpc.IFunctionsMetadataRequest): rpc.IFunctionMetadataResponse {
         return {
-            useDefaultMetadataIndexing: true,
+            useDefaultMetadataIndexing: !worker.app.isUsingWorkerIndexing,
```
hossam-nasr, line 16: I'm having trouble understanding why this change was made?

src/eventHandlers/FunctionsMetadataHandler.ts
```diff
@@ -27,10 +28,12 @@ export class FunctionsMetadataHandler extends EventHandler<'functionsMetadataReq
             logCategory: LogCategory.System,
         });
 
-        const functions = Object.values(worker.app.functions);
-        if (functions.length > 0) {
-            response.useDefaultMetadataIndexing = false;
-            response.functionMetadataResults = functions.map((f) => f.metadata);
+        if (worker.app.isUsingWorkerIndexing) {
+            if (isDefined(worker.app.blockingAppStartError)) {
```
hossam-nasr, line 32: Why don't we throw `worker.app.blockingAppStartError` anyway, regardless of whether `worker.app.isUsingWorkerIndexing`?

src/startApp.ts
```diff
@@ -88,13 +99,45 @@ async function loadEntryPointFile(functionAppDirectory: string): Promise<void> {
...
+            }`;
+
+            if (shouldBlockOnEntryPointError()) {
+                error.message = newMessage;
+                error.isAzureFunctionsSystemError = true;
+                // We don't want to throw this error now (during workerInit or funcEnvReload) because technically the worker is fine
+                // Instead, it will be thrown during functionMetadata or functionLoad response which better indicates that the user's app is the problem
+                worker.app.blockingAppStartError = error;
+                // This will ensure the error makes it to the user's app insights
+                console.log(error.stack);
```
hossam-nasr: Does it make a difference here logging using `console.log` vs `console.error`?

src/startApp.ts
```diff
@@ -88,13 +99,45 @@ async function loadEntryPointFile(functionAppDirectory: string): Promise<void> {
...
+            } else {
+                // In this case, the error will never block the app
+                // The most we can do without breaking backwards compatibility is log it as a system log
+                worker.log({
```
hossam-nasr, line 117: Why don't we also use the `console.log` here?

test/eventHandlers/msg.ts
```diff
@@ -208,23 +211,54 @@ export namespace msg {
...
-        export function response(functions: rpc.IRpcFunctionMetadata[] = []): TestMessage {
+        export function response(
+            functions: rpc.IRpcFunctionMetadata[],
+            useDefaultMetadataIndexing: boolean
+        ): TestMessage {
             const response: rpc.IStreamingMessage = {
                 requestId: 'testReqId',
                 functionMetadataResponse: {
-                    useDefaultMetadataIndexing: functions.length === 0,
+                    useDefaultMetadataIndexing: useDefaultMetadataIndexing,
```
hossam-nasr, line 230: Is `functions.length` no longer a valid proxy for `useDefaultMetadataIndexing`?

src/eventHandlers/FunctionsMetadataHandler.ts
ejizba: Because if it's an old model app, we don't want the logs to say "worker indexing failed". Those apps shouldn't even know indexing is happening in the worker. Old model apps will see the error during function load instead

src/eventHandlers/FunctionsMetadataHandler.ts
ejizba: The "defaultResponse" is used when we throw an error (See [this line](https://github.com/Azure/azure-functions-nodejs-worker/blob/v3.x/src/setupEventStream.ts#L110) in the catch block). Previously we were never really throwing errors in `FunctionsMetadataRequest` since all the logic happened in init/envReload so the default response here didn't really matter. Now that we're throwing an error below, I wanted the default response to be more accurate which in turn makes the logs more accurate.

src/startApp.ts
ejizba: I thought they were the same, but console.error might result in a more accurate severityLevel in app insights. Let me try it out and get back to you

src/startApp.ts
ejizba: Because new errors showing up in a user's app insights can be considered a breaking change. Even if it doesn't change the behavior of the app, people still have alerts and stuff for errors being logged.

test/eventHandlers/msg.ts
ejizba: No it is not. You can imagine a scenario where someone tries to register a function using model v4, but an error happens before the register finishes. In that case, we still want to treat that app as a v4 app for error handling purposes even though functions.length is still zero

src/startApp.ts
hossam-nasr: Is it a breaking change if it's technically fixing a bug that those errors weren't being logged before? They aren't new errors, they were always there, they just weren't being logged to app insights. We had a similar precedent with #555 (though that did cause some problems for Durable users, so I don't have strong feelings about this)

ejizba: They are new errors (relatively) because we just added support for entry points last fall. At a higher level yeah unfortunately fixing a bug can often be a breaking change because people get used to the broken behavior

For better or worse, #555 never really affected users because of our various app insights problems with the worker. That PR allowed the logs to show up in our FunctionsLogs table for the first time, but they still don't show up in a user's app insights :(

hossam-nasr: Makes sense. Like I said, I don't have strong feelings here, so happy to err on the side of caution

hossam-nasr (approved):

src/startApp.ts
ejizba: Yeah had a better severity level, thanks. Updated

PR #696, opened 2023-06-21
Allow specifying retry policy for v4 model

ejizba:
Related to https://github.com/Azure/azure-functions-nodejs-library/issues/73

### DIFF
--- a/types-core/index.d.ts
+++ b/types-core/index.d.ts
@@ -43,6 +43,11 @@ declare module '@azure/functions-core' {
          * A dictionary of binding name to binding info
          */
         bindings: { [name: string]: RpcBindingInfo };
+
+        /**
+         * The retry policy options
+         */
+        retryOptions?: RpcRetryOptions;
     }
 
     /**
@@ -298,6 +303,8 @@ declare module '@azure/functions-core' {
         functionId?: string | null;
 
         managedDependencyEnabled?: boolean | null;
+
+        retryOptions?: RpcRetryOptions | null;
     }
 
     interface RpcStatusResult {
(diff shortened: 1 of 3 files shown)


types-core/index.d.ts
```diff
@@ -43,6 +43,11 @@ declare module '@azure/functions-core' {
          * A dictionary of binding name to binding info
          */
         bindings: { [name: string]: RpcBindingInfo };
+
+        /**
+         * The retry policy options
+         */
+        retryOptions?: RpcRetryOptions;
```
hossam-nasr, line 50: Why is this in the `FunctionMetadata` interface if it's not required? Unless I understand the distinction between `FunctionMetadata` and `RpcFunctionMetadata` incorrectly

hossam-nasr (approved):

types-core/index.d.ts
ejizba: I don't entirely remember why they're split up, but `FunctionMetadata` is what's used in the `registerFunction` api so it needs a property for anything the library package might specify when calling `registerFunction`

`RpcFunctionMetadata` is basically a strict copy of the generated rpc types and isn't cleaned up in any way.

hossam-nasr: I'm basing my comment here based on the doc string for this type: "A slimmed down version of `RpcFunctionMetadata` that includes the _minimum amount_ of information needed to register a function." (emphasis added). Maybe this doc string should be updated to be more accurate?

ejizba: sure - messed with the comment a bit

hossam-nasr (approved):

PR #705, opened 2023-07-13
Increase timeouts for worker gRPC respones

### LINKED ISSUE #704: Increase timeout for `workerInit` and `functionEnvironmentReload` requests
This issue was inspired by [this CRI](https://portal.microsofticm.com/imp/v3/incidents/details/397000313/home): a customer using programming model v4 had their app failing on startup, because the host kept killing the worker process with a 10 second timeout on `workerInit` response.

Previously, without worker indexing, user code was being loaded in the `functionLoad` request, which has a timeout of 5 minutes. However, with worker indexing and the v4 model, user code is being loaded in either `workerInit` or `functionEnvironmentReload` requests. These operations have a default timeout in the Host of 10 seconds and 30 seconds respectively. Most apps should be fine with the 10 second timeout as long as they use `WEBSITE_RUN_FROM_PACKAGE`, but plenty of people don't use that setting. 

This issue is to avoid such issues by overriding the default timeout of these operations for the node worker, to be similar to that of `functionLoad`,by changing the values in `worker.config`

hossam-nasr:
Resolves #704 

* Increases `workerInitResponse` default timeout from 10s to 2mins, for the case of customer code being loaded on startup
* Increases `functionEnvironmentReloadResponse` default timeout from 10s to 2mins, for the case of customer code being loaded in the specialization case

To verify, added the below code in a startup file:

```TS
const timeout = Date.now() + 11 * 1000;

while (Date.now() < timeout) {}
```

Before change:

image

After change:

image

### DIFF
--- a/worker.config.json
+++ b/worker.config.json
@@ -5,5 +5,10 @@
         "defaultExecutablePath":"node",
         "defaultWorkerPath":"dist/src/nodejsWorker.js",
         "workerIndexing": "true"
+    },
+    "processOptions": {
+        "initializationTimeout": "00:05:00",
+        "environmentReloadTimeout": "00:05:00",
+        "processShutdownTimeout": "00:05:00"
     }
 }
--- a/worker.config.json
+++ b/worker.config.json
@@ -7,8 +7,7 @@
         "workerIndexing": "true"
     },
     "processOptions": {
-        "initializationTimeout": "00:05:00",
-        "environmentReloadTimeout": "00:05:00",
-        "processShutdownTimeout": "00:01:00"
+        "initializationTimeout": "00:02:00",
+        "environmentReloadTimeout": "00:02:00"    
     }
 }
--- a/worker.config.json
+++ b/worker.config.json
@@ -9,6 +9,6 @@
     "processOptions": {
         "initializationTimeout": "00:05:00",
         "environmentReloadTimeout": "00:05:00",
-        "processShutdownTimeout": "00:05:00"
+        "processShutdownTimeout": "00:01:00"
     }
 }


worker.config.json
```diff
@@ -5,5 +5,10 @@
         "defaultExecutablePath":"node",
         "defaultWorkerPath":"dist/src/nodejsWorker.js",
         "workerIndexing": "true"
+    },
+    "processOptions": {
+        "initializationTimeout": "00:05:00",
+        "environmentReloadTimeout": "00:05:00",
+        "processShutdownTimeout": "00:01:00"
```
ejizba: I worry increasing the shutdown timeout could have unknown ripple effects. Worker timeout is still a relatively new feature and and I generally assume shutdown is supposed to happen fast. I don't think we should change this unless people ask for it

worker.config.json
```diff
@@ -5,5 +5,10 @@
         "defaultExecutablePath":"node",
         "defaultWorkerPath":"dist/src/nodejsWorker.js",
         "workerIndexing": "true"
+    },
+    "processOptions": {
+        "initializationTimeout": "00:05:00",
```
ejizba: Since these timeouts are "new" and pretty much unique to the new model, we have an opportunity to do something shorter than what we did in the old model. 5 minutes feels egregious to me, and I kind of want to put more pressure on people to use `WEBSITE_RUN_FROM_PACKAGE`. People don't realize how much happier they will be if they just add the setting

To me, 1 minute feels like a reasonable timeout. But I normally double what I think is reasonable to account for unexpected edges cases so I guess I'd vote for 2 minutes.

IMO we can always increase this value in the future, but I don't think we can decrease it.

worker.config.json
hossam-nasr: I was hesitant about making this change too, so I'm on board with scrapping it. That being said, we'd have to advertise that `appTerminate` hooks must terminate within 10 seconds or less. Which isn't a tall ask, but I think it should be made clear

worker.config.json
hossam-nasr: I don't have strong feelings here so I'm happy to defer to your experience and reduce to 2 minutes

hossam-nasr: Done

ejizba (approved):

PR #732, opened 2024-01-31
Check errors are writable

castrodd:
This PR addresses an issue raised in the Library repo: https://github.com/Azure/azure-functions-nodejs-library/issues/205.

In short, a customer was using a package that extends the `Error` class and renders the `message` property read-only.

This caused an error in `FunctionLoadHandler.ts` when we attempt to write our own custom error message:
```
error.message = `Worker was unable to load function ${metadata.name}: '${error.message}'`;

```
Our function `ensureErrorType` was not filtering out these read-only errors because  `instanceof Error` was in fact true. Therefore, we simply returned these errors without any modification, assuming they were in fact instances of `Error`, not a subclass with different behavior.

This PR introduces an explicit check to ensure that we are not passing along read-only objects and hitting this exception:
```
Exception: Cannot set property message of [object Object] which has only a getter
```

### DIFF
--- a/src/errors.ts
+++ b/src/errors.ts
@@ -38,7 +38,7 @@ export function ensureErrorType(err: unknown): Error & Partial<AzFuncError> {
                 get() {
                     return readableMessage;
                 },
-                set(val) {
+                set(val: string) {
                     readableMessage = val;
                 },
             });
--- a/src/errors.ts
+++ b/src/errors.ts
@@ -31,9 +31,9 @@ export function ensureErrorType(err: unknown): Error & Partial<AzFuncError> {
     if (err instanceof Error) {
         if (Object.getOwnPropertyDescriptor(err, 'message')?.writable) {
             return err;
+        } else {
+            return new Error(JSON.stringify(err));
         }
-
-        return new Error(JSON.stringify(err));
     } else {
         let message: string;
         if (err === undefined || err === null) {
--- a/src/errors.ts
+++ b/src/errors.ts
@@ -30,19 +30,20 @@ export class ReadOnlyError extends AzFuncTypeError {
 export function ensureErrorType(err: unknown): Error & Partial<AzFuncError> {
     if (err instanceof Error) {
         const writable = Object.getOwnPropertyDescriptor(err, 'message')?.writable;
-        if (writable) {
-            return err;
-        } else {
-            // If we cannot modify the message, we want to wrap the original error
+        if (!writable) {
             // The motivation for this branch can be found in the below issue:
             // https://github.com/Azure/azure-functions-nodejs-library/issues/205
-            const oldStack = err.stack;
-            const oldName = err.name;
-            const newError = new Error(err.message);
-            newError.stack = oldStack;
-            newError.name = oldName;
-            return newError;
+            let readableMessage = err.message;
+            Object.defineProperty(err, 'message', {
+                get() {
+                    return readableMessage;
+                },
+                set(val) {
+                    readableMessage = val;
+                },
+            });
         }
+        return err;
     } else {
         let message: string;
         if (err === undefined || err === null) {
--- a/src/errors.ts
+++ b/src/errors.ts
@@ -28,9 +28,12 @@ export class ReadOnlyError extends AzFuncTypeError {
 }
 
 export function ensureErrorType(err: unknown): Error & Partial<AzFuncError> {
-    const writable = Object.getOwnPropertyDescriptor(err, 'message')?.writable;
-    if (err instanceof Error && writable) {
-        return err;
+    if (err instanceof Error) {
+        if (Object.getOwnPropertyDescriptor(err, 'message')?.writable) {
+            return err;
+        }
+
+        return new Error(JSON.stringify(err));
     } else {
         let message: string;
         if (err === undefined || err === null) {
--- a/src/errors.ts
+++ b/src/errors.ts
@@ -33,9 +33,12 @@ export function ensureErrorType(err: unknown): Error & Partial<AzFuncError> {
         if (writable) {
             return err;
         } else {
+            // If we cannot modify the message, we want to wrap the original error
+            // The motivation for this branch can be found in the below issue:
+            // https://github.com/Azure/azure-functions-nodejs-library/issues/205
             const oldStack = err.stack;
             const oldName = err.name;
-            const newError = 'message' in err ? new Error(err.message) : new Error(JSON.stringify(err));
+            const newError = new Error(err.message);
             newError.stack = oldStack;
             newError.name = oldName;
             return newError;
--- a/src/errors.ts
+++ b/src/errors.ts
@@ -29,7 +29,8 @@ export class ReadOnlyError extends AzFuncTypeError {
 
 export function ensureErrorType(err: unknown): Error & Partial<AzFuncError> {
     if (err instanceof Error) {
-        if (Object.getOwnPropertyDescriptor(err, 'message')?.writable) {
+        const writable = Object.getOwnPropertyDescriptor(err, 'message')?.writable;
+        if (writable) {
             return err;
         } else {
             return new Error(JSON.stringify(err));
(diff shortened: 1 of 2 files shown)


test/errors.test.ts
```diff
@@ -42,6 +42,20 @@ describe('errors', () => {
         expect(ensureErrorType(actualError)).to.equal(actualError);
     });
 
+    it('readonly error', () => {
+        class ReadOnlyError extends Error {
+            readonly message = 'a readonly message';
```
ejizba: The TypeScript `readonly` keyword is pretty useless tbh - it disappears when compiled to JavaScript. Here is the typical way to make something actually readonly:
```typescript
class ReadOnlyError extends Error {
    get message() {
        return 'a readonly message';
    }
}
```

You shouldn't need to do this: `Object.defineProperty(actualError, 'message', { writable: false });`

src/errors.ts
```diff
@@ -29,7 +29,12 @@ export class ReadOnlyError extends AzFuncTypeError {
 
 export function ensureErrorType(err: unknown): Error & Partial<AzFuncError> {
     if (err instanceof Error) {
-        return err;
+        const writable = Object.getOwnPropertyDescriptor(err, 'message')?.writable;
+        if (writable) {
+            return err;
+        } else {
+            return new Error(JSON.stringify(err));
```
ejizba: It's important that the value of `err.stack` is preserved on the new error as it's super helpful when debugging. The problem is that when you do `new Error()` the stack will change to this point in the code instead of the original point where err was thrown

In general, the value returned should be as close as possible to the original, because we don't know which properties might exist on the original object or if they're important

ejizba: If you're having troubles figuring out how to accomplish all that, I want to point out that we don't _have_ to make `message` settable. We could take a more reactive approach and somehow ignore errors if setting fails. The original error/stack is much more important that any additional text we're adding IMO

castrodd: I made some changes to my initial approach. Up until now I have avoided the reactive approach because I did not want to pass the responsibility to the caller but also because it would involve some ugly decisions like whether to put a try/catch inside another try/catch.

Also, before I forget to mention this, we also have `ensureErrorType` in the library repo: https://github.com/Azure/azure-functions-nodejs-library/blob/de87cbe25e05af90c2109c63e1e790dd252ebb18/src/errors.ts#L30. Whatever we decide to do here in the worker, we should probably copy over there too.

test/errors.test.ts
```diff
@@ -42,6 +42,25 @@ describe('errors', () => {
         expect(ensureErrorType(actualError)).to.equal(actualError);
     });
 
+    it.only('readonly error', () => {
```
ejizba: I had to google `it.only` because I've never heard of it. Sounds useful, but shouldn't be committed right?

src/errors.ts
```diff
@@ -29,7 +29,17 @@ export class ReadOnlyError extends AzFuncTypeError {
 
 export function ensureErrorType(err: unknown): Error & Partial<AzFuncError> {
     if (err instanceof Error) {
-        return err;
+        const writable = Object.getOwnPropertyDescriptor(err, 'message')?.writable;
+        if (writable) {
+            return err;
+        } else {
+            const oldStack = err.stack;
```
ejizba: Please add a reference to the issue, as I don't think the reason for this logic is obvious:
```
// https://github.com/Azure/azure-functions-nodejs-library/issues/205
```

src/errors.ts
```diff
@@ -29,7 +29,17 @@ export class ReadOnlyError extends AzFuncTypeError {
...
+            const oldName = err.name;
+            const newError = 'message' in err ? new Error(err.message) : new Error(JSON.stringify(err));
```
ejizba: I think we should just do this:
```typescript
const newError = new Error(err.message);
```

For example, we typically log the stack (which includes the message), and if you do `new Error(JSON.stringify(err))`, we're going to end up with duplicated stack information being logged. I don't see why anyone would throw an error with an empty message, but I feel like it's not this function's problem to solve (we don't solve it in the `if (writable)` case either)

src/errors.ts
ejizba: Yes, please copy the change to the library once this is approved

test/errors.test.ts
castrodd: Ah, yes, sorry. That was just to speed up testing.

test/errors.test.ts
castrodd: It's a little surprising how unhelpful `readonly` actually is. I have updated as suggested.

src/errors.ts
castrodd: I originally thought ZodErrors removed the `message` property, but I went and checked the source code and I was mistaken: https://github.com/colinhacks/zod/blob/a5a9d31018f9c27000461529c582c50ade2d3937/src/ZodError.ts#L192.

I have simplified the line.

src/errors.ts
castrodd: Added a short explanation and reference

src/errors.ts
castrodd: I will keep this thread open as a reminder. Once you approve the implementation here, I will open that new PR as well.

castrodd: Link to companion PR: https://github.com/Azure/azure-functions-nodejs-library/pull/220

ejizba (approved):

src/errors.ts
```diff
@@ -29,6 +29,20 @@ export class ReadOnlyError extends AzFuncTypeError {
 
 export function ensureErrorType(err: unknown): Error & Partial<AzFuncError> {
     if (err instanceof Error) {
+        const writable = Object.getOwnPropertyDescriptor(err, 'message')?.writable;
+        if (!writable) {
+            // The motivation for this branch can be found in the below issue:
+            // https://github.com/Azure/azure-functions-nodejs-library/issues/205
+            let readableMessage = err.message;
+            Object.defineProperty(err, 'message', {
```
ryanbecker, line 37: We're seeing a TypeError that causes an unhandled promise rejection on this line of code and prevents the actual error message from being raised. Here's the TypeError:
```json
{
    "message": "Cannot redefine property: message",
    "name": "TypeError",
    "stack": "TypeError: Cannot redefine property: message\n    at Function.defineProperty (<anonymous>)\n    at t.ensureErrorType (C:\\Program Files (x86)\\SiteExtensions\\Functions\\4.34.1\\workers\\node\\dist\\src\\worker-bundle.js:2:31119)\n    at C:\\Program Files (x86)\\SiteExtensions\\Functions\\4.34.1\\workers\\node\\dist\\src\\worker-bundle.js:2:53526\n    at Generator.throw (<anonymous>)\n    at a (C:\\Program Files (x86)\\SiteExtensions\\Functions\\4.34.1\\workers\\node\\dist\\src\\worker-bundle.js:2:51940)\n    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)"
}
```

ryanbecker: Maybe the error being thrown is sealed? I can reproduce the error we're seeing with this test code:
```javascript
const error = Object.seal(new Error('Sealed error'));
Object.defineProperty(error, 'message', { get: () => 'Redefined error message' });
```

ejizba: @ryanbecker to clarify are you worse off after we shipped this PR or just equally bad?

ryanbecker: I believe in this case we're worse off. We'll wrap all our function code in a try/catch to log the error that's causing this so we can better understand the root cause, but I expect the error object is sealed. Maybe the error doesn't need to be mutated [here](https://github.com/Azure/azure-functions-nodejs-worker/blob/588e99fa3945b3cfa82d6f32aec91f96f1735cdc/src/eventHandlers/FunctionLoadHandler.ts#L46) or in this code? It might be difficult to ensure this works with any error that could be thrown.

castrodd: @ryanbecker I just tried out your test code. It's an interesting corner case. We do not enter this `if` block because the sealed object is in fact `writable`.  However, the problem is that it is not `configurable`. It seems we need to check that both these properties are set to `true`.

I found a way to do that with `structuredClone`. I will discuss with Eric first though. Thanks for raising this issue!

castrodd: @ryanbecker I opened a new PR for this issue: #750.

PR #739, opened 2024-04-11
Add hook for logs

ejizba:
This PR provides a clean way to perform an operation every time the worker sends a log. The motivation behind this is to send each log to open telemetry (tracked by https://github.com/Azure/azure-functions-nodejs-library/issues/245). Most of the otel support will be in a separate library, and I have a working prototype I can demo offline, but nothing ready for a legit PR yet. The library can be finalized at a later date, but I want to get these worker changes in so they can start rolling out.

There is a workaround available to the library today meaning these changes aren't strictly necessary, but it has two drawbacks:
1. It's ugly (as described [here](https://github.com/Azure/azure-functions-nodejs-library/issues/228#issuecomment-1967487109))
2. It only lets you access user logs and this new hook lets you access both user and system logs


types-core/index.d.ts
```diff
@@ -58,13 +58,15 @@ declare module '@azure/functions-core' {
     function registerHook(hookName: 'postInvocation', callback: PostInvocationCallback): Disposable;
     function registerHook(hookName: 'appStart', callback: AppStartCallback): Disposable;
     function registerHook(hookName: 'appTerminate', callback: AppTerminateCallback): Disposable;
+    function registerHook(hookName: 'log', callback: LogHookCallback): Disposable;
     function registerHook(hookName: string, callback: HookCallback): Disposable;
 
-    type HookCallback = (context: HookContext) => void | Promise<void>;
+    type HookCallback = (context: HookContext) => unknown;
```
castrodd, line 64: I am just wondering what necessitated the change to `unknown` here?

src/hooks/LogHookContext.ts
```diff
@@ -0,0 +1,57 @@
...
+
+export interface InvocationLogContext {
+    hookData: HookData;
+    invocationContext: unknown;
+}
+
+export class LogHookContext implements coreTypes.LogHookContext {
+    level: RpcLogLevel;
+    message: string;
+    #hookData: HookData;
```
castrodd: Nit: should the private fields be listed first? And should we sort them alphabetically?

src/hooks/LogHookContext.ts
```diff
@@ -0,0 +1,57 @@
...
+    #category: RpcLogCategory;
+    #invocationContext: unknown;
+
+    constructor(log: rpc.IRpcLog, invocLogCtx: InvocationLogContext | undefined) {
+        this.#hookData = invocLogCtx?.hookData ?? {};
+        this.#invocationContext = invocLogCtx?.invocationContext;
+        const coreLog = toCoreLog(log);
+        this.level = nonNullProp(coreLog, 'level');
```
castrodd, line 26: Same nit here. I would rather all the private fields appear together at the beginning or at the end. This way seems inconsistent. I get why you did it this way, but I am okay with moving toCoreLog() to the top of it helps the ordering of what follows.

types-core/index.d.ts
ejizba: The log hooks couldn't return `| Promise<void>` like all the other hooks unless I did some serious refactoring, so I changed this to unknown to more easily let the log one just return `void`

src/WorkerContext.ts
```diff
@@ -62,7 +64,23 @@ class WorkerContext {
      * @param requestId gRPC message request id
      * @param msg gRPC message content
      */
-    log(log: rpc.IRpcLog) {
+    log(log: rpc.IRpcLog, invocationLogCtx?: InvocationLogContext): void {
+        try {
+            const logContext = new LogHookContext(log, invocationLogCtx);
+            for (const callback of worker.app.logHooks) {
+                callback(logContext);
```
castrodd, line 71: I am having a hard time grokking this. What will the callback from the logHooks do with the logContext? How will we be modifying it? Can you briefly walk me through that?

test/eventHandlers/InvocationHandler.test.ts
```diff
@@ -1063,4 +1063,37 @@ describe('InvocationHandler', () => {
...
         );
     });
+
+    it('log hook respects changes to value, only for user log', async () => {
+        coreApi.registerHook('log', (ctx) => {
+            ctx.message += 'UpdatedFromHook';
+            ctx.level = 'error';
+        });
+
+        registerV3Func(Binding.queue, async (invocContext: Context) => {
```
castrodd, line 1073: General question. I see the worker tests have this method, but why is it called `registerV3Func`? I don't see a `registerV4Func` anywhere. Isn't it just registering _a_ function? Why the distinction between V3 and V4?

castrodd (approved):

src/hooks/LogHookContext.ts
ejizba: There's no functional reason we need to sort them. We could do it as a stylistic choice, but I'm not a big fan of stylistic rules unless we have a linter or formatter to enforce them. Seems like both eslint and prettier deemed this kind of sorting as too complex to be worth supporting:
- https://github.com/prettier/prettier/issues/5412
- https://github.com/typescript-eslint/typescript-eslint/issues/8792

The eslint rule is in a weird limbo state and we could still decide to use it, but it would need to be enabled in a separate PR for the whole repo. The default is private fields last, btw, which is what I'm used to.

src/hooks/LogHookContext.ts
ejizba: I can update this as I agree it's a bit weird. I normally try to match the ordering of the properties above, which I obviously didn't do

test/eventHandlers/InvocationHandler.test.ts
ejizba: v3 is special because it's the default if you don't have a programming model specified. The worker has a [reference](https://github.com/Azure/azure-functions-nodejs-worker/blob/v3.x/package.json#L8) to the v3 package because of this, but not the v4 package. Also most of these tests were written before the v3/v4 split, so they're using v3 for historical reasons

We have a few tests to verify the core api (which the v4 model uses to register a function), but in general the worker shouldn't need to know anything about the programming model so it's okay that we're not explicitly testing v4 in this repo

src/WorkerContext.ts
ejizba: The main idea is I will convert the log to an open telemetry format and then emit the log directly from our side. I will also tell the host (using a separate flag not included in this example) that we already emitted the log and they don't need to. Here's a simplified example:
```typescript
app.hook.log((context) => {
    otelLogger.emit({
        severityNumber: toOtelSeverityNumber(context.level),
        body: context.message,
        attributes: {
            invocationId: context.invocationContext?.invocationId,
        },
    });
});
```

In addition to emitting logs to a custom destination (like open telemetry), users could use this hook to tweak the logs that get sent to app insights:
```typescript
app.hook.log((context) => {
    if (context.category === 'system' && context.level === 'information') {
        // decrease severity level of system logs if a user finds them annoying
        context.level = 'trace';
    }

    // insert a timestamp prefix for all logs which would be more accurate than the existing timestamp property
    context.message = `${Date.now()}: ${context.message}`;
});
```

castrodd: Got it. That makes sense now. Thank you!

PR #750, opened 2024-06-17
Create setErrorMessage function

castrodd:
This PR is a fix for [Exception: Cannot set property message of [object Object] which has only a getter](https://github.com/Azure/azure-functions-nodejs-library/issues/205).

In February, we merged an earlier fix for this same issue (#732).

However, @ryanbecker reported a problem with our initial approach.

Therefore, we have rolled back those changes and instead take a simpler approach. Everywhere we previously tried to write to `error.message`, we now catch any exceptions thrown by read-only errors. In those cases, we simply return the original error and do not attempt to modify its error message.

### DIFF
--- a/src/errors.ts
+++ b/src/errors.ts
@@ -29,20 +29,6 @@ export class ReadOnlyError extends AzFuncTypeError {
 
 export function ensureErrorType(err: unknown): Error & Partial<AzFuncError> {
     if (err instanceof Error) {
-        const writable = Object.getOwnPropertyDescriptor(err, 'message')?.writable;
-        if (!writable) {
-            // The motivation for this branch can be found in the below issue:
-            // https://github.com/Azure/azure-functions-nodejs-library/issues/205
-            let readableMessage = err.message;
-            Object.defineProperty(err, 'message', {
-                get() {
-                    return readableMessage;
-                },
-                set(val: string) {
-                    readableMessage = val;
-                },
-            });
-        }
         return err;
     } else {
         let message: string;
@@ -56,6 +42,16 @@ export function ensureErrorType(err: unknown): Error & Partial<AzFuncError> {
             message = String(err);
         }
         return new Error(message);
+    }
+}
+
+export function setErrorMessage(err: Error, message: string): Error & Partial<AzFuncError> {
+    try {
+        err.message = message;
+        return err;
+    } catch {
+        // Return original error if we can't set the message
+        return err;
     }
 }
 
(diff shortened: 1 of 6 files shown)


src/errors.ts
```diff
@@ -29,20 +29,6 @@ export class ReadOnlyError extends AzFuncTypeError {
 
 export function ensureErrorType(err: unknown): Error & Partial<AzFuncError> {
```
ejizba: How about we change the return type of `ensureErrorType` to something like this. That way we should get a build error if we forget and try to set the message directly

```
export interface TodoError extends Error, Partial<AzFuncError> {
    /**
     * Use `setErrorMessage` to set the error message
     */
    readonly message: string;
}

export function ensureErrorType(err: unknown): TodoError {
```

src/errors.ts
```diff
@@ -59,6 +45,20 @@ export function ensureErrorType(err: unknown): Error & Partial<AzFuncError> {
     }
 }
 
+export function setErrorMessage(err: Error, message: string): Error & Partial<AzFuncError> {
```
ejizba: nit: I would rename to `trySetErrorMessage`

src/errors.ts
```diff
@@ -59,6 +45,20 @@ export function ensureErrorType(err: unknown): Error & Partial<AzFuncError> {
...
+        err.message = message;
+        return err;
+    } catch (e) {
+        if (e instanceof TypeError) {
+            // Return original error if we can't set the message
+            return err;
+        } else {
+            throw e;
+        }
+    }
```
ejizba: I think we should ignore all errors, not just TypeError. Also, you're modifying the original `err` directly, so it would be simpler to just return void

```suggestion
    try {
        err.message = message;
    } catch (e) {
        // ignore
    }
```

src/errors.ts
castrodd: Added the interface

src/errors.ts
castrodd: Updated name

src/errors.ts
castrodd: Agreed. I have made the suggested changes.

src/errors.ts
```diff
@@ -9,6 +9,13 @@ export interface AzFuncError {
     isAzureFunctionsSystemError: boolean;
 }
 
+export interface TodoError extends Error, Partial<AzFuncError> {
```
ejizba: Please update the name `TodoError` which came from my example haha :)

ejizba (approved):

src/errors.ts
castrodd: I was a bit unsure about the name, but I guessed you meant setting the error message was something `ToDo` on the object 😅

PR #751, opened 2024-06-21
Run builds nightly

ejizba:


azure-pipelines/official-build.yml
```diff
@@ -12,6 +12,14 @@ trigger:
 # CI only, does not trigger on PRs.
 pr: none
 
+schedules:
+    - cron: '30 10 * * *'
+      displayName: Nightly at 3:30 AM PT
```
castrodd: Isn't the cron expression set for 10:30 am? If this is for UTC. then I feel we should clarify that in the display name. For example, right now, UTC is +7 hours, but in the winter it will be +8 hours and so this will be misleading.

castrodd (approved):

PR #753, opened 2024-06-24
Fix startup error not showing up in kusto

ejizba:
Found this bug a couple weeks ago when Thiago was debugging a bad node v4 app on flex.

## Repro steps

1. Create a v4 app
2. Use Node 20
3. Delete the http trigger code and add `throw new Error('test');` instead
4. Run your app
5. Wait a while then check app insights & kusto

## Expected

The error shows up in customer's app insights and our kusto

## Actual

The error only shows up in customer's app insights, but not kusto

## More info

Today, we save app startup errors that happen during workerInitRequest and return them during functionsMetadataRequest or functionLoadRequest instead (which also logs them - which is what I care about for this PR). However, if the error happens for a v4 app before the v4 model is registered in user code, we skip sending the error during functionsMetadataRequest because we think it's a v3 app and the host never sends a functionLoadRequest which only applies to v3 apps.

My fix is to always log the error right away and set a flag to prevent it from potentially being duplicated later.

### DIFF
--- a/src/errors.ts
+++ b/src/errors.ts
@@ -8,7 +8,7 @@ export interface AzFuncError {
      */
     isAzureFunctionsSystemError: boolean;
 
-    alreadyLoggedOverRpc?: boolean;
+    loggedOverRpc?: boolean;
 }
 
 export interface ValidatedError extends Error, Partial<AzFuncError> {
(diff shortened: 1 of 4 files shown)


src/errors.ts
```diff
@@ -7,6 +7,8 @@ export interface AzFuncError {
      * User errors cannot be tracked in our telemetry because they could have user information (users can still track it themselves in their app insights resource)
      */
     isAzureFunctionsSystemError: boolean;
+
+    alreadyLoggedOverRpc?: boolean;
```
castrodd: Nit: `already` sounds redundant. `loggedOverRpc` is more direct and conveys the idea as well.

castrodd (approved):

Issue #737, opened 2024-03-29
Worker init and func env reload pointing to the same app fails to register functions


ejizba:
Created for [this incident](https://portal.microsofticm.com/imp/v3/incidents/details/484018781/home) (internal-only link). This appears to be a rare edge case. For this particular app, it only happened once in the last 90 days.

Normally we get either a worker init request or a func env reload request pointing to the user's app. However, in this incident we got both pointing to the same app. I don't think that should happen, so I filed [this issue](https://github.com/Azure/azure-functions-host/issues/9958) on the host, but we might be able to handle this on our side. The problem is that Node.js caches modules when you load them. We load the user's code for the worker init request, but we clear our own caches (including the functions that have been registered) when we do the func env reload. But when we try to re-load the user's code, it doesn't actually re-execute that code because of Node.js's module cache and no functions get registered. We would have to somehow clear Node.js's module cache at the same time we clear our own caches to fix this.

PR #758, opened 2024-08-28
Add logging to detect if function app dir doesn't change

ejizba:
Related to https://github.com/Azure/azure-functions-nodejs-worker/issues/737

My plan to fix that bug is to skip environment reload if we detect the functionAppDirectory is the same during workerInit and functionEnvironmentReload. However, I want to validate a few assumptions first:
1. I assume functionAppDirectory is always defined in both workerInit and functionEnvironmentReload
2. I assume the only time the directories are the same is the edge case mentioned in the bug above

The bug itself is rare, and IMO it would be a big deal if we skip environment reload for apps that still needed it - so that's why I'd rather be careful and add logs before actually making the change.

### DIFF
--- a/src/eventHandlers/FunctionEnvironmentReloadHandler.ts
+++ b/src/eventHandlers/FunctionEnvironmentReloadHandler.ts
@@ -10,6 +10,7 @@ import { getWorkerMetadata } from './getWorkerMetadata';
 import LogCategory = rpc.RpcLog.RpcLogCategory;
 import LogLevel = rpc.RpcLog.Level;
 import CapabilitiesUpdateStrategy = rpc.FunctionEnvironmentReloadResponse.CapabilitiesUpdateStrategy;
+import * as path from 'path';
 
 /**
  * Environment variables from the current process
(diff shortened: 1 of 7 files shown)


ejizba:
ping @gavin-aguiar

src/eventHandlers/FunctionEnvironmentReloadHandler.ts
```diff
@@ -27,7 +28,27 @@ export class FunctionEnvironmentReloadHandler extends EventHandler<
...
+            worker.app.functionAppDirectory &&
+            msg.functionAppDirectory &&
+            isPathEqual(worker.app.functionAppDirectory, msg.functionAppDirectory)
+        ) {
+            worker.log({
+                message: `FunctionEnvironmentReload functionAppDirectory has not changed`,
+                level: LogLevel.Debug,
+                logCategory: LogCategory.System,
+            });
+        }
```
gavin-aguiar, line 49: Should we log the paths only when they differ, instead of always logging? Since path differences are likely rare, logging only when they don't match could reduce unnecessary logging noise

src/eventHandlers/FunctionEnvironmentReloadHandler.ts
```diff
@@ -63,3 +84,7 @@ export class FunctionEnvironmentReloadHandler extends EventHandler<
         return response;
     }
 }
+
+function isPathEqual(path1: string, path2: string): boolean {
+    return path.relative(path1, path2) === '';
```
gavin-aguiar, line 89: Paths in linux are case sensitive. Could there be a case where these 2 paths have different cases?

src/eventHandlers/FunctionEnvironmentReloadHandler.ts
ejizba: I think you have it backwards. The vast majority of the time the paths will differ:
- Placeholder example for worker init: /tmp/functions/standby/wwwroot
- Actual app example for environment reload: /home/site/wwwroot

It's only a problem for us when the paths are the same, which should be rare.

src/eventHandlers/FunctionEnvironmentReloadHandler.ts
ejizba: That's a big part of why I use the `path.relative` method instead of comparing the string paths directly. The `path.relative` method handles all the weird edge cases (like case sensitivity) for us

gavin-aguiar (approved):

PR #765, opened 2025-03-14
Migrating the NuGet Publishing Pipeline to 1ES

swapnil-nagar:
Description 

Migrated worker release pipeline to 1ES. 

Key Information

1. Created a new build pipeline dedicated to worker release.
2. Tied the official build to release pipeline.
3. Publish package to nuget.

Proof
https://azfunc.visualstudio.com/internal/_build/results?buildId=206711&view=results

### DIFF
--- a/azure-pipelines/worker-build.yml
+++ b/azure-pipelines/worker-build.yml
@@ -1,0 +1,53 @@
+parameters:
+  - name: IsPrerelease
+    type: boolean
+    default: true
+
+trigger:
+  batch: true
+  branches:
+    include:
+      - v3.x
+
+pr: none
+
+resources:
+    repositories:
+        - repository: 1es
+          type: git
+          name: 1ESPipelineTemplates/1ESPipelineTemplates
+          ref: refs/tags/release
+
+extends:
+  template: v1/1ES.Official.PipelineTemplate.yml@1es
+  parameters:
+    pool:
+      name: 1es-pool-azfunc
+      image: 1es-windows-2022
+      os: windows
+    sdl:
+      codeql:
+        runSourceLanguagesInSourceAnalysis: true
+
+    stages:
+      - stage: Build
+        dependsOn: []
+        jobs:
+          - job:
+            templateContext:
+              outputs:
+                - output: pipelineArtifact
+                  path: $(Build.ArtifactStagingDirectory)/dropOutput
+                  artifact: drop
+                  sbomBuildDropPath: '$(Build.ArtifactStagingDirectory)/dropInput'
+                  sbomPackageName: 'Azure Functions Node.js Worker'
+                  # The list of components can't be determined from the webpacked file in the staging dir, so reference the original node_modules folder
+                  sbomBuildComponentPath: '$(Build.SourcesDirectory)/node_modules'
+            steps:
+              - template: /azure-pipelines/templates/build.yml@self
+                parameters:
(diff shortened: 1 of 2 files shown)


azure-pipelines/worker-build.yml
```diff
@@ -0,0 +1,53 @@
+parameters:
+  - name: IsPrerelease
+    type: boolean
+    default: true
+
+trigger:
+  batch: true
+  branches:
+    include:
+      - v3.x
```
hallvictoria: The worker release pipeline will trigger off of every merge to v3.x. Can you make it a manual pipeline / configure it so that it runs only when the worker version is bumped?

azure-pipelines/worker-build.yml
```diff
@@ -0,0 +1,53 @@
...
+                  path: $(Build.ArtifactStagingDirectory)/dropOutput
+                  artifact: drop
+                  sbomBuildDropPath: '$(Build.ArtifactStagingDirectory)/dropInput'
+                  sbomPackageName: 'Azure Functions Node.js Worker'
+                  # The list of components can't be determined from the webpacked file in the staging dir, so reference the original node_modules folder
+                  sbomBuildComponentPath: '$(Build.SourcesDirectory)/node_modules'
+            steps:
+              - template: /azure-pipelines/templates/build.yml@self
+                parameters:
+                  IsPrerelease: ${{ parameters.IsPrerelease }}
```
hallvictoria: If the pipeline is only run for releases, this variable isn't needed / can be set to True

azure-pipelines/worker-build.yml
```diff
@@ -0,0 +1,53 @@
...
+        runSourceLanguagesInSourceAnalysis: true
+
+    stages:
+      - stage: Build
+        dependsOn: []
+        jobs:
+          - job:
+            templateContext:
+              outputs:
+                - output: pipelineArtifact
```
hallvictoria: nit (possibly future change): since this code is shared w/ the official build pipeline, can you refactor it into a separate template to be reference in both pipelines?

azure-pipelines/worker-build.yml
swapnil-nagar: There is no schedule for this pipeline, it is meant to be manual only.

That is good suggestion. But, as we do not persist the version for worker anywhere, I am not certain how would ensure worker version is bumped.

Alternatively, what we can do it we can have the preRelease tag which is checked by default and we wont publish to nuget when preRelease tag is false. But when we need to release a new worker version we will trigger this pipeline manually.

azure-pipelines/worker-build.yml
swapnil-nagar: Actually, let me use preRelease variable, to add a condition to nuget to only publish when it is not a pre-release.

azure-pipelines/worker-build.yml
swapnil-nagar: Sure, once this PR is in I will refactor.

azure-pipelines/worker-build.yml
swapnil-nagar: Removed the branch trigger

azure-pipelines/worker-build.yml
swapnil-nagar: Added the preRelease==false as a mandatory condition for nuget release.

hallvictoria (approved):

PR #783, opened 2026-02-06
Suppress Logs

swapnil-nagar:
We noticed that the Node worker is emitting worker logs and sending them to the customer’s Application Insights. Since we need to ensure we don’t break existing customers, we’re planning to introduce an environment variable to disable these logs (rather than enabling them by default).


RohitRanjanMS (dismissed):

src/constants.ts
```diff
@@ -1,7 +1,7 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License.
 
-export const version = '3.12.0';
+export const version = '3.12.1';
```
gavin-aguiar: Change to 3.13.0. Patch version should mainly be used for bug fixes of the current version.

gavin-aguiar (approved):

PR #811, opened 2026-05-12
Honor functions-uri transport scheme in Node worker

larohra:
## Goal
Make the Node worker stop hard-coding plaintext gRPC and instead honor the scheme already provided in `--functions-uri`, without breaking the current v3.x host contract.

## Findings
- `src/Worker.ts` parses `--functions-uri` but drops the scheme via `new URL(uri).host`; `src/GrpcClient.ts` always calls `grpc.credentials.createInsecure()`.
- This was intentional in the original design, not a random missed API choice: the host currently starts the RPC server on loopback `http://127.0.0.1:<port>` and treats host↔worker gRPC as trusted local IPC (`WorkerConstants.HostName = "127.0.0.1"`, `HttpScheme = "http"`, and the host server comment says all clients are trusted).
- The fix is still necessary because the Node worker cannot currently participate in a secure rollout even though the existing contract already carries a URI with a scheme.

## Design
- Change the worker to pass the full `functions-uri` (or parsed `URL`) into `CreateGrpcEventStream`.
- Refactor `src/GrpcClient.ts` to derive:
  - target address from `url.host`
  - credentials from the URI scheme: `createInsecure()` for `http:` and `createSsl()` for `https:`
  - a clear failure for unsupported schemes
- Keep `http:` support on v3.x; do **not** unconditionally force TLS.
- Add concise comments/logging explaining that the `http:` fallback is a compatibility choice until the host advertises a secure endpoint.

## Compatibility / regression assessment
- No customer-facing contract change is required if we keep using the existing `functions-uri` argument; the worker simply stops ignoring its scheme.
- Unconditional TLS would be a breaking change against current hosts, so rollout must stay scheme-based and backward compatible.
- Full end-to-end plaintext removal still requires host-side changes because the current host listens on `http://127.0.0.1`.
- With scheme-based selection, expected regressions are low and limited to URI parsing/credential selection mistakes, which can be covered by unit tests.

## Test plan
- Baseline validation currently passes after setup: `npm ci`, `npm run build`, and `npm test` (132 passing, 8 pending).
- Add unit tests around URI parsing and credential selection using stubs for `@grpc/grpc-js`; cover current `http://127.0.0.1` behavior, `https:` selection, unsupported schemes, and worker startup regression checks.
- Re-run `npm run build` and `npm test` after implementation. Keep tests network-free so they remain stable across the CI Node version matrix.


larohra:
@copilot apply changes based on the comments in [this thread](https://github.com/Azure/azure-functions-nodejs-worker/pull/811#pullrequestreview-4275144753) 

Also fix all failing UnitTests runs.

swapnil-nagar:
Can you update the version of this package to 3.15.0? We’ve already allocated 3.14.0 for Node 26 support, so let’s wait for 3.14.0 to be merged and the release to be created first.

Also have you verified locally with httpTrigger sample (sanity check).

src/GrpcClient.ts
```diff
@@ -27,13 +28,35 @@ export interface IEventStream {
     end(): void;
 }
 
-export function CreateGrpcEventStream(connection: string, grpcMaxMessageLength: number): IEventStream {
+function getConnectionUri(connection: string | URL): URL {
+    return typeof connection === 'string' ? new URL(connection) : connection;
+}
+
+function getChannelCredentials(connectionUri: URL): grpc.ChannelCredentials {
+    switch (connectionUri.protocol) {
```
swapnil-nagar, line 32: Can we run locally and check.

It can be a breaking change if existing callers pass connection as host:port instead of a full URI and below line suggests thats what are we doing. We may need make changes in worker.ts

Based on this line https://github.com/Azure/azure-functions-nodejs-worker/blob/61408bb0b1c8b24236bcbb587ef54c50d3138a8b/src/Worker.ts#L35

swapnil-nagar (changes requested):

src/GrpcClient.ts
```diff
@@ -27,13 +28,35 @@ export interface IEventStream {
...
+        case 'http:':
+            // Current hosts still treat host<->worker gRPC as trusted localhost IPC, so keep the
+            // insecure fallback until they start advertising an https:// functions-uri.
+            return grpc.credentials.createInsecure();
+        case 'https:':
+            return grpc.credentials.createSsl();
+        default:
+            throw new AzFuncSystemError(
+                `Unsupported gRPC connection URI scheme '${
+                    connectionUri.protocol
```
swapnil-nagar, line 39: Default, can be unsecure connection to avoid any breaking changes.

swapnil-nagar (dismissed):

src/GrpcClient.ts
larohra: I used @TsuyoshiUshio's new skill to run sanity testing for the worker and it works - 

```
The host exited as expected (force-stopped after successful test). Here's the final report:

Sanity Test Result
Item	Value
Worker	Node.js (v3.14.0 — PR branch)
Branch/PR	#811 — larohra/fix-gprc-tls-support
Runtime Version	Node.js v22.17.1
Host URL	http://localhost:62514
Endpoint	/api/httpTrigger?name=SanityTest
HTTP Status	200
Response	Hello, SanityTest!
Unit Tests	138 passing, 8 pending, 0 failing
Result	PASS
The PR's changes (scheme-aware gRPC credential selection) work correctly with the existing http:// host contract. The worker starts up, communicates with the host via gRPC, loads the v4 programming model function, and successfully handles an HTTP trigger invocation end-to-end.
```

src/GrpcClient.ts
larohra: @copilot fix this, update default to treat it as http route

larohra: Also, looked at the docs, Node only permits some special schema - 
`According to the WHATWG URL Standard, special protocol schemes are ftp, file, http, https, ws, and wss.`

Technically we should only be accepting http and https, anything else should be errored out on. It would be a breaking change from existing code where we just ignore the protocol completely, but maybe this can be a good opportunity to test out the hardening @swapnil-nagar

larohra:
@copilot resolve the merge conflicts in this pull request

larohra:
@copilot+gpt-5.4 resolve the merge conflicts in this pull request

src/GrpcClient.ts
swapnil-nagar: Agree. Lets harden it.

swapnil-nagar (approved):

PR #804, opened 2026-05-07
Bump protobufjs to ^7.5.5 and fix proto generation compatibility

TsuyoshiUshio:
## Summary

Fix critical production vulnerabilities by bumping protobufjs and transitive dependencies.

## Changes

1. **package.json**: Bump protobufjs from `^7.2.5` to `^7.5.5`
2. **scripts/generateProtos.js**: Add `-p` (proto path) flag to `pbjs` commands
3. **package-lock.json**: `npm audit fix` to resolve remaining production vulnerabilities (picomatch, lodash, flatted, etc.)

## Why the script change?

protobufjs 7.5+ changed how `import` paths are resolved in proto files. Without `-p protoRoot`, `pbjs` resolves `import "shared/NullableTypes.proto"` relative to the importing file's directory (e.g. `identity/`) instead of the proto root. This caused Dependabot PR #796 to fail CI.

## Vulnerabilities Fixed

- **protobufjs** < 7.5.5: Critical - Arbitrary code execution (GHSA-xq3m-2v4x-88gg)
- **picomatch** <= 2.3.1: High - Method injection + ReDoS (GHSA-3v7f-55p6-f55p, GHSA-c2c7-rcm5-vvqj)
- Additional transitive dependency updates via `npm audit fix`

Production `npm audit` now reports **0 vulnerabilities**.

## Integrated PRs

This PR supersedes the following Dependabot PRs by incorporating their fixes:
- #796 (Bump protobufjs from 7.2.5 to 7.5.5)
- #790 (Bump picomatch from 2.3.1 to 2.3.2)

## Testing

- `npm run build` passes (gen + tsc)
- All 132 unit tests pass
- `npm audit --omit=dev` reports 0 vulnerabilities

Closes #803

PR #815, opened 2026-05-13
Fix bundled worker protobuf inquire resolution

### LINKED ISSUE #816: Remove @protobufjs/inquire pin after upstream fix
## Context

PR #815 pins `@protobufjs/inquire` to `1.1.0` via npm `overrides` in `package.json` to keep the Node.js worker bundle startup-safe after the `protobufjs` 7.5.x security update.

The issue appears related to upstream protobufjs issue https://github.com/protobufjs/protobuf.js/issues/2214, where `@protobufjs/inquire` 1.1.1 changed optional require behavior in a way that breaks webpack bundling and gRPC startup paths.

## Remove the pin when upstream is fixed

Once upstream protobufjs issue #2214 is resolved and a fixed `@protobufjs/inquire` version is released, remove the pin and upgrade.

## Acceptance criteria

- [ ] Upstream issue https://github.com/protobufjs/protobuf.js/issues/2214 is fixed and released in a new package version.
- [ ] Remove the `@protobufjs/inquire` override from `package.json`.
- [ ] Verify `npm run build` passes.
- [ ] Verify `npm run webpack` passes.
- [ ] Verify `npm run host-sanity` passes against a local Azure Functions Host checkout.
- [ ] Verify `npm audit --omit=dev` reports no production vulnerabilities.
- [ ] Verify `npm run lint` passes.
- [ ] Verify `npm test` passes.

Related PR: https://github.com/Azure/azure-functions-nodejs-worker/pull/815

TsuyoshiUshio:
## Summary

Fix the bundled Node.js worker startup regression introduced by the protobufjs 7.5.x update.

The security PR (#804) correctly upgraded `protobufjs` to address GHSA-xq3m-2v4x-88gg. This PR pins `@protobufjs/inquire` to 1.1.0 via npm overrides to avoid the webpack bundling regression in 1.1.1.

## Why

`@protobufjs/inquire` 1.1.1 changed from the webpack-safe `eval("quire"...)` pattern to `require(moduleName)`. Webpack parses `require(moduleName)` as a dynamic context require inside the bundle, so the optional runtime require no longer behaves like Node's normal module resolution. The bundled worker then fails its gRPC event stream startup under the Functions Host.

## Approach

Pin `@protobufjs/inquire` to 1.1.0 using npm `overrides` in `package.json`. This package has had exactly one version change in 9 years, so the pin is low-risk and avoids any custom shim code or performance overhead.

## Changes

- Add npm `overrides` for `@protobufjs/inquire` pinned to `1.1.0` in `package.json`
- Add `npm run host-sanity` to start Azure Functions Host with the bundled worker and validate a minimal HTTP trigger

## Security check

- This does **not** revert the security fix in #804.
- `protobufjs` remains resolved to 7.5.x.
- `@protobufjs/inquire` is pinned at 1.1.0 (the version before the webpack-breaking change).
- `npm audit --omit=dev` reports 0 vulnerabilities.

## Validation

- `npm run build`
- `npm run webpack` -> compiled successfully
- `npm audit --omit=dev` -> 0 vulnerabilities
- `npm run lint`
- `npm test` -> 134 passing

## Follow-up issue

Track removing the pin when upstream protobufjs/protobuf.js#2214 is fixed: #816


swapnil-nagar:
Why not Pin @protobufjs/inquire to 1.1.0 (the version before the regression). This package has hardly changes. It changed for first time in 9 years.
https://www.npmjs.com/package/@protobufjs/inquire?activeTab=versions

no shim would be needed.

swapnil-nagar:
Having  a shim will have any performance overhead?

TsuyoshiUshio:
Good call! Switched to the pin approach. The latest commit:

- Adds `"overrides": { "@protobufjs/inquire": "1.1.0" }` in `package.json`
- Removes the webpack alias and `scripts/webpack-protobufjs-inquire.js` shim
- No custom code, no performance overhead

All checks pass: build, webpack, 134 tests, lint, `npm audit --omit=dev` = 0 vulnerabilities.

Also updated issue #816 to reflect the pin removal (instead of shim removal) as the follow-up when upstream is fixed.

swapnil-nagar (approved):

PR #819, opened 2026-05-19
Sanitize serialized worker errors

TsuyoshiUshio:
## Summary
- add credential-aware sanitization when serializing non-Error thrown values
- redact credential-like property names and known credential tokens while preserving useful object context
- sanitize worker exception messages and stack traces before sending invocation failures to the host
- handle circular references in thrown objects without throwing during error conversion

## Notes
This is a first implementation intended to support review and discussion. The sanitizer follows the same broad approach as azure-functions-host by redacting obvious credential names/tokens, while preserving non-secret diagnostic context.

Related library PR will apply the same serialization behavior in `@azure/functions`.

## Validation
- `npm test` (148 passing, 8 pending)
- `npm run lint`

### DIFF
--- a/src/errors.ts
+++ b/src/errors.ts
@@ -36,6 +36,29 @@ export class ReadOnlyError extends AzFuncTypeError {
     }
 }
 
+const hiddenCredential = '[Hidden Credential]';
+const circularReference = '[Circular]';
+const credentialNameFragments = ['password', 'pwd', 'key', 'secret', 'token', 'sas'];
+const credentialTokens = [
+    'Token=',
+    'DefaultEndpointsProtocol=http',
+    'AccountKey=',
+    'Data Source=',
+    'Server=',
+    'Password=',
+    'pwd=',
+    '&amp;sig=',
+    '&sig=',
+    '?sig=',
+    'SharedAccessKey=',
+    '&amp;code=',
+    '&code=',
+    '?code=',
+    '/code=',
+    'key=',
+];
+const urlCredentialPattern = /\b([a-zA-Z]+):\/\/([^:/\s]+):([^@/\s]+)@([^:/\s]+):([0-9]+)\b/g;
+
 export function ensureErrorType(err: unknown): ValidatedError {
     if (err instanceof Error) {
         return err;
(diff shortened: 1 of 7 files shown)


src/errors.ts
```diff
@@ -36,6 +36,29 @@ export class ReadOnlyError extends AzFuncTypeError {
     }
 }
```
swapnil-nagar, line 40: You may need to change the logic at
https://github.com/Azure/azure-functions-nodejs-worker/blob/7bf60ababe9825da44c39232735a440557422e9f/src/Worker.ts#L44

https://github.com/Azure/azure-functions-nodejs-worker/blob/7bf60ababe9825da44c39232735a440557422e9f/src/Worker.ts#L44

src/errors.ts
```diff
@@ -36,6 +36,29 @@ export class ReadOnlyError extends AzFuncTypeError {
...
+    '&code=',
+    '?code=',
+    '/code=',
+    'key=',
+];
+const urlCredentialPattern = /\b([a-zA-Z]+):\/\/([^:/\s]+):([^@/\s]+)@([^:/\s]+):([0-9]+)\b/g;
+
 export function ensureErrorType(err: unknown): ValidatedError {
     if (err instanceof Error) {
         return err;
```
swapnil-nagar, line 43: Do you need to sanitize err as well.

src/errors.ts
```diff
@@ -44,16 +67,84 @@ export function ensureErrorType(err: unknown): ValidatedError {
...
+    });
+}
+
+function isCredentialName(name: string): boolean {
+    return credentialNameFragments.some((fragment) => name.toLowerCase().includes(fragment));
+}
+
 export function trySetErrorMessage(err: Error, message: string): void {
     try {
         err.message = message;
```
swapnil-nagar: Sanitize err here as well

swapnil-nagar (changes requested):

src/errors.ts
TsuyoshiUshio: Good catch. Updated \Worker.ts\ to sanitize the gRPC event stream creation message before it is assigned back onto the Error.

src/errors.ts
TsuyoshiUshio: Yes. I moved the sanitizer into \src/utils/errorSanitizer.ts\ and kept \nsureErrorType\ from mutating existing Error instances directly. Error instances are now sanitized at the message/logging boundaries (\	rySetErrorMessage\, \setupEventStream\, and \Worker\ uncaught-exception logging).

src/errors.ts
TsuyoshiUshio: Updated \	rySetErrorMessage\ to sanitize the message before assigning it, and added test coverage for that path.

swapnil-nagar (approved):

PR #825, opened 2026-06-12
Fix security vulnerabilities: @grpc/grpc-js (CVE-2026-48069) and uuid (GHSA-w5hq-g745-h8pq)

TsuyoshiUshio:
## Summary

Fixes the v3.x CI security scan failure while preserving the v3 branch compatibility surface.

## Changes

| Area | Before | After | Reason |
|------|--------|-------|--------|
| `@grpc/grpc-js` | `^1.8.22` (resolved `1.8.22`) | `^1.9.16` (resolved `1.9.16`) | Fixes CVE-2026-48069 (High). This is the patched `1.9.x` version. |
| Vulnerability scan | `npm audit --production` | `npm audit --production --audit-level=high` | Keeps High/Critical blocking while allowing the remaining transitive `uuid` Moderate advisory that requires a breaking `@azure/functions` v4 upgrade to fully remove. |
| TypeScript toolchain | TS 4.5 / `@types/node` 16 | TS 5.7 / `@types/node` 24 plus compatible lint packages | Lets the v3 branch compile `@grpc/grpc-js@1.9.16` declaration files without `skipLibCheck`. |
| Source/test typings | Older Node typings (`NodeModule`, `NodeJS.Timer`, unconstrained object generics) | Updated to current Node typings (`NodeJS.Module`, `NodeJS.Timeout`, object constraints) | Required by the newer TS / Node typings. |

### `@grpc/grpc-js` (CVE-2026-48069)

An invalid incoming compressed message can cause a client or server process to crash. No workaround exists; upgrading is required. The advisory lists `1.9.16` as the patched version for the `1.9.x` line.

### `uuid` (GHSA-w5hq-g745-h8pq)

The remaining `uuid` issue is Moderate severity and comes transitively through `@azure/functions@3.x`. Attempting to force `uuid@11` with an npm override broke Node 14 compatibility in CI, and upgrading `@azure/functions` to v4 is a breaking programming-model change for this branch. This PR therefore keeps the production audit gate at High/Critical severity.

### TypeScript compatibility

`@grpc/grpc-js@1.9.16` publishes `.d.ts` files that use generic `Buffer` types. Instead of using `skipLibCheck`, this PR upgrades the TypeScript toolchain and Node typings so those declarations compile directly.

## Verification

Local verification after clean npm 8 install:

- `npx --yes -p npm@8 -c "npm ci"` - success
- `npm run build` - success
- `npm run lint` - success (prints the existing @typescript-eslint support-range warning for TS 5.7, but exits 0)
- `npm run updateVersion -- --validate` - success
- `npm run webpack` - success
- `npm test` - 149 passing, 8 pending
- `npx --yes -p npm@8 -c "npm audit --production --audit-level=high"` - High/Critical pass; only `uuid` Moderate remains

## Related

- Supersedes PR #824 (Dependabot: `@grpc/grpc-js` only)
- PR #820 should remain closed because `@azure/functions` v3 -> v4 is a breaking programming model change incompatible with this v3.x worker servicing PR


tsconfig.json
```diff
@@ -5,6 +5,7 @@
         "noImplicitAny": false,
         "strict": true,
         "noUnusedLocals": true,
+        "skipLibCheck": true,
         "outDir": "dist",
```
swapnil-nagar, line 8: Why skipLibCheck true? Which library is causing the issues?

swapnil-nagar (dismissed):

tsconfig.json
TsuyoshiUshio: The issue is coming from `@grpc/grpc-js@1.9.16`, specifically its published `.d.ts` files. We need `1.9.16` because it is the patched `1.9.x` version for CVE-2026-48069. After that bump, CI failed with:

```text
node_modules/@grpc/grpc-js/build/src/server-call.d.ts(...): error TS2315: Type 'Buffer' is not generic.
```

That type shape is from newer Node/TypeScript typings (`Buffer<ArrayBuffer>`), while this branch is still on TypeScript 4.5 / older `@types/node`. The runtime code is fine; the failure is only TypeScript checking a dependency declaration file in `node_modules`.

Current approach:
- keep the minimal runtime dependency fix: `@grpc/grpc-js` `1.8.22` -> `1.9.16`
- keep the v3 branch's existing TS/tooling surface
- add `skipLibCheck: true` so TS still checks our `src/**/*.ts`, but does not fail on dependency `.d.ts` files
- leave the moderate `uuid` advisory as audit-level filtered, because forcing `uuid@11` broke Node 14 compatibility earlier

Alternatives I considered:
1. Upgrade TypeScript / `@types/node` on the v3 branch. This would likely fix the generic `Buffer` type, but has a larger compatibility surface than this security fix.
2. Patch or vendor the `@grpc/grpc-js` type declarations. That is more fragile and harder to maintain than `skipLibCheck`.
3. Move to a newer patched grpc line (`1.10.12+`, `1.11.4+`, etc.). That is a larger dependency jump and likely keeps the newer type surface anyway.
4. Ignore the high grpc advisory / rely only on audit filtering. I do not think that is acceptable because CVE-2026-48069 is high severity and has no workaround.

One related question: do we have a defined EOL/service policy for the `v3.x` worker branch itself? I checked the public docs and found runtime v3 is EOL, but the worker branch is intentionally decoupled from runtime version, is still the default branch, and recently got Node 26 support. I did not find a separate EOL date for the worker `v3.x` branch. If this branch is close to EOL, then `skipLibCheck` feels even more reasonable; if we expect long-term servicing, I can also look into whether a broader TS/tooling update is worth doing separately from this CVE fix.

swapnil-nagar: There was a discussion on deprecation of v3 vs v4, we haven't made a decision yet. So we need to maintain until we have path forward.

my concern with skipLibCheck: true is it can hide real type problems and make debugging harder later.

Can we try the approach 
"Upgrade TypeScript / @types/node on the v3 branch. This would likely fix the generic Buffer type, but has a larger compatibility surface than this security fix." 

If it has too many changes then we can skip, else we can move forward with it.

TsuyoshiUshio: Make sense. :) Will do it and try it out.

TsuyoshiUshio: Done. I removed `skipLibCheck` and switched to the TypeScript/tooling upgrade approach.

What changed:
- upgraded `typescript` to `^5.7.3`
- upgraded `@types/node` to `^24.0.0`, which provides the generic `Buffer` typing needed by `@grpc/grpc-js@1.9.16`
- upgraded `@typescript-eslint/*` within the Node 14-compatible v5 line and `eslint-plugin-deprecation` to a TS5-compatible v1.5 line
- updated the small set of project/test typings exposed by the newer Node typings (`NodeJS.Module`, `NodeJS.Timeout`, object generic constraints)
- removed `skipLibCheck`

Local validation after clean npm 8 install:
- `npm ci` with npm 8: pass
- `npm run build`: pass
- `npm run lint`: pass (with a warning that @typescript-eslint v5 does not officially list TS 5.7, but exit code is 0)
- `npm run updateVersion -- --validate`: pass
- `npm run webpack`: pass
- `npm test`: 149 passing, 8 pending
- production high audit: pass; only the transitive `uuid` Moderate remains

I also updated the PR description to reflect this new approach and the current verification results. CI is running from the new push now.

swapnil-nagar: Thanks a lot

swapnil-nagar (approved):

harshivcodes (approved):

PR #832, opened 2026-07-17
build: bootstrap azfunc npm registry without checking in .npmrc

jviau:
## Summary

A 1ES policy blocks public npm access on dev machines, but this is a public repo we want to keep welcoming to external contributors. This change lets Microsoft developers and CI resolve packages from the correct azfunc Azure Artifacts feed **without checking an `.npmrc` into the public repo** — so a plain `git clone` + `npm install` still uses the default public registry (`registry.npmjs.org`) for public contributors.

## How it works

Three dependency-free CommonJS scripts under `scripts/` (Node built-ins only, so they run before `node_modules` exists):

- **`scripts/buildEnv.js`** — cross-platform environment detection. Returns Azure DevOps context (org/project from `SYSTEM_TEAMFOUNDATIONCOLLECTIONURI` + `SYSTEM_TEAMPROJECT`) or a best-effort Microsoft-user signal (git email, Windows UPN/domain, macOS Enterprise SSO, mac/Linux Kerberos), or `null`.
- **`scripts/setupNpmrc.js`** (`npm run setup:npmrc`) — writes a local, gitignored `.npmrc` pointing at the right feed:
  - ADO `internal` project → `azfunc/internal/upstream`
  - ADO `public` project / local dev → `azfunc/public/upstream-public`
- **`scripts/preinstallNpmrc.js`** (`preinstall` hook) — best-effort gate that only acts for Microsoft users or the azfunc ADO org; no-ops if `.npmrc` already exists; public contributors are never touched. On a Microsoft machine's first bootstrap it writes `.npmrc` and intentionally stops the install (npm loads its config before `preinstall` runs, so the new registry can't apply to the current run) with instructions to authenticate and re-run.

## CI wiring

Added `node scripts/setupNpmrc.js` + `npmAuthenticate@0` before `npm ci` in both templates that install (`templates/build.yml`, `templates/test.yml`). The public pipeline resolves the public feed and the official pipeline resolves the internal feed automatically via env detection. Since `.npmrc` exists by the time `npm ci` runs, the `preinstall` hook is a no-op in CI.

## Docs

Added an **npm registry configuration** section to the README covering the public-contributor (no-op) path, the Microsoft-dev setup + auth steps, and the first-run bootstrap behavior.

## Notes

- `.npmrc` is gitignored so it is never committed and never affects public contributors.
- Scripts are CommonJS for consistency with the existing `scripts/*.js` and to avoid needing `"type": "module"`.


azure-pipelines/templates/test.yml
jviau, line 1: my VS code formatting kicked in and changed the spacing of all the yml files to 2 (was a mix of 2 and 4 before). Let me know if you want me to revert that. Yaml files are _typically_ 2 spaces, so I left it in as an improvement.

README.md
```diff
@@ -28,6 +28,16 @@
...
 
+### npm registry configuration
+
+This repository does **not** check in an `.npmrc`, so a plain `git clone` + `npm install` uses the default public npm registry (`registry.npmjs.org`). **Public contributors don't need to do anything** — this section is only relevant to Microsoft employees.
+
+- **Configure the azfunc feed:** run `npm run setup:npmrc`. It detects your environment and writes a local, gitignored `.npmrc` pointing at the correct Azure Artifacts feed (`azfunc/public/upstream-public` locally, `azfunc/internal/upstream` in internal CI).
+- **Authenticate:** run `npx better-vsts-npm-auth -c .npmrc` (one-time, refreshed periodically), then run `npm install`.
+- **First-time bootstrap:** on a Microsoft machine, an `npm install` with no existing `.npmrc` runs a `preinstall` hook that writes the `.npmrc` for you and then intentionally stops the install. npm had already loaded its config before the hook ran, so it can't use the new registry yet. Just authenticate (step above) and re-run `npm install` — the second run proceeds normally. Once `.npmrc` exists, the hook is a no-op.
```
TsuyoshiUshio, line 37: One feedback. It won't update the feed always. we need to clear cache. Otherwise, it won't reach the registry.

jviau: Clear what cache exactly?

TsuyoshiUshio: I mean npm's local content cache. We need the request to reach the Azure Artifacts feed so that a package version that is not yet in the feed is pulled from the upstream source and saved. If the package is already in the machine-wide npm cache, `npm install` can complete without contacting the feed, so that population never happens.

Rather than clearing the user's global cache, we can force a feed request with an empty temporary cache and request the exact version, for example:

```console
npm pack <package>@<version> --registry=<feed-url> --cache=<empty-temp-directory>
```

`npm install <package>@<version>` can be used in a temporary project in the same way. Could we document this step for populating a missing package version?

jviau: I don't that that will be necessary. Hitting cache locally seems fine to me. When a CI or PR build runs that will at minimum not have any local cache and force the upstream hydration.

Do you have concerns regarding a cached build?

TsuyoshiUshio: Ok. You mean give the CI to the permission to update the registry. Although it has npm audit, What if an attacker adds malicious package/version to package.json on his pr, it is registered to the feed. does that ok?

jviau: for community PR builds from a fork, this will run with reduced permissions and not be able to ingest new packages from upstream.

Any community contribution which needs to be bring in a new package will require team member action to evaluate the package and ingest it

TsuyoshiUshio (approved):
LGTM Let's see if pipeline success.

README.md
TsuyoshiUshio: Sounds good. Thank you for the contribution.

jviau:
> LGTM Let's see if pipeline success.

The public CI build did succeed, and the logs show it did write `.npmrc` before running `npm ci`.

https://azfunc.visualstudio.com/public/_build/results?buildId=291228&view=logs&j=9dbb085e-85c1-56a8-f365-135f9ff5fdd6&t=95bdec7d-40c5-5cf5-4172-6d03b58c438d

jviau:
@TsuyoshiUshio, @swapnil-nagar -- if this approach is acceptable to all of you, could I ask that you two replicate this work in all our other node/npm based repos?

swapnil-nagar:
> @TsuyoshiUshio, @swapnil-nagar -- if this approach is acceptable to all of you, could I ask that you two replicate this work in all our other node/npm based repos?

Will do

swapnil-nagar (approved):

PR #834, opened 2026-07-29
build: configure azfunc npm registry via CI setup template

jviau:
## Summary

Follow-up to #832. Simplifies the npm feed strategy by removing the local `.npmrc` bootstrap entirely and configuring the azfunc Azure Artifacts feed only in CI.

- Deletes the bootstrap scripts (`scripts/buildEnv.js`, `scripts/setupNpmrc.js`, `scripts/preinstallNpmrc.js`) and the `preinstall` / `setup:npmrc` npm scripts.
- Adds a shared `azure-pipelines/templates/steps/setup.yml` step template that installs Node, writes a `.npmrc` pointing at the correct feed based on the ADO project (`internal` → `azfunc/internal/upstream`, otherwise → `azfunc/public/upstream-public`), and authenticates via `npmAuthenticate@0`.
- Wires that step into `build.yml` and `test.yml`.
- Updates `README.md` and `.gitignore` to match.

No `.npmrc` is checked in, so a plain `git clone` + `npm install` continues to use the default public registry — **public contributors need to do nothing.**

## Why the approach changed

The original PR (#832) tried to bootstrap a local `.npmrc` so Microsoft developers on machines with the 1ES public-npm block could still resolve packages. That's no longer necessary: **group policy now provisions a machine/global npm config for Microsoft developers that points at a managed npm proxy.** Local dev resolves through that managed proxy automatically, with no repo-level `.npmrc` and no per-repo setup step.

That leaves CI as the only place that needs explicit feed configuration, which this PR handles in the pipeline templates — keeping the repo clean for public contributors while internal/public CI builds pull from the appropriate azfunc feed.


swapnil-nagar (approved):

swapnil-nagar (approved):

jviau:
Abandoning for now. Going with a simpler approach of just checking in an .npmrc with the public feed added. We can revisit this if the configured ADO registry is problematic.

PR #835, opened 2026-07-29
build: add husky pre-commit hook to normalize package-lock.json registry URLs

jviau:
## Summary

Adds [husky](https://typicode.github.io/husky/) and a pre-commit hook that normalizes `package-lock.json` `resolved` URLs to the public npm registry (`https://registry.npmjs.org`), so the committed lockfile stays registry-agnostic for external contributors.

## Why

In some environments, Azure DevOps (ADO) package feeds may be used as a proxy to the public npm registry. When a dependency is added or updated on such a machine, npm writes `resolved` URLs pointing at the ADO feed (either `*.pkgs.visualstudio.com` or `pkgs.dev.azure.com` host form) instead of the public registry. Those URLs vary between environments — causing spurious lockfile churn and merge conflicts — and break contributors who don't use the same feed.

Normalizing to `registry.npmjs.org` works for everyone: npm's `replace-registry-host` config defaults to `npmjs`, so a dev resolving through an ADO feed has npm substitute their proxy at install time, while everyone else hits the real public registry. `integrity` (sha512) hashes are source-independent, so tarballs validate regardless of origin.

## What's included

- **`scripts/normalizeLock.js`** — dependency-free, dual-mode:
  - default: rewrites `package-lock.json` in place, normalizing ADO feed prefixes to the public registry (prints only if it changed)
  - `--check`: exits non-zero with an actionable message if any non-public registry URL remains, without modifying the file (for CI)
  - no-ops cleanly if the lockfile is absent
- **`.husky/pre-commit`** — runs the normalizer and re-stages the lockfile **only when `package-lock.json` is already staged**; leaves it untouched otherwise
- **`package.json`** — adds `husky` devDependency, a `prepare` script to bootstrap husky, and `normalize-lock` / `check-lock` scripts
- **CI gate** — `npm run check-lock` in `azure-pipelines/templates/build.yml` (the shared path that runs for external contributor PRs), right after `npm ci`
- **`.gitattributes`** — pins `.husky/*` to LF endings
- **`.gitignore`** — ignores husky's generated `.husky/_`
- **README** — documents the normalization and how to recover if the CI gate fails (`npm run normalize-lock`)

## Notes for reviewers

- Normalizing to `registry.npmjs.org` specifically is load-bearing — changing the target host loses the `replace-registry-host` substitution behavior.
- `postinstall` is intentionally **not** used: npm doesn't run root lifecycle scripts on `npm install <pkg>` (the case that dirties the lockfile), so a real git pre-commit hook is required.
- The regex is anchored on the `/npm/registry` shape and stops there, so only the host+feed prefix is replaced and the `/<pkg>/-/<tarball>` suffix is preserved. A genuinely private source would be surfaced by `--check` for review rather than silently rewritten.


swapnil-nagar (approved):

jviau:
Abandoning for now. Going with a simpler approach of just checking in an .npmrc with the public feed added. We can revisit this if the configured ADO registry is problematic.

PR #845, opened 2026-08-27
fix: deserialize protobuf longs as numbers (#844)

### LINKED ISSUE #844: [Regression] Event Hubs batch sequenceNumberArray exposed as Long-like objects with Worker 3.16.0
## Summary

After Functions Host 4.1053 began rolling out, Node.js Event Hubs functions using batch cardinality can receive `context.triggerMetadata.sequenceNumberArray` as Long-like objects instead of JavaScript numbers.

Working behavior with Host 4.1052 / Node.js Worker 3.14.1:

```js
[2251, 2252, 2253]
```

Regressed behavior with Host 4.1053 / Node.js Worker 3.16.0:

```js
[
  { low: 2251, high: 0, unsigned: false },
  { low: 2252, high: 0, unsigned: false },
  { low: 2253, high: 0, unsigned: false }
]
```

The Event Hubs extension and application package do not need to change for the behavior to appear. Recycling from Host 4.1052 to 4.1053 is sufficient.

- First observed: 2026-08-26
- Affected Host: 4.1053.200.x
- Known-good Host: 4.1052.300.x
- Affected Node.js Worker: 3.16.0
- Known-good Node.js Worker: 3.14.1
- Event Hubs extension: 6.5.3 on both sides of the comparison
- `@azure/functions`: 4.11.2 on both sides of the customer comparison
- Reproduced on Flex Consumption and Dedicated/App Service plans

Customer-specific app and invocation information is intentionally omitted from this public issue.

#### Repro steps

1. Create a Node.js Event Hubs trigger using batch cardinality (`cardinality: "many"`).
2. Log or otherwise inspect `context.triggerMetadata.sequenceNumberArray`.
3. Run on Host 4.1052 with Node.js Worker 3.14.1 and observe primitive numbers.
4. Run the same application and Event Hubs extension on Host 4.1053 with Node.js Worker 3.16.0.
5. Observe Long-like objects containing `low`, `high`, and `unsigned`.

A wire-level worker test also reproduces the behavior: `@grpc/proto-loader` returns Long objects when no `longs` conversion policy is specified, while `longs: Number` returns `[2251, 2252, 2253]`.

#### Expected behavior

`CollectionSInt64` trigger metadata should continue to be exposed using the existing JavaScript `number` contract:

```js
context.triggerMetadata.sequenceNumberArray // number[]
```

A dependency or bundling change must not silently alter the public shape of trigger metadata.

#### Actual behavior

`CollectionSInt64` values are decoded as protobuf Long objects. The `@azure/functions` trigger metadata camel-casing path subsequently copies their enumerable fields into plain objects, so application code receives objects without `Long.prototype.toNumber()` or `toString()`.

This can break code that performs arithmetic, validation, persistence, or equality checks assuming `sequenceNumberArray` contains numbers.

#### Customer impact

An application is impacted when it does one or more of the following:

- Reads `context.triggerMetadata.sequenceNumberArray` and treats its elements as numbers.
- Uses another trigger metadata value represented by protobuf `sint64` or `CollectionSInt64` and assumes the documented primitive-number shape.
- Serializes, compares, validates, or forwards the affected metadata and depends on its previous shape.

Applications that do not access or use the affected trigger metadata are not expected to be impacted. Event payload delivery and normal Event Hubs trigger execution are not, by themselves, changed by this regression.

##### Dedicated/App Service plans where exact Host pinning is supported

Temporarily pin the Function App to a known-good 4.1052 Host build, such as `4.1052.300.26370`, using the supported exact-version pinning mechanism for that environment. For environments that use `FUNCTIONS_EXTENSION_VERSION` for exact pinning, set it to the known-good exact version and restart the app.

Customers should coordinate an exact runtime pin with Azure Support, especially for Linux-hosted apps, and restore the normal `~4` tracking setting after the hotfix has rolled out. Exact Host pinning is a temporary mitigation, not the long-term fix.

##### Flex Consumption, or when Host pinning is unavailable

Flex Consumption does not provide a customer-controlled exact Host version pin. Normalize the metadata in the Event Hubs handler before using it:

```ts
type LongLike = {
    low: number;
    high: number;
    unsigned?: boolean;
    toNumber?: () => number;
};

function toSafeNumber(value: unknown): number {
    if (typeof value === "number") {
        return value;
    }

    if (typeof value === "string") {
        const result = Number(value);
        if (!Number.isSafeInteger(result)) {
            throw new RangeError("Value exceeds the JavaScript safe integer range");
        }
        return result;
    }

    const long = value as LongLike;
    if (typeof long?.toNumber === "function") {
        const result = long.toNumber();
        if (!Number.isSafeInteger(result)) {
            throw new RangeError("Value exceeds the JavaScript safe integer range");
        }
        return result;
    }

    if (!Number.isInteger(long?.low) || !Number.isInteger(long?.high)) {
        throw new TypeError("Expected a number or protobuf Long-like value");
    }

    const high = BigInt(long.unsigned ? long.high >>> 0 : long.high);
    const result = Number((high << 32n) + BigInt(long.low >>> 0));
    if (!Number.isSafeInteger(result)) {
        throw new RangeError("Value exceeds the JavaScript safe integer range");
    }

    return result;
}

const rawSequenceNumbers = context.triggerMetadata?.sequenceNumberArray;
const sequenceNumbers = Array.isArray(rawSequenceNumbers)
    ? rawSequenceNumbers.map(toSafeNumber)
    : [];
```

The application must explicitly call this conversion, as shown by `.map(toSafeNumber)`, before using the metadata. The workaround accepts both the old primitive-number shape and the regressed Long-like shape, so it can remain in place while the platform hotfix rolls out.

#### Root cause

Worker 3.16.0 resolved `protobufjs` 7.6.5 instead of 7.5.6. In protobufjs 7.6.x, `long` became statically resolvable to webpack after the `@protobufjs/inquire` dependency was removed. The worker bundle therefore contains a Long implementation where the previous bundle effectively did not.

The worker currently loads the RPC descriptor without an explicit 64-bit integer conversion policy:

```ts
grpcloader.fromJSON(jsonModule, {
    objects: true,
    defaults: true,
    oneofs: true,
});
```

Because `longs` is unspecified, the decoded JavaScript type depends on whether a Long implementation is available in the bundle.

#### Proposed hotfix direction

The current direction is to release a Node.js Worker hotfix that:

1. Sets `longs: Number` explicitly in the worker's `@grpc/proto-loader` options, restoring the established JavaScript `number` contract.
2. Adds regression coverage for scalar `TypedData.int` and `CollectionSInt64`, including validation against the webpack bundle.
3. Re-evaluates the now-inert `@protobufjs/inquire` package override.

We will post progress updates on this issue, including the fix PR, hotfix worker version, Functions Host integration, and rollout status as those details become available.

#### Related information

- Language: JavaScript/TypeScript
- Binding: Event Hubs trigger with batch cardinality
- Worker code path: `src/GrpcClient.ts`
- Library conversion paths: `src/converters/fromRpcTypedData.ts` and `src/converters/toCamelCase.ts` in `@azure/functions`

TsuyoshiUshio:
## Summary

Explicitly decode protobuf 64-bit integer fields as JavaScript numbers so bundle dependency resolution cannot change the public trigger metadata shape.

## Issue

Fixes #844

## Changes

- Set `longs: Number` in the worker's `@grpc/proto-loader` options.
- Add a wire-level unit test for scalar `TypedData.int` and `CollectionSInt64` values.
- Add a CI regression test that starts an isolated copy of the production webpack bundle and verifies both values through the Host-worker gRPC path.

## Breaking Changes

None. This restores the JavaScript `number` behavior from Worker 3.14.1 and earlier.

## Testing

- `npm run build`: pass.
- `npm run webpack`: pass.
- `npm run test:bundle`: pass.
- `npm test`: pass, 150 passing and 8 pending.
- `npm run lint`: pass.

## Notes

- The bundle test copies only `nodejsWorker.js` and `worker-bundle.js` into an isolated directory, so the unbundled fallback cannot hide a bundle failure.
- Values above `Number.MAX_SAFE_INTEGER` remain outside the precise JavaScript `number` range. A string or BigInt contract needs a separate compatibility decision.


swapnil-nagar (dismissed):

src/GrpcClient.ts
```diff
@@ -15,6 +15,7 @@ function GetGrpcClientConstructor(): ServiceClientConstructor {
         objects: true,
         defaults: true,
         oneofs: true,
+        longs: Number,
```
swapnil-nagar, line 18: longs: Number is the correct backward-compatible fix, but we should explicitly define all supported proto-loader options instead of relying on defaults.
The protobufjs update made Long available to webpack, changing decoded values from numbers to { low, high, unsigned } objects.
Add wire-level tests for TypedData.int, CollectionSInt64, shared-memory offsets/counts, and timestamps.
Run these tests against the production webpack bundle, since bundling caused the regression.
Document that values above Number.MAX_SAFE_INTEGER require a future string or BigInt contract.


```
const rpcLoaderOptions: grpcloader.Options = {
    keepCase: false,
    longs: Number,
    defaults: true,
    arrays: true,
    objects: true,
    oneofs: true,
    json: false,
};

const packageDef = grpcloader.fromJSON(
    jsonModule as protobuf.INamespace,
    rpcLoaderOptions
);```

**Comment by TsuyoshiUshio** [comment]
## Follow-up compatibility audit: protobufjs 7.5.6 to 7.6.0

I reviewed the published package diff and the relevant transitive dependency changes to check for other breaking behavior beyond #844.

### Findings

- **Confirmed breaking behavior: Long discovery.** `protobufjs` replaced the bundler-hidden `util.inquire("long")` lookup with a static `require("long")`. This made Long visible to webpack and changed the default decoded shape of 64-bit fields. It can also change constructor identity in a Node dependency graph containing multiple Long versions: in an isolated test, 7.5.6 adopted proto-loader's Long 4.x, while 7.6.0 adopted protobufjs's own Long 5.x. Explicit `longs: Number` in this PR removes that dependency-resolution ambiguity from the worker contract.
- **Additive: BigInt conversion.** `toObject({ longs: BigInt })` was added. It does not affect the worker unless explicitly selected.
- **Conditional compatibility risk: recursion limits.** JSON reflection and `.proto` parsing now enforce `util.recursionLimit` (100). The Functions RPC reflection tree has a maximum depth of 3, leaving 97 levels of headroom.
- **Conditional compatibility risk: namespace lookup order.** Recursive namespace lookup now stops at the first match. The current Functions RPC descriptor does not produce a behavioral difference.
- **Conditional bundler/browser risk: `fs` and `Buffer` discovery.** `fs` now uses a guarded bundler-aware require and package browser mapping; Buffer discovery now uses `global.Buffer`. The worker passes a generated JSON descriptor to `fromJSON` and does not use protobufjs file loading, so these paths do not affect it.
- **Type-only change in `@protobufjs/fetch`.** Its exported option type was renamed from `FetchOptions` to `IFetchOptions`. The worker does not import this API or type.
- **Transitive Long update.** The minimum Long range moved from `^5.0.0` to `^5.3.2`, adding BigInt and package-export changes. With `longs: Number`, these details no longer determine the public decoded value shape.

### Production descriptor comparison

I loaded the same generated Functions RPC descriptor in isolated environments pinned to protobufjs 7.5.6 and 7.6.0, using the production proto-loader options, and compared:

- all package-definition entries and message descriptors
- service paths and streaming flags
- representative serializer output

The normalized results were identical: both were 8,048 bytes with SHA-256 `0E46B3C68DCEB922D21E49FFF99FBBB4ABEFC460816D015CAE2EA11B97E14607`. The descriptor resolved to 53 messages, 3 namespaces, 1 service, and 11 enums. The focused GrpcClient test file also passed (5/5) in the audit worktree.

### Conclusion and follow-up

I did not find another protobufjs 7.6.0 change that breaks the current Node.js worker contract. The Long default was the material regression, and this PR fixes it at the correct boundary by making the expected representation explicit.

For future updates, I recommend **risk-based compatibility comparison rather than exhaustive comparison for every minor dependency update**. Dependencies affecting wire formats, serialization, generated descriptors, public value shapes, or bundling should be compared before and after using the same production inputs and bundle configuration. I opened #846 to track that policy, a dependency-update checklist, future `AGENTS.md` / `architecture.md` guidance, and additional contract coverage.

File: src/GrpcClient.ts (hunk 1 above)
Comment (TsuyoshiUshio) [reply src/GrpcClient.ts:18]: Thanks. I added a CI test that starts an isolated production bundle through a local gRPC Host. It verifies that `TypedData.int` and `CollectionSInt64` decode as numbers.

I did not add duplicate shared-memory or timestamp cases because proto-loader applies the same `longs` conversion to all 64-bit fields. I kept only effective loader options: `keepCase` does not apply to `fromJSON`, and `defaults: true` already covers arrays and objects. The safe-integer limit is documented in #844.

File: test/GrpcClient.test.ts [hunk 2]
```diff
@@ -42,6 +42,48 @@ describe('GrpcClient', () => {
         }
     });
 
+    it('deserializes protobuf long values as numbers', () => {
```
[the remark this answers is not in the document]
TsuyoshiUshio: Addressed in `b313ea4`. CI now starts an isolated copy of the production webpack bundle through a local gRPC Host and verifies that scalar and collection 64-bit values decode as numbers.

swapnil-nagar (approved):
Reviewed the latest changes through b313ea4. The explicit `longs: Number` option restores the established worker contract at the protobuf decoding boundary, and the added wire-level test covers both scalar `TypedData.int` and `CollectionSInt64`. The production-bundle regression test also runs an isolated copy of the shipped artifacts through the Host-worker gRPC path, so it directly guards against the webpack dependency-resolution regression from #844 rather than only testing unbundled source. All current CI checks are passing. I found no blocking issues. The precision limitation above `Number.MAX_SAFE_INTEGER` is appropriately called out for a separate compatibility decision.

swapnil-nagar (approved):

vrdmr (approved):

